PR #4, opened 2014-06-04
Made parsing of includes, fields and parameters work based on comma sepa...

lgebhardt:
...rated lists.


lib/json/api/resource_controller.rb
```diff
@@ -163,6 +164,12 @@ def verify_params(params, klass, resource_param_set)
         return checked_params, checked_associations
       end
 
+      def parse_includes(includes)
+        included_resources = []
+        included_resources += CSV.parse_line(includes) unless includes.nil? || includes.empty?
```
dgeb, line 169: doesn't empty? cover nil?

PR #9, opened 2014-07-02
Reverted back from using before_filters to parse fields, includes and filters.

lgebhardt:
The before filters were causing load order issues when used with other filters, particularly authorization filters.


dgeb:
Ah, that makes sense. This functional approach may be more clear anyway.

PR #18, opened 2014-09-10
Update Readme

dchacke:
It seems like you explicitly have to require the serializer, otherwise it doesn't work for me in the Rails console.


lgebhardt:
In the master branch there are some missing require statements in `jsonapi-resources/lib/jsonapi-resources.rb`. I have these in the content-type branch, which will hopefully be merged soon. I think these will fix the issue you are addressing here without needing the require statement. I'm going to merge this though since this may help someone out until we get the code changes merged. Thanks.

PR #23, opened 2014-09-24
Remove unused require

ahx:
Removes unused `require 'action_dispatch/routing/mapper'`

We are trying to use jsonapi-resources in a non-rails environment and found this along the way. Seems to work just as before.


lgebhardt:
Thanks!

PR #30, opened 2014-10-15
Sorting

whatthewhat:
http://jsonapi.org/format/#fetching-sorting

Hi, thanks for the awesome gem!

I've tried to implement basic resource sorting (`GET /people?sort=age,name`, not `GET /posts?include=author&sort[posts]=-created,title&sort[people]=name`)

The easies way to implement sorting was to shoehorn it into the `.find` method, which is what I did.
The downside is that `.find`'s signature has changed.

What I would really like to do is something like this:

``` ruby
def index
  render json: JSONAPI::ResourceSerializer.new.serialize_to_hash(
    resource_klass
      .find(@request.filters)
      .sort(@request.sort_params)
      #...
```

But that would be a pretty big refactoring..

Anyway, will appreciate any feedback.


lgebhardt:
@whatthewhat, thanks for this. It looks like a good start. I'll see about getting it merged soon.

dgeb:
I think we should consider changing the `find()` signature to take a hash as a second arg, so that `context` and other params, like `sort`, can be passed together.

whatthewhat:
@lgebhardt @dgeb 
Thanks for the feedback! I'll update the PR in a couple days.

lgebhardt:
I would like to make an attribute option to control if an attribute can be sorted or not. I'm thinking name this `sortable`, default it to true in `default_attribute_options`.

If we moved the check_sort_param to the resource class it would allow a resource to override the default logic. I don't have a good use case for needing this, but it would be consistent with the way we handle filters (`_allowed_filter?`).

lgebhardt:
Is there value in allowing sorting on names that do not exist as an attribute, for example `sort=name` where `name` would expand to some set of actual attributes (`last_name` and `first_name`) to sort on? I'm not sure if this is JSON:API compliant. I don't have a good use for this in mind, but if there is one I'd like to code it in such a way that we could add it later. Any thoughts?

whatthewhat:
@lgebhardt, we actually have this exact use case in one of our apps (name->first_name,last_name).

Controlling whether an attribute can be sorted or not is a great idea, but I think implementing it in `default_attribute_options` might be a bit weird (i.e. resource already has an attribute named `sorting`).

What do you think about adding a `sortable_fields` hook to resources, similar to `fetchable_fields`?

whatthewhat:
I've changed `.find` to accept an options hash, but this does create some inconsistency with other methods, like `find_by_key(key, context = nil)`..

The signature could be `find(filters, context = nil, options = {})`, but I don't think multiple optional arguments is a very good idea either.

lgebhardt:
@whatthewhat, after further thought I agree we should use `sortable_fields` instead of an option on the attribute definition to determine if a column is sortable.

`find_by_id` should be consistent and take an options hash. We should review all the signatures in the codebase with an eye towards making them more consistent. But I don't we need to address that in this pull request.

lgebhardt:
After further thought and discussion I removed the sort_columns option in my last comment. We should either break out the sorting and the where clause generation into easily overridable methods, or look at a refactor of the find logic to something like ActiveRecord's Arel like is suggested in the initial comment.

whatthewhat:
I've added `sortable_fields` to the resource class, but now I'm having some second thoughts about `InvalidSortParam` error.

Maybe it's better to just ignore invalid (or not present in `sortable_fields`) params, than returning a 400?

lgebhardt:
I think we need to return an error when an invalid sort param is supplied. Otherwise the consumer of the API will assume they are sorted as specified.

lgebhardt:
Thanks!

whatthewhat:
> Otherwise the consumer of the API will assume they are sorted as specified.

Good point.

Thanks for merging!

PR #32, opened 2014-10-20
Reworks key option into primary and foreign keys

lgebhardt:
excludes a few lines from coverage calculations to get us to 100% coverage
fix so name spaced models can be used


dgeb:
:shipit:

PR #34, opened 2014-10-27
Options refactor

lgebhardt:
Bring find_by_key into line with find as far as context and options are concerned.


dgeb:
@lgebhardt This is a good start. We should do a full review of everywhere `context` is passed.

lgebhardt:
@dgeb Thanks. I agree. I don't see an obvious need to pass other options in the remaining methods we're using context, but it would certainly make sense to review it.

PR #35, opened 2014-11-07
Fix N+1 issue with show multiple action

sineed:
With this changes it makes only one DB query for requests like `/users/1,2,3`.

I thought using `.find` method, but i think it's wrong based on [this](http://jsonapi.org/format/#fetching-filtering).

### DIFF
--- a/lib/jsonapi/resource.rb
+++ b/lib/jsonapi/resource.rb
@@ -261,6 +261,21 @@ module JSONAPI
         self.new(model, context)
       end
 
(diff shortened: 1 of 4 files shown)


lgebhardt:
Thanks for catching this. It should speed things up nicely.

I made a few changes, which are now the PR/35 branch if you want to review them. I renamed the get_resources to find_by_keys to be consistent, and moved the special logic to handle requests for a single id as a single resource into the controller.

Issue #36, opened 2014-11-10
`undefined method 'id' for nil:NilClass` when include nullable association 


hesalx:
undefined method 'id' for nil:NilClass on [resource.rb line 24](https://github.com/cerebris/jsonapi-resources/blob/v0.0.9/lib/jsonapi/resource.rb#L24).

Occurs when include resources with association set to nil.

``` ruby
class ContactResource < JSONAPI::Resource
  has_one :manager # may be nil
end
```

`GET /contacts?include=manager` causes error only if whatever contact has manager = nil

Looks like some explicit check for `model` is needed on [resource_serializer.rb line 174](https://github.com/cerebris/jsonapi-resources/blob/v0.0.9/lib/jsonapi/resource_serializer.rb#L174).

p.s.: full lib trace

```
jsonapi-resources (0.0.9) lib/jsonapi/resource.rb:24:in `id'
jsonapi-resources (0.0.9) lib/jsonapi/resource_serializer.rb:175:in `block in links_hash'
jsonapi-resources (0.0.9) lib/jsonapi/resource_serializer.rb:153:in `each'
jsonapi-resources (0.0.9) lib/jsonapi/resource_serializer.rb:153:in `each_with_object'
jsonapi-resources (0.0.9) lib/jsonapi/resource_serializer.rb:153:in `links_hash'
jsonapi-resources (0.0.9) lib/jsonapi/resource_serializer.rb:117:in `object_hash'
jsonapi-resources (0.0.9) lib/jsonapi/resource_serializer.rb:189:in `block (2 levels) in links_hash'
jsonapi-resources (0.0.9) lib/jsonapi/resource_serializer.rb:185:in `each'
jsonapi-resources (0.0.9) lib/jsonapi/resource_serializer.rb:185:in `block in links_hash'
jsonapi-resources (0.0.9) lib/jsonapi/resource_serializer.rb:153:in `each'
jsonapi-resources (0.0.9) lib/jsonapi/resource_serializer.rb:153:in `each_with_object'
jsonapi-resources (0.0.9) lib/jsonapi/resource_serializer.rb:153:in `links_hash'
jsonapi-resources (0.0.9) lib/jsonapi/resource_serializer.rb:117:in `object_hash'
jsonapi-resources (0.0.9) lib/jsonapi/resource_serializer.rb:100:in `block in process_primary'
jsonapi-resources (0.0.9) lib/jsonapi/resource_serializer.rb:94:in `each'
[...]

PR #37, opened 2014-11-10
Fix include associations which might be nil.

hesalx:
Fixes the issue #36
has_one association should return nil is source model's one is nil. Or I think so.
You can try the test without other changes to reproduce the exact issue I had encountered to find better solutions if possible.


lgebhardt:
@arantir Thanks!

PR #38, opened 2014-11-11
Fix N+1 issue with show multiple action - slight changes

lgebhardt:
Makes a few changes to @sineed's pull request. 

Renamed the get_resources to find_by_keys to be consistent, and moved the special logic to handle requests for a single id as a single resource into the controller.


dgeb:
Looks good to me! :shipit:

sineed:
+1

PR #43, opened 2014-11-24
Associate api error detail with model error message

sineed:


lib/jsonapi/resource.rb
```diff
@@ -107,7 +107,7 @@ def save
                           code: JSONAPI::VALIDATION_ERROR,
                           status: :bad_request,
```
sineed: @lgebhardt @dgeb Why there is 400 code? Thought , that 422 is more suitable.

lgebhardt: I agree, 422 would be better.

lib/jsonapi/resource.rb
```diff
@@ -107,7 +107,7 @@ def save
...
                           title: "#{element[0]} - #{message}",
-                          detail: "can't be blank",
+                          detail: message,
                           path: "\\#{element[0]}"))
```
whatthewhat: @lgebhardt @dgeb @sineed 
Hi, I'm not quite sure about `path`.. shouldn't it be `"/#{element[0]}"` instead of `"\\#{element[0]}"`?

Probably related to https://github.com/json-api/json-api/pull/281

sineed:
I've updated code and made some changes:
1.    Move `JSONAPI` errors generation into `ValidationErrors` class;
2.    Return `422` status code instead of `400` when getting validation errors;
3.    Validation error path is now `"/my_attribute"`;
4.    Preface: my test in `controller_test` `test_create_with_invalid_data` generates a post with 2 validation   errors. So the `ValidationError` exception is raised and it has 2-element array in `errors`. And because of the size of `errors` I get `400` in response instead of `422`!
   
   I assumed that all errors in exception must have uniform status and that there can be only one `JSONAPI::Exceptions::Error` raise while processing request. So I always return the **first** error status.

lgebhardt:
@sineed Thanks.

Ideally, we would check if all the codes are the same and if they are return the first one. If there are multiple codes I think we would be better off with a more general 400. That was the logic I was using when I wrote the initial code. I had intended to implement the check to see if the codes were all the same, but never got around to it. However I think returning the first error code is probably fine in practice. It will probably be correct more often than always returning 400 if there are multiple errors.

sineed:
@lgebhardt I think there can be only one `JSONAPI::Exceptions::Error` object per one request when we're dealing with `AR` objects:
- On `Operation` level I think there is one `Error` ( I hope that there is no place in code where  `JSONAPI::Exceptions::Error` generates another  `JSONAPI::Exceptions::Error`);
- On `ActiveRecordOperationProcessor` level there is also one `Error` because when error occurs the processing flow stops (rollback exception is raised).

I think that we should determine that errors are the same [here](https://github.com/cerebris/jsonapi-resources/blob/master/lib/jsonapi/resource_controller.rb#L166) and maybe [here](https://github.com/cerebris/jsonapi-resources/blob/master/lib/jsonapi/resource_controller.rb#L146) to support non `AR` objects.

lgebhardt:
Thanks. It would be good to check the errors for consistency, but in the AR case it shouldn't be an issue.

PR #44, opened 2014-11-25
Fix handling singleton resources without ID given in the URL

wijet:
This fixes case when we have a singleton resource without ID given like `/account` which returns info for current logged in account. Without this fix it was raising exception `split` called on `nil`

btw. name `Preferences` is a bit unfortunate, it causes issues because it's uncountable. I could rename this model to `Account` everywhere or leave it like this, with configuring `ActiveSupport::Inflector` properly.

### DIFF
--- a/lib/jsonapi/resource_controller.rb
+++ b/lib/jsonapi/resource_controller.rb
@@ -120,7 +120,8 @@ module JSONAPI
     end
 
     def parse_key_array(raw)
-      return resource_klass.verify_keys(raw.split(/,/), context)
+      keys = raw.nil? || raw.empty? ? [] : raw.split(',')
+      resource_klass.verify_keys(keys, context)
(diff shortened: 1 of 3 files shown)


lgebhardt:
Looks good, but https://github.com/BookingSync/jsonapi-resources/blob/master/lib/jsonapi/routing_ext.rb#L19 is no longer covered by the testing. Removing the has_many association from the preferences seems to have done that. So we need another way to test that.

You can generate a coverage report by running `COVERAGE=true rake test`. This will be automatic soon.

wijet:
@lgebhardt I've added that has_many association back. Thanks for noticing that. No test failed and there was no actual association on the model so I've removed it :smile:

lgebhardt:
@wijet thanks for fixing this.

PR #46, opened 2014-12-01
Add a hook to access the underlying Active Record relation

whatthewhat:
[One of the examples](https://github.com/cerebris/jsonapi-resources#finders) in the readme shows overriding the `find` method to achieve scoping resources to current user:

``` ruby
def self.find(attrs, options = {})
  ...
  authors = context.current_user.find_authors(attrs)
  ...
end
```

I believe this a very common use case in API's. The problem is that for this to really work one must also override `find_by_key` and `find_by_keys`.

One way to make this easier is to provide a hook to override the Active Record relation:

``` ruby
  attribute :id, :title, :body

  # instead of overriding find, find_by_key and find_by_keys
  def self.relation(options = {})
    context = options[:context]
    context.current_user.posts
  end
end
```

This could also be used for something like pagination:

``` ruby
  attribute :id, :title, :body

  def self.relation(options = {})
    context = options[:context]
    _model_class.paginate(page: context.params[:page])
  end
end
```

_p.s.:_ The implementation is just a quick example and would need tests and documentation (and maybe a better name for the hook) if you think this is a viable approach.

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,7 +260,7 @@ end
 
 ##### Finders
 
-Basic finding by filters is supported by resources. However if you have more complex requirements for finding you can override the `find` and `find_by_key` methods on the resource.
+Basic finding by filters is supported by resources. However if you have more complex requirements for finding you can override the `find`, `find_by_key` and `find_by_keys` methods on the resource class.
 
 Here's an example that defers the `find` operation to a `current_user` set on the `context` option:
 
@@ -279,6 +279,23 @@ class AuthorResource < JSONAPI::Resource
     return authors.map do |author|
       self.new(author)
     end
+  end
+end
+```
+
+##### Customizing base records for finder methods
+
+If you need to change the base records on which `find`, `find_by_key` and `find_by_keys` operate, you can override the `records` method on the resource class.
+
+For example to allow a user to only retrieve his own posts you can do the following:
+
+```ruby
+class PostResource < JSONAPI::Resource
+  attribute :id, :title, :body
+
+  def self.records(options = {})
+    context = options[:context]
+    context.current_user.posts
   end
 end
 ```
--- a/lib/jsonapi/resource.rb
+++ b/lib/jsonapi/resource.rb
@@ -239,7 +239,7 @@ module JSONAPI
 
         resources = []
         order_options = construct_order_options(sort_params)
-        _model_class.where(where_filters).order(order_options).includes(includes).each do |model|
+        records(options).where(where_filters).order(order_options).includes(includes).each do |model|
           resources.push self.new(model, context)
         end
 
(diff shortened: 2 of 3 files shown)


whatthewhat:
@lgebhardt Could you please take a look at the tests I've added in https://github.com/cerebris/jsonapi-resources/commit/38dfafdcf932c56530c07baebdd031e794ad92a8?
I'm much more used to rspec.. :smile:

lgebhardt:
@whatthewhat The tests look fine to me. Thanks. Let me know if you plan any more changes.

whatthewhat:
@lgebhardt I'm still not sure about the name `relation`, but I guess it's fine..
Do you think this needs documentation?

lgebhardt:
I'm not sure on a better name than `relation`, but I'll give it some thought. Yes, we'll need to document it.

whatthewhat:
Added some docs and squashed the commits

opsb:
Seeing as the thing that is being represented is the resource how about just calling the method "resource"?

whatthewhat:
> Seeing as the thing that is being represented is the resource how about just calling the method "resource"?

Well.. it's more of a "collection of resources that is going to be filtered and sorted and is also happened to be an Active Record relation" :smile:

lgebhardt:
I see it more as the base Arel object/expression.

whatthewhat:
`base_relation`? I'm not sure about mentioning Arel in the method name..

lgebhardt:
> I'm not sure about mentioning Arel in the method name..

I agree. Maybe `active_record_base_resource`? It's long, but descriptive. Could also be shortened to `ar_base_resource`. Still, I'm not sure mentioning active record in the name is much better than Arel.

whatthewhat:
I'm already starting to think that "relation" was not so bad :)

Let's just choose one and be over with it:
- relation
- resource
- base_relation
- base_resource
- active_record_base_resource

opsb:
I suppose I was thinking about the name from the http end, i.e. what's the resource that you're representing? "Relation" is pretty clear though. I'd steer clear of a name including anything to do with active_record as I could see jsonapi-resources supporting other backends.

dgeb:
Since this method will return instances of the model class, maybe we should go with the ActiveRecord-agnostic `records`. These records have not yet been translated into "resources", which I think is an important distinction. Perhaps `records` should have a prefix / suffix to indicate that it _could be_ based on the context or other pre-processing _could be_ done, but I'm not sure this is necessary.

opsb:
@dgeb "records" gets my vote, I like the distinction between the unprocessed records and the resource(s) that are returned.

lgebhardt:
@dgeb :+1: on `records`.

whatthewhat:
`records` looks good! Updated and rebased.

lgebhardt:
Thanks!

PR #49, opened 2014-12-05
Serialize empty resource collections

lgebhardt:
A fix for #48.

### DIFF
--- a/README.md
+++ b/README.md
@@ -345,12 +345,14 @@ These codes can be customized in your app by creating an initializer to override
 
 ### Serializer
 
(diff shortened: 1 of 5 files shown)


whatthewhat:
:+1: 

There are some places in the readme that need to be updated (https://github.com/cerebris/jsonapi-resources/blob/master/README.md#serializer)

dgeb:
I guess I'm on the fence about initialization of the serializer with the resource class, but I can't offer a better alternative now. And we need this for #48, so let's :shipit:

PR #53, opened 2015-01-07
Apply filter

lgebhardt:
Adds apply_filter method.
Removes value coercion for DefaultValueFormatter.


dgeb:
Seems good! :shipit:

Issue #55, opened 2015-01-09
Should id values be formatted as strings by default?


barelyknown:
The [json:api specification](http://jsonapi.org/format/#document-structure-individual-resource-representations) says that the value of the `id` should be a string, but `jsonapi-resources` is not calling `to_s` on it by default. Should it?

lgebhardt:
Yes, at this point we should be calling to_s on ids in order to be compliant with JSON API. 

I think the spec should be changed to allow integer ids, since JSON supports integers. However until this is changed you are correct that `jsonapi-resources` is not compliant. I'll see about getting this fixed soon.

barelyknown:
I plan on using this library for a number of projects, so I'd be happy to make that change if you'd like. Should we add the feature through a configuration flag, and if so, should we make the current behavior the default, or the compliant behavior the default?

lgebhardt:
@barelyknown if you want to fix this that would be great. I think we should just go for compliance, so no configuration option is needed. If the spec changes we'll just go with that.

barelyknown:
Sounds good. PR on its way later today.

barelyknown:
Thanks for merging!

PR #56, opened 2015-01-09
Format ids as strings to conform to specification.

barelyknown:
Addresses issue https://github.com/cerebris/jsonapi-resources/issues/55.

From the JSON API specification:

> An individual resource SHOULD be represented as a ... string value containing its ID...


lgebhardt:
Looks good. Thanks!

PR #60, opened 2015-01-20
Fix resource class lookup for namespaced resources.

barelyknown:
The `jsonapi_resources` method was not getting the namespaced resource. That wasn't showing as an error because there weren't any resources that were only exposed within a namespace (and so it was looking up the non-namespaced resource). This fixes that.


lgebhardt:
Thanks for catching that!

PR #64, opened 2015-01-22
Use resource type with module prefix in all parts of the routing_ext.

barelyknown:


lgebhardt:
Thanks. I just bumped into this issue last night.

PR #69, opened 2015-01-28
Enforce application/vnd.api+json Content-Type header on Create and Update requests

kellysutton:
Per the JSON API spec: http://jsonapi.org/format/#crud

> Any requests that contain content **MUST** include a Content-Type header whose value is `application/vnd.api+json`.

Currently, using a `JSONAPI::ResourceController` routed via the `jsonapi_resources` route directive will return a cryptic error message with an `HTTP 400`, when the Content-Type header is set correctly to `application/vnd.api+json`. The error message complains that the required root key parameter is missing.

**Expected result:**

JSONAPI::ResourceController should reject any creation or update with an `HTTP 415 Unsupported Media Type` when the Content-Type header of the request is not set to `application/vnd.api+json`


kellysutton:
**Please Note:** This introduces a breaking change, and as such, the semver should be bumped to 0.2.0 after this has been merged in.

A few points of discussion:
- Please doublecheck my work in writing tests for this. I'm more familiar with RSpec, and I realize the test code is not very DRY. Any tips appreciated.
- I register a new `Mime::Type` with the alias of `:api_json`. 
- This code does not enforce the `application/json-patch+json` media type when issuing a `PATCH` request.

kellysutton:
Noted. I'll take another pass at this and push some updates. Other than those two things, any other corrections?

lgebhardt:
I'm still going through it. I don't see where the responses are setting the CONTENT_TYPE. I think this would need to be done in the `index`, `show`, `process_request_operations`, and `render_errors` methods. In addition we should be setting the ACCEPT header.

As an aside, I started work on this too (it's in the https://github.com/cerebris/jsonapi-resources/tree/jsonapi_content_type branch). I think your work is more complete, but you could look at my old code if you wish.

kellysutton:
Ah, I should have clarified. This does not set the `Content-Type` on responses. It only enforces the `Content-Type` on payloads coming from a client, that is: `create` and `update` actions.

lgebhardt:
Ok, we can certainly add the response headers in a separate PR.

kellysutton:
Alright, just pushed the latest work to enforce the `application/vnd.api+json` media type on `:create_association` and `:update_association` requests. Once the CI tests pass, this should be good to go.

After reading the spec, we do not need to enforce the media type on `DELETE` requests, because they do not have any content.

I'll open a separate PR for including the correct `Content-Type` in responses from `ResourceController` instances.

lgebhardt:
Yes, that makes sense to leave `DELETE` off the list. Someone can add it in if their controller needs additional params for a delete.

lgebhardt:
Ok, I've rethought my comment about the filter name and decided I like the name you've given it. Sorry for that confusion, and thanks for the work!

PR #73, opened 2015-01-31
Specify the correct Content-Type in the response headers

kellysutton:
Currently, all response headers generated by `ResourceController` instances specify the `application/json` Content-Type header. While not explicitly declared in the spec, the examples imply that that all responses with payloads should specify the `application/vnd.api+json` Content-Type response header.

Expect result:

All `ResourceController` responses with data should include the `application/vnd.api+json` Content-Type header.

### DIFF
--- a/lib/jsonapi/resource_controller.rb
+++ b/lib/jsonapi/resource_controller.rb
@@ -15,7 +15,7 @@ module JSONAPI
 
     before_filter :ensure_correct_media_type, only: [:create, :update, :create_association, :update_association]
     before_filter :setup_request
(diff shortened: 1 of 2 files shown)


kellysutton:
Actually, after doing some reading, I believe we should be setting the response Content-Type in an `after_filter`, not a `before_filter`. I'll make the correct modifications. 

In the case of a response without any content, we should likely also set the `Content-Type` to `text/plain`. Otherwise, I believe a client will attempt to parse the empty string, which is not valid JSON. I might be wrong there, though.

kellysutton:
Alright, changes made. This should be good to go.

lgebhardt:
Merged. Thanks!

PR #92, opened 2015-02-26
Remove support fetching multiple resources by id

### LINKED ISSUE #88: Remove support fetching multiple resources by id
JSON API RC2 removed support for GETting multiple resources by a comma delimited list of ids. We should remove this from JR as well.

sineed:
This PR resolves https://github.com/cerebris/jsonapi-resources/issues/88

@lgebhardt I have doubts about assertions of some tests.
For example, the test `test_tags_show_multiple_with_include` returns `200` only because of such behavior of `ActiveRecord` finder methods.

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,6 +166,7 @@ The system will lookup a value formatter named `DateWithTimezoneValueFormatter` 
 #### Primary Key
 
 Resources are always represented using a key of `id`. If the underlying model does not use `id` as the primary key you can use the `primary_key` method to tell the resource which field on the model to use as the primary key. Note: this doesn't have to be the actual primary key of the model. For example you may wish to use integers internally and a different scheme publicly.
+By default only integer values are allowed for primary key. To change this behavior you can override `verify_key` class method:
 
 ```ruby
 class CurrencyResource < JSONAPI::Resource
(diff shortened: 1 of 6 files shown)


lgebhardt:
@sineed, the assertions do seem a bit sketchy. I'm not sure how to best address that. I wonder if we should modify the `verify_key` method to assume and check for integer keys by default.

sineed:
@lgebhardt I've made some changes. Please take a look if I understood you correctly

lgebhardt:
That's what I had in mind. I think that's better than the way it was, though I wonder if there are any systems in common use that use strings for ids that we will break. UUIDs come to mind, though in my code where we use them I've already overridden the `verify_key` to enforce the format of UUIDs.

I want to leave this open for a day or two and see if anyone else has an opinion.

sineed:
@lgebhardt agreed, the change can break resources with custom  `primary_key` and lack of custom logic in `verify_key`. So I think that it's necessary to mention in README.
Also rebased and squashed.

dgeb:
I think this is fine for now, although I'd like to eventually add the concept of a "key type", which could then be referenced by `verify_key`. This would allow for cleaner and more consistent handling of UUIDs and custom key types.

lgebhardt:
@sineed Thanks!

@dgeb I'll put an issue in for the "key type" concept.

PR #95, opened 2015-03-04
Update README to reflect change in default key formatting.

adamworrall:
Updates to the readme. People incorrectly thinking the default key option is still `:underscored_key` will get param not allowed errors until they check this.

Added language reflecting JSON API naming recommendations, I hope this accurately reflects the reason for the change in defaults?


lgebhardt:
Looks good. Thanks!

Issue #96, opened 2015-03-04
Unable to override apply_filter


achan:
I highly suspect this is a problem on my side but I just want to confirm as well as see if you have any insight on what could be wrong.

I'm trying to implement filtering for one of my resources like so:

``` ruby
+  filter :latest
+
+  def apply_filter(records, filter, value)
+    raise "ERROR"
+  end
```

 https://github.com/openmcac/basechurch/pull/18/files

But my resource's `#apply_filter` never seems to be called. I've put breakpoints in `Resource#find` and have verified that, at the point that it's supposed to call `apply_filter`, `self` is my subclass... However, when it's fired, it is only calling `Resource#apply_filter` and my method is never called nor is "ERROR" ever raised.

In fact, it seems that I can't override **anything** from `Resource`.

Calling the method directly from my specs, does raise the runtime error though.

Are there any public examples of `apply_filter` working?

sineed:
@achan it is a class method so to override you must declare _self._apply_filter

achan:
Thanks @sineed, I sent a pull request to update the example in the README.

PR #97, opened 2015-03-04
Show the correct way to override apply_filter

achan:
Updated README showing the proper way to override `apply_filter`. See #96.


lgebhardt:
Thanks!

PR #98, opened 2015-03-05
Update README to mention where to find client libraries

jsmestad:
@lgebhardt  Feel free to kill off the Ruby bit if you feel its not necessary. We use `jsonapi_consumer` to consume the output of this gem so it may be worth calling out? I know personally when starting with `jsonapi-resources`, I was struggling to find a good ruby client to use even for testing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,10 @@ JR is designed to work with Rails, and provides custom routes, controllers, and 
 
 We have a simple demo app, called [Peeps](https://github.com/cerebris/peeps), available to show how JR is used.
 
+## Client Libraries
+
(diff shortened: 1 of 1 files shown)


dgeb:
@jsmestad please remove the individual recommendation, since we don't want to play favorites here.

jsmestad:
@dgeb no problem. updated.

dgeb:
Merged via d8818883a1ebe46530e5277e6caba095e5ea4f46 and a829c06ad46d485c9b6e35414c7f20ec9de889c0

Thanks @jsmestad!

PR #99, opened 2015-03-08
Fix issue with route names using the association name instead of the class name

lgebhardt:
Routes were being created using resources based on the association name, instead of the class name for the association. This broke associations that used a different class name from the association name, for example the post.author association references the people resource, not the author resource.

### DIFF
--- a/test/integration/routes/routes_test.rb
+++ b/test/integration/routes/routes_test.rb
@@ -69,11 +69,6 @@ class RoutesTest < ActionDispatch::IntegrationTest
   end
 
   # V2
-  # def test_routing_v2_posts_show
-  #   assert_routing({path: '/api/v2/authors/1', method: :get},
-  #                  {action: 'show', controller: 'api/v2/authors', id: '1'})
-  # end
-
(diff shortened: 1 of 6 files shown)


barelyknown:
I saw this bug and hadn't submitted the issue yet. This fixes it for me. Nice catch.

test/integration/routes/routes_test.rb
```diff
@@ -85,10 +69,10 @@ def test_routing_v1_posts_links_writer_show
...
 
   # V2
-  def test_routing_v2_posts_show
-    assert_routing({path: '/api/v2/authors/1', method: :get},
-                   {action: 'show', controller: 'api/v2/authors', id: '1'})
-  end
+  # def test_routing_v2_posts_show
+  #   assert_routing({path: '/api/v2/authors/1', method: :get},
+  #                  {action: 'show', controller: 'api/v2/authors', id: '1'})
+  # end
```
dgeb: Why is this commented out? Let's remove it or add a TODO if necessary.

Issue #85, opened 2015-02-20
Support for PUTting [] to to-many relationship URLs


lgebhardt:
JR should support PUTting an empty array to a to-many relationship URL to clear the associated records.

Note, this is missing from the current spec at http://jsonapi.org/format/#crud-updating-to-many-relationships, however it will be added soon.

``` text
PUT /articles/1/links/tags
Content-Type: application/vnd.api+json
Accept: application/vnd.api+json

{
  "data": []
}
```

PR #105, opened 2015-03-09
Support for PUTting [] to to-many relationship URLs

sineed:
Related to #85 

@lgebhardt it may seem strange, but this feature is already in the source code. I've only added a test to ensure.

I can't find an issue or PR in `json-api` gem about adding this feature in specification. Can you add a reference to #85 or in this PR?


lgebhardt:
Thanks for the test. I knew we weren't testing for it, but I wasn't sure about the implementation.

Issue #108, opened 2015-03-10
Extracting the loading of source and related resources in controller actions


barelyknown:
I'd like to ensure that a user is authorized to view a related resource before it is rendered, but there's no way currently without completely reimplementing the `get_related_resource` method.

Here's the current implementation of `get_related_resource` in `JSONAPI::ResourceController`:

``` ruby
def get_related_resource
  association_type = params[:association]
  source_resource = @request.source_klass.find_by_key(@request.source_id, context: context)

  serializer = JSONAPI::ResourceSerializer.new(@request.source_klass,
                                               include: @request.include,
                                               fields: @request.fields,
                                               base_url: base_url,
                                               key_formatter: key_formatter,
                                               route_formatter: route_formatter)

  render json: serializer.serialize_to_hash(source_resource.send(association_type))
end
```

Ideally, we would extract two methods that could be overridden without needing to completely reimplement this method. Something like:

``` ruby
def get_related_resource
  source_resource = load_source_resource(@request)
  related_resource = load_related_resource(source_resource, params[:association])
  serializer = JSONAPI::ResourceSerializer.new(@request.source_klass,
                                               include: @request.include,
                                               fields: @request.fields,
                                               base_url: base_url,
                                               key_formatter: key_formatter,
                                               route_formatter: route_formatter)

  render json: serializer.serialize_to_hash(related_resource)
end

def load_source_resource(request)
  # ...
end

def load_related_resource(source_resource, association_name)
  # ...
end
```

With these methods extracted, users of the library would have the option of authorizing the source and related resources before they're rendered. Something like:

``` ruby
def load_related_resource(source_resource, association_name)
  related_resource = source_resource.public_send(association_name)
  unless authorized?(context[:current_user], related_resource.model)
    raise NotAuthorizedError
  end
  related_resource
end
```

I could also imagine handling this requirement by adding a new callback (`before_render` maybe?). But I think that I prefer the method that I mentioned above.

For what it's worth, a similar problem exists with other controller actions too (some can be handled with the existing callbacks, some can't). We should use similar approaches for each action.

I'd be happy to write a PR for this tomorrow, but I wanted to discuss this issue and approach before adding the feature.

lgebhardt:
[...]

PR #109, opened 2015-03-10
Add associated_records method to enable override.

barelyknown:
This implements the idea that we discussed in issue https://github.com/cerebris/jsonapi-resources/issues/108. I made the `associated_records` method similar to the `records` method as @lgebhardt suggested and I think that it works nicely and is clear.

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,13 +283,29 @@ class PostResource < JSONAPI::Resource
 end
 ```
 
-You can also customize how associated records are found by overriding the `associated_records` method in the resource class.
-
-For example, raise an error if the user is not authorized to view the association records.
+When you create a relationship, a method is created to fetch record(s) for that relationship. This method calls `records_for(association_name)` by default.
 
 ```ruby
 class PostResource < JSONAPI::Resource
-  def associated_records(association_name, options={})
+  has_one :author
+  has_many :comments
+
+  # def record_for_author(options = {})
+  #   records_for("author", options)
+  # end
+
+  # def records_for_comments(options = {})
+  #   records_for("comments", options)
+  # end
+end
+
+```
+
+For example, you may want raise an error if the user is not authorized to view the associated records.
+
+```ruby
+class BaseResource < JSONAPI::Resource
+  def records_for(association_name, options={})
     context = options[:context]
     records = model.public_send(association_name)
 
--- a/README.md
+++ b/README.md
@@ -283,6 +283,25 @@ class PostResource < JSONAPI::Resource
 end
 ```
 
+You can also customize how associated records are found by overriding the `associated_records` method in the resource class.
+
+For example, raise an error if the user is not authorized to view the association records.
+
+```ruby
+class PostResource < JSONAPI::Resource
+  def associated_records(association_name, options={})
+    context = options[:context]
+    records = model.public_send(association_name)
+
+    unless context.current_user.can_view?(records)
+      raise NotAuthorizedError
+    end
+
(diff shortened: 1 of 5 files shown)


lgebhardt:
A single `associated_records` could become unwieldy with a lot of associations, all with different permissions systems. I'd rather we went with one method for each association. That could be created in the `_associate` method only if it isn't already defined in the resource. The only downside I see is the user would need to know to create the method with a certain name, so documentation is important.

I guess another option is to allow the user to specify the name when the association is defined.

barelyknown:
It seems more unwieldy to me to have one method for each association in most situations. I would guess that the most common case would be something similar to what I'm doing:

``` ruby
  def associated_records(association_name, options = {})
    associated_records = model.public_send(association_name)
    current_user = options[:context][:current_user]
    controller = options[:context][:controller]
    permission = controller.action_name + "?"

    case associated_records.class
    when ActiveRecord::Relation, Array
      associated_records = Pundit.policy_scope!(current_user, associated_records)
      controller.policy_scoped!
    when NilClass
      controller.policy_authorized!
    else
      policy = Pundit.policy!(current_user, associated_records)
      unless policy.public_send(permission)
        error = Pundit::NotAuthorizedError.new("not allowed to #{permission} this #{associated_records}")
        error.query, error.record, error.policy = permission, associated_records, policy
        raise error
      end
      controller.policy_authorized!
    end
    associated_records
  end
end
```

If you need to implement a different approach for a resource class or for a relationship in a resource, you could override at a lower level.

Finally, I'd like to limit the metaprogramming the definition of methods where we can. It's fine where it's needed, but I don't think that this situation warrants it.

dgeb:
Since association and attribute values can already be overridden with meta-programming, I think @lgebhardt's suggestion is consistent.

For example:

``` ruby
class AuthorResource < JSONAPI::Resource
  attribute :id, :name

  has_many :posts

  def name(options)
    @model.name.capitalize
  end

  def posts(options)
    // return sorted, filtered, paginated posts
  end

  def posts_records(options)
    // return raw set of models
  end
end
```

How about chaining both approaches, wherein the meta-method (`posts_records` in this case) calls `associated_records` by default.

barelyknown:
I think that having the meta-method return the `associated_records` by default would be ok. I still don't exactly see the need to generate the meta-methods TBH, but as long as there is a default `associated_records` and you see it as important, it's fine with me.

barelyknown:
If we're good with that approach, I can update the PR. Let me know.

lgebhardt:
I'm good with that. Thanks, @barelyknown

barelyknown:
For `has_one` relationships, do you think that the method should be `post_record` or `post_records` (for example)? `post_record` feels right to me.

lgebhardt:
I agree the singular feels right. However it is still returning an ActiveRecord relation, same as the has_many, so maybe we should still go with the plural? @dgeb any thoughts?

dgeb:
How about `records_for_posts` (has-many) and `record_for_author` (has-one)?

lgebhardt:
:+1:

barelyknown:
I implemented our decision and renamed `associated_records` to `records_for` which is good and more consistent with the `records` method name. Good suggestions. This turned out right.

dgeb:
@barelyknown thanks for following through! Glad those naming suggestions worked for you. And I agree that `records_for` is nice and consistent.

lgebhardt:
Looks good. Thanks!

PR #114, opened 2015-03-10
Put top-level library file where Bundler expects

nwjsmith:
This simplifies the installation experience and removes the need to
explicitly `require` everywhere.


lgebhardt:
Before your changes were you needing to add the "require 'jsonapi/resource'" lines? I thought I had that fixed, though maybe not per the bundler standard. It does look like the readme didn't get cleaned up, but I'm not sure on the other changes.

fotinakis:
I just added jsonapi-resources to my Rails app and autoloading works fine for me without this—I don't have to add any manual require lines with current master (I _did_ have to with the current out-of-date gem release).

lgebhardt:
@fotinakis Thanks for confirming that it's good in master. I'm going to close this for now, and we can reopen it if that turns out to be premature.

PR #116, opened 2015-03-12
Fix included relationships when relationship is not direct.

barelyknown:
If you include a resource through a relationship and the included resource is not a direct association, the request fails with:

``` json
{
  "errors":[
    {
      "title":"Invalid field",
      "detail":"preferences is not a valid association of people",
      "id":null,
      "href":null,
      "code":112,
      "path":null,
      "links":null,
      "status":"bad_request"
    }
  ]
}
```

According to the spec, this should work fine.

> In order to request resources related to other resources, a dot-separated path for each relationship name can be specified:
> 
> `GET /articles/1?include=comments.author`

I added a test that shows that the `preferences` relationship can be included from its author, and another test that fails when you try to include the `preferences` for a post through `author.preferences`.

Before I wrote a fix, I wanted to run this by the team to see if you're surprised by this behavior or if you have an opinion about how the fix should be implemented.

### DIFF
--- a/test/controllers/controller_test.rb
+++ b/test/controllers/controller_test.rb
@@ -1629,7 +1629,6 @@ end
 class Api::V5::PostsControllerTest < ActionController::TestCase
   def test_limit_fields_in_included_resource_that_is_not_a_direct_relationship
     get :show, { id: "1", include: "author.preferences" }
-    puts response.body
     assert_equal response.status, 200
   end
 end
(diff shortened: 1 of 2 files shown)


barelyknown:
I've seen this not fail in some cases in a real app, but I haven't figured out yet how those situations are different than this one. I originally found this problem when I noticed that the "fields" functionality wasn't working correctly on associations of associations.

barelyknown:
I verified that this was fixed by PR https://github.com/cerebris/jsonapi-resources/pull/237. 👍🏻👍🏻

Issue #120, opened 2015-03-17
Request.parse_sort_criteria may not handle literal + in sort params


realityendshere:
I opened an equivalent issue on the JSON API spec project: https://github.com/json-api/json-api/issues/486

If I am not mistaken, many web servers see the plus symbol ("+") as a substitute for a white space or "%20" when sent as a query param in the URL. For example, this request

```
GET /people?sort=+age,+name
```

shows up as `" age, name"` to the Request.parse_sort_criteria method. Which is a total bummer for the `if` statement

``` ruby
if sort.start_with?('+')
  sort_criteria[:direction] = :asc
elsif sort.start_with?('-')
  sort_criteria[:direction] = :desc
else
  @errors.concat(...)
end
```

The spec compliant GET request leads to a 400 error because `" age"` and `" name"` both start with a white space instead of a `"+"`!

I find I can address this by either allowing a leading white space to behave as a + (e.g. `sort.start_with?(' ')`) or using `%2b` to represent `+` in the query params and using something like `sort_criteria = URI.unescape(sort_criteria)` to decode `%2b` into `+`.

Is there a preference for using one, the other, or both?

vrybas:
@realityendshere can be closed :)

realityendshere:
No longer applicable since JSON API dropped "+" in ascending sort in https://github.com/json-api/json-api/commit/8c63a72b42869835271df63800d980ad4989e005

PR #121, opened 2015-03-17
Fix sort param plus symbol handling [Issue #120]

realityendshere:
Many web servers substitute white spaces for "+". This causes

```
GET /people?sort=+age,+name
```

to reach the parse_sort_criteria method as `" age, name"` and throw an error.

This change allows the parse_sort_criteria method to accept the literal '+', a leading white space ' ', or an encoded plus '%2b' as valid sort ascending indicators.

Two really simple additional tests are included.

See #120 and https://github.com/json-api/json-api/issues/486


lgebhardt:
Thanks.

PR #131, opened 2015-03-27
Refactor #process_request_operations to be a bit easier to follow.

adamworrall:
I was reading through the Controller code and found this method to be a bit confusing as written. I've tried to tighten it up a bit.


adamworrall:
Accidentally submitted from master branch, will resubmit from feature.

PR #132, opened 2015-03-27
Refactor process operations

adamworrall:
I was reading through the Controller code and found this method to be a bit confusing as written. I've tried to tighten it up a bit.


lgebhardt:
Thanks. That is cleaner.

PR #136, opened 2015-03-28
Remember to require 'jsonapi/resource_controller'

spundun:
One less error for new users to deal with. you have to explicitly require jsonapi controller when deriving from it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -487,6 +487,8 @@ Callbacks can also be defined for `JSONAPI::OperationsProcessor` events:
 For example:
 
 ```ruby
+require 'jsonapi/resource_controller'
+
 class PeopleController < JSONAPI::ResourceController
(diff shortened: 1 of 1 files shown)


lgebhardt:
What version were you testing with? On the latest, 0.3.0.pre1 I fixed the missing require for the controller in `lib/jsonapi-resources.rb`. So I don't think we need this change.

Sorry for the confusion. Please reopen this if that is not the case.

spundun:
I'm using 0.2. Should I use github snapshot instead?

spundun:
I updated my Gemfile to use your latest master, and that seems to fix the issue. Thanks :)

lgebhardt:
Glad to hear it, and thanks for confirming it.

PR #137, opened 2015-03-29
Respect formatting config when validating type property 

sirvine:
This is to correct various errors caused by comparison of the camelcase key vs snake formatted class names, and vice versa. Please note that I'm using `config.json_key_format = :camelized_key`.


sirvine:
Sorry!  Didn't mean to include my other PR in this one.  I'm going to close this and break them out.

PR #138, opened 2015-03-29
Respect formatting config when validating type property.

sirvine:
This is to correct various errors caused by comparison of the formatted key vs class names, and vice versa. 

Please note that I'm using `config.json_key_format = :camelized_key`.


sirvine:
Closing due to errant git push!

PR #139, opened 2015-03-29
Fixes for formatting of type property checks

sirvine:
This is to correct various errors caused by comparison of the formatted key vs class names, and vice versa.

Please note that I'm using config.json_key_format = :camelized_key.


lgebhardt:
@sirvine Which version was this failing in? I had made similar changes in 0.3.0pre1 and I'm wondering if this is still an issue.

sirvine:
I just bumped to 0.3.0pre2, and all is indeed well. This PR is superfluous now.

Issue #134, opened 2015-03-27
Trouble parsing has_one relationships


sirvine:
I'm getting error code 115 on all POST requests that have has_one relationships.  I've traced the issue to the following code from [here](https://github.com/cerebris/jsonapi-resources/blob/4b837118cb1e10612d21f709bb0d2be9c451a531/lib/jsonapi/request.rb#L239):

```
   if !raw.is_a?(Hash) || raw.length != 2 || !(raw.has_key?('type') && raw.has_key?('id'))
      raise JSONAPI::Exceptions::InvalidLinksObject.new
    end
```

Looking at my Rails logs, the inbound params for a has_one component is in the following form `{"linkage"=>{"id"=>"24", "type"=>"suppliers"}}`, which is failing the last three conditionals above.

Update:

I assume this is related to https://github.com/json-api/json-api/pull/447 and https://github.com/cerebris/jsonapi-resources/issues/115

fotinakis:
Just hit this too.

vrybas:
@sirvine Fixed by #140 ?

sirvine:
@vrybas I think so. It's been a year since I opened this, and I haven't encountered this issue since. Closing now.

Issue #115, opened 2015-03-12
Support update to resource linkage


dgeb:
See https://github.com/json-api/json-api/pull/447

_Note for the weary: this is probably the last breaking change for JSON API as we prepare for 1.0._

PR #140, opened 2015-03-29
Check for linkage key in associations [Issues #134 and #115]

sirvine:
Possible fix for https://github.com/cerebris/jsonapi-resources/issues/134 and https://github.com/cerebris/jsonapi-resources/issues/115.  Steps into "linkage" key, where available.

### DIFF
--- a/lib/jsonapi/request.rb
+++ b/lib/jsonapi/request.rb
@@ -233,37 +233,46 @@ module JSONAPI
           type: nil,
           id: nil
         }
-      elsif raw.is_a?(Hash) && raw.has_key?('linkage')
-        raw = raw['linkage']
-      end
-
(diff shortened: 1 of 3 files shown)


fotinakis:
+1 for fixing this, going to patch this in for myself locally for now.

lgebhardt:
Thanks for addressing this. However the spec is now requiring `linkage` for updates and creates. These changes do not enforce that new requirement, but rather supports both versions, which isn't correct.

Would you like to update this PR, including the tests, to enforce this? If not, I can get to it soon.

sirvine:
Great point. I'll work on this, although I'm no expert on the JSONAPI spec. Is there a changelog for the spec somewhere that I can use to enforce the changes that don't affect my app (yet)?

sirvine:
OK – updated tests, and added a check for a blank array in place of `linkage` key (i.e., where update is to clear linked objects).  Let me know if you want me to make any changes.

PR #144, opened 2015-04-02
Added a few request tests for has_many associations

slaskis:
I'm trying out your great gem of a JSONAPI implementation but I was a bit stumped while trying to create associations in my app so I've added a few failing tests in the hope that it might help me help you help me out with this.

It's just a few request tests for [this part of the spec](http://jsonapi.org/format/#crud-updating-to-many-relationships).

### DIFF
--- a/test/integration/requests/request_test.rb
+++ b/test/integration/requests/request_test.rb
@@ -158,7 +158,7 @@ class RequestTest < ActionDispatch::IntegrationTest
 
   def test_update_association_without_content_type
     ruby = Section.find_by(name: 'ruby')
-    patch '/posts/3/links/section', { 'sections' => {type: 'sections', id: ruby.id.to_s }}.to_json
+    patch '/posts/3/links/section', { 'data' => {type: 'sections', id: ruby.id.to_s }}.to_json
 
     assert_equal 415, status
   end
@@ -175,6 +175,27 @@ class RequestTest < ActionDispatch::IntegrationTest
     put '/posts/3/links/section', { 'data' => {type: 'sections', id: ruby.id.to_s }}.to_json, "CONTENT_TYPE" => JSONAPI::MEDIA_TYPE
 
     assert_equal 204, status
+  end
+
+  def test_patch_update_association_has_many
+    like = Comment.find_by(body: 'i liked it')
+    patch '/posts/3/links/comments', { 'data' => [{type: 'comments', id: like.id.to_s }]}.to_json, "CONTENT_TYPE" => JSONAPI::MEDIA_TYPE
+
+    assert_equal 204, status
+  end
+
+  def test_post_update_association_has_many
+    like = Comment.find_by(body: 'i liked it')
+    post '/posts/3/links/comments', { 'data' => [{type: 'comments', id: like.id.to_s }]}.to_json, "CONTENT_TYPE" => JSONAPI::MEDIA_TYPE
+    
+    assert_equal 204, status
+  end
+
+  def test_put_update_association_has_many
+    like = Comment.find_by(body: 'i liked it')
+    put '/posts/3/links/comments', { 'data' => [{type: 'comments', id: like.id.to_s }]}.to_json, "CONTENT_TYPE" => JSONAPI::MEDIA_TYPE
+
+    assert_equal 405, status
   end
 
   def test_index_content_type


lgebhardt:
The `comment` association is not allowed to be overwritten entirely by use of the `acts_as_set` directive on line /test/fixtures/active_record.rb:426

``` ruby
  attribute :title
  attribute :body
  attribute :subject

  has_one :section
  has_many :tags, acts_as_set: true
  has_many :comments, acts_as_set: false
 ...
end
```

See https://github.com/cerebris/jsonapi-resources#options

The http://jsonapi.org/format/#crud-updating-to-many-relationships section has changed a bit since this code was written. It looks a few changes are needed to bring things into compliance. I believe the `acts_as_set` option should only apply to `PATCH` from a quick read through.

slaskis:
Aha, I think I misunderstood that option. I read it as a [Set](http://ruby-doc.org/stdlib-2.2.1/libdoc/set/rdoc/Set.html) instead, as in unordered without duplicates. So I didn't take that into account.

But besides that. Would you, by your understanding of the current spec, agree that those tests should pass?

Because the above currently results in these failures:

```
  1) Failure:
RequestTest#test_put_update_association_has_many [/Users/robert/Work/Department/Adamsky/jsonapi-resources/test/integration/requests/request_test.rb:198]:
Expected: 405
  Actual: 404


  2) Failure:
RequestTest#test_patch_update_association_has_many [/Users/robert/Work/Department/Adamsky/jsonapi-resources/test/integration/requests/request_test.rb:184]:
Expected: 204
  Actual: 404


  3) Failure:
RequestTest#test_post_update_association_has_many [/Users/robert/Work/Department/Adamsky/jsonapi-resources/test/integration/requests/request_test.rb:191]:
Expected: 204
  Actual: 400
```

And it's in particular that last one (the POST giving a 400 error) that I'm having troubles with in my own app. I expect that syntax/request to be correct and that it would simply append that comment to the post.

The PUT I understand is not even part of the spec, which is why I expected a 405. And is the 404 from the PATCH because of `act_as_set`? Because a 404 would mean to me that a post does not have a comments relationship which seems invalid to me.

But every time I re-read the spec I find new things so I might just have missed these parts.

lgebhardt:
@slaskis I agree most of those tests should pass, and I created pr #146 that fixes the issues you brought up. One design decision I made in the app is that we will treat PUTs the same as PATCH since that's how Rails does it. So the first test you have should return a 204, just like the second.

If you try out #146 could you let me know if it fixes the issue in the last test that you are finding in your app?

slaskis:
Hmm, not sure why but I can't get the tests to run now. It just keeps complaining with: 

```
$ rake
rake aborted!
LoadError: cannot load such file -- rack/test
/Users/slaskis/Work/jsonapi-resources/test/test_helper.rb:14:in `<top (required)>'
/Users/slaskis/Work/jsonapi-resources/Rakefile:4:in `<top (required)>'
(See full trace by running task with --trace)
```

Even if I add it to the Gemfile, something I didn't need before. And looking at the diff I don't see why it would be needed now...

Nevermind, it's probably something wrong with my gems or ruby 2.1.5 install. When I switched to ruby 2.2.1 it worked fine.

slaskis:
Although running my app tests I get a 500 now. With this error:

```
JSONAPI: Could not find resource 'projects'. (Class ProjectResource not found)
```

And there is a ProjectResource in the app, but it's in a V1 module. But I'll try to find where the issue might lie...

slaskis:
Hmm, ok so I starting to think it's related to [these comments](https://github.com/cerebris/jsonapi-resources/blob/master/lib/jsonapi/resource.rb#L207-L209) where the lazy registered resources are stored with a module and these are stored without a module.

lgebhardt:
You mention that ProjectResource is a V1 module. If you have a resource in another module that's referencing the ProjectResource it won't find it unless there's also a ProjectResource in the same module. In other words all related resources must be in the same module. You can simply derive the resource from another to satisfy this, but the class must exist.

It could be related to the eager loading. I haven't seen that create problems either way, but I also haven't tested without it in a while. My guess is if you have a missing resource this setting will just determine when the error shows up.

slaskis:
Well I have all my resources in the V1 module. What I see if I log the `@@resource_types` is this:

```
{"v1/users"=>"V1::UserResource", 
 "v1/identity"=>"V1::IdentityResource", 
 "v1/project"=>"V1::ProjectResource", 
 "v1/projects"=>"V1::ProjectResource", 
 "v1/content"=>"V1::ContentResource", 
 "v1/identities"=>"V1::IdentityResource", 
 "v1/contents"=>"V1::ContentResource", 
 :users=>"UserResource",
 :identities=>"IdentityResource", 
 :projects=>"ProjectResource", 
 :contents=>"ContentResource"}
```

I've posted the resources [in a gist](https://gist.github.com/slaskis/8dd2ae0f92025ededaea) if that helps.

If I understand properly the ones created  "lazily" have "V1" and the ones created by `inherited(base)` have that stripped.

And I guess that could be fine but the `users` has_many relationship of `projects` won't currently find `V1::ProjectResource` even though both are in the V1 namespace. It does when GETing the projects but not when I try to POST to `/users/:id/links/projects`.

slaskis:
I tried setting eager_load to true as well for the test environment and I get the same error...

lgebhardt:
We shouldn't be seeing the non-namespaced resources in the list. I can confirm that I'm also seeing them, so it looks like a bug to me. I'll dig into this a bit.

slaskis:
Well even if I "fix" that resource_type map to do the same as in the inherited(base) method it still won't find the associated resource.

Could that perhaps be a scoping issue? 

That instead of running from the context of the `V1::UserResource` it is running from the context of `JSONAPI::Resource`?

slaskis:
Nvm, I just don't know my rails-fu. [safe_constantize](http://apidock.com/rails/v4.1.8/ActiveSupport/Inflector/safe_constantize) always work in top-level scope. 

Does that mean it should not demodularize the class_names?

lgebhardt:
I've add a new branch (https://github.com/cerebris/jsonapi-resources/tree/name_map_fix) that removes the code that was creating the incorrect entries in `@@resource_types`. Could you try this branch in your project for me.

You mentioned `safe_constantize`. The latest versions of the project only use `safe_constantize`, which makes me wonder which version you are using. Or maybe I'm misunderstanding you.

slaskis:
With that branch I got the same issue I had originally with the POST:

```
{"errors":[{"title":"Param not allowed","detail":"projects is not allowed.","id":null,"href":null,"code":105,"path":null,"links":null,"status":"bad_request"}]}
```

And when I rebased that onto the pr/144 branch I instead got this again:

```
JSONAPI: Could not find resource 'projects'. (Class ProjectResource not found)
```

So I added back the resource_types log right before that exception is and I get this:

```
{"v1/users"=>"V1::UserResource", "v1/identity"=>"V1::IdentityResource", "v1/project"=>"V1::ProjectResource", "v1/contents"=>"V1::ContentResource", "v1/projects"=>"V1::ProjectResource", "v1/content"=>"V1::ContentResource", "v1/identities"=>"V1::IdentityResource", :projects=>"ProjectResource"}
```

I'm guessing that last one was create by the last attempt. I've [gisted the full error stack](https://gist.github.com/slaskis/06eca816ca0de8831d19) if that helps.

slaskis:
And regarding the `safe_constantize` I mentioned, I was thinking if `class_name = "ProjectResource"` then `class_name.safe_constantize` would not work because it the resource is in the V1 module.

But the way it looks nice it will probably work better except I guess the module scope needs to be added somewhere because `has_many :x` only defines a type. Maybe in `_associate()`?

slaskis:
Allright, it's probably not the best way to do it but I've got my tests passing in [this branch](https://github.com/slaskis/jsonapi-resources/tree/module_path_prefix) which is your two branches merged with [this commit](https://github.com/slaskis/jsonapi-resources/commit/0920a9872efc1017813bcbe42be128d4c2fb8f5f). All the tests in this repo passes as well.

Now I'd just like to be able to figure out a failing tests for this first so we can catch some regressions.

lgebhardt:
I think you have found it with your new branch. I'm doing the same logic (adding in the module path) in https://github.com/cerebris/jsonapi-resources/blob/master/lib/jsonapi/resource.rb#L551, https://github.com/cerebris/jsonapi-resources/blob/master/lib/jsonapi/resource.rb#L560 and elsewhere. Looks like I forgot to account for the module path in the association.

I'm missing tests to exercise the create association logic under a namespaced resource.

Also, I don't think this has anything to do with the issue, but you no longer need to the line "require 'jsonapi/resource' " in your resources.

slaskis:
Oh that's nice, I was wondering why those were necessary. :)

Thank's a lot for your help. Looking forward to this becoming the de-facto implementation of JSONAPI.

lgebhardt:
And thanks for your help. Looking at this code again has me thinking there should be a clearer way to handle things. I'll look into a refactor for this.

PR #145, opened 2015-04-02
Expand routing section of  the README.

lgebhardt:


lgebhardt:
Failed Travis check is due to merging, in the middle of the runs, a PR that interacted with updated test data. This should be safe since it's only a README change.

PR #147, opened 2015-04-03
Spec compliance for null linkage for to-one relationships.

fotinakis:
I think I'm reading the spec correctly here, it says:

```
Resource linkage MUST be represented as one of the following:
- null for empty to-one relationships.
```

http://jsonapi.org/format/#document-structure-resource-relationships

But right now these relationships return empty hashes.


lgebhardt:
Yes, I read the spec the same way. Thanks for catching and fixing this.

Issue #151, opened 2015-04-07
Request spec (RSpec-Rails) returns "The required parameter, data, is missing."


jamonholmgren:
I'm trying to test my API with an RSpec-Rails request spec. Here's the spec:

``` ruby
    # logged_in_headers is defined above

    describe "POST /create" do
      it "creates a new campaign" do
        post "/api/v2/campaigns", { "data" => {} }, logged_in_headers
        expect(JSON.parse(response.body)).to eq({ not_sure: "what" })
      end
    end
```

I'm getting this error:

```
  1) /api/v2/campaigns logged in POST /create creates a new campaign
     Failure/Error: expect(JSON.parse(response.body)).to eq({ not_sure: "what" })

       expected: {:not_sure=>"what"}
            got: {"errors"=>[{"title"=>"Missing Parameter", "detail"=>"The required parameter, data, is missing.", "id"=>nil, "href"=>nil, "code"=>106, "path"=>nil, "links"=>nil, "status"=>"bad_request"}]}

       (compared using ==)

       Diff:
       @@ -1,2 +1,2 @@
       -:not_sure => "what"
       +"errors" => [{"title"=>"Missing Parameter", "detail"=>"The required parameter, data, is missing.", "id"=>nil, "href"=>nil, "code"=>106, "path"=>nil, "links"=>nil, "status"=>"bad_request"}]
```

It appears the params is not passing through when we're posting. Any ideas?

jamonholmgren:
Could be related to the fact I'm on Rails 3.2. See issue #152.

adamworrall:
To be a valid JSON-API request you need to at least have the "type" attribute set in the data object, so I believe your request should give an error of some kind.

There is something funny going on with blank data requests though - it seems like attributes whose values are empty are getting discarded altogether - I have a POST endpoint I'd like to work without requiring any parameters at all but I'm not able to get a "data": {"type": "tokens"} request to complete properly. It works fine if I add just an id. 

Additionally, I tried to see if I could replicate your issue and I succeeded on Rails 4.2. When I add a "type" attribute, though, it returns 201 and fires my after_create callbacks even though the model didn't save correctly. Seems like this case ignores the model validations?

lgebhardt:
The request is certainly invalid, but the error message doesn't look like what we should be returning in that case. However it's returned that way under Rails 4 as well. I'll look into what's going on soon.

lgebhardt:
[...]

PR #153, opened 2015-04-08
Type only post requests

adamworrall:
First shot at addressing POST where the data object only contains a "type" member.

Relates to discussion in #151. Created resources with only "type" in the data don't get @save_needed set, so they never actually save the ActiveRecord model and they don't get a change to generate validation errors, leading to incorrect 201 status.

I've added an explicit check to Resource#change to save if the object is_new, but that's ugly. Any thoughts on where this should be fixed?


jamonholmgren:
:+1:

PR #158, opened 2015-04-11
Fix error on column names with underscores.

yawboakye:
The allowed fields are formatted and collected in
`formatted_allowed_fields` array, which is used during the comparison.
During comparison though, keys from the `params[:data]` hash are not
formatted before verifying whether they are allowed. Thus keys with
underscores will always fail. This fix makes sure the comparison is done
using the formatted version of the keys.

### DIFF
--- a/lib/jsonapi/request.rb
+++ b/lib/jsonapi/request.rb
@@ -347,7 +347,7 @@ module JSONAPI
             params_not_allowed.push(links_key) unless formatted_allowed_fields.include?(links_key.to_sym)
           end
         else
-          params_not_allowed.push(key) unless formatted_allowed_fields.include?(key.to_sym)
+          params_not_allowed.push(key) unless formatted_allowed_fields.include?(format_key(key).to_sym)
         end
       end
       raise JSONAPI::Exceptions::ParametersNotAllowed.new(params_not_allowed) if params_not_allowed.length > 0


lgebhardt:
I think I should clean up the logic in `verify_permitted_params`, since I guessing it's confusing. It's comparing the keys in the request to the formatted version of the allowed keys in the Resource. Therefor it's not necessary to format the key in the request, since it should match one of the formatted resource key. 

I'm going to close this PR. If it's still an issue please open a new issue and we can discuss the details.

yawboakye:
@lgebhardt Is the keys in the request payload expected to be dasherized? Then that's fine. Otherwise, even though the elements of `formatted_allowed_fields` are ok, the keys in `params[:data]` may contain underscores during comparison (`key` in https://github.com/cerebris/jsonapi-resources/blob/master/lib/jsonapi/request.rb#L350 has underscores).

lgebhardt:
@yawboakye The `json_key_format` configuration option controls the format of the keys in the request payload. If it's set to `dasherized_key` then the keys in the payload are required to be dasherized.

I've added #168 to test this.

yawboakye:
@lgebhardt I made the wrong assumptions about the default `json_key_format` and it's effect on the request payload. Really sorry for bothering you.

lgebhardt:
@yawboakye No problem. This made me see I was missing some tests for the formatted keys, so that helps.

PR #159, opened 2015-04-14
Add ability to use scoped routes, and named routes

dphaener:
I had the need to use scoped and named routes in my API:

``` ruby
namespace 'api' do
  namespace 'v1' do
    scope ':brewery_id' do
      jsonapi_resources :recipes do
          jsonapi_resources :recipe_grains, path: :grains
          jsonapi_resources :recipe_hops, path: :hops
          jsonapi_resources :recipe_mash_rests, path: :mash_rests
        end
    end
  end
end
```

This pull request allows for the path option to work. 

For the scoping, just add:

```
def scope_id
  ...
end
```

To define your scope to your controller and you're good to go.


lgebhardt:
Thanks. Could you please squash your commits (or equivalent) so the .idea folder isn't in the commit log. Also, please make sure all tests pass and the project has 100% coverage. This can be done with: `COVERAGE=true rake test`

dphaener:
Build is running right now, but should be good to go.

lgebhardt:
Please add some tests to show how this will be used. Make sure you show how related resources references will be serialized. There are many options on the rails scope method. How does this change work with them? It would be good to add some documentation to the readme to cover the usage of this.

Why did you name the controller method `scope_id`?

Also, please squash your commits so the .idea directory addition and removal isn't in the change log. See http://stackoverflow.com/questions/5667884/how-to-squash-commits-in-git-after-they-have-been-pushed

dphaener:
> Please add some tests to show how this will be used. Make sure you show how related resources references will be serialized. There are many options on the rails scope method. How does this change work with them? It would be good to add some documentation to the readme to cover the usage of this.

Will do. I dug into the Rails implementation of `scope` in routing, and I think I will have to implement this differently. I really implemented it for my specific use case, but there are several other things to think about. I think this logic probably belongs in the route_ext.rb patch. I'll make sure to add tests for all the other use cases.

> Why did you name the controller method scope_id?

There really wasn't much of a reason, other than I was using an id to scope my routes. I'll rethink this with the refactor. Seems like someone could use a dasherized scope, among other things, so `scope_id` is not necessarily the best name.

> Also, please squash your commits so the .idea directory addition and removal isn't in the change log. See http://stackoverflow.com/questions/5667884/how-to-squash-commits-in-git-after-they-have-been-pushed

Pretty sure I already did that? The .idea commits aren't showing up in the commit history anymore as far as I can see.

maxhungry:
scoped route is nice, any update on this?

dphaener:
@maxhungry Just life getting in the way of me fixing this PR. Feel free to fork and hack away if you want! Thanks for the reminder though. I'll try to set some time aside to get this going again.

dphaener:
@thibaudgg @maxhungry @lgebhardt I'm going to have to punt on this. I am involved in way too many projects right now, and just looked at the code base and the master branch has diverged quite a bit since I last looked at this. Sorry everyone! I hope someone is brave enough to take over on it.

Issue #104, opened 2015-03-09
Key Type


lgebhardt:
Add a default configuration option for "key type", which can be overridden at the resource level.

The "key type" will be used by the `verify_key` method to check the key format. We should by default handle Integer, GUID and String formatted keys.

lgebhardt:
Closed through #420

PR #160, opened 2015-04-14
Key type config option

adamworrall:
Addresses #104. Allows users to set :integer, :uuid, :string, or custom lambda validation globally instead of overriding each resource. 

All current tests passing with minor mods to existing verify_key methods in the existing tests, it's working on my app as well - if you like this approach I can expand coverage to include the non-integer options. I noticed there is a resource, `IsoCurrencyResource`, in the test setup that was configured to verify String keys but I didn't see any tests that seem to use it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,18 +170,34 @@ The system will lookup a value formatter named `DateWithTimezoneValueFormatter` 
 #### Primary Key
 
 Resources are always represented using a key of `id`. If the underlying model does not use `id` as the primary key you can use the `primary_key` method to tell the resource which field on the model to use as the primary key. Note: this doesn't have to be the actual primary key of the model. For example you may wish to use integers internally and a different scheme publicly.
-By default only integer values are allowed for primary key. To change this behavior you can override `verify_key` class method:
+By default only integer values are allowed for primary key. To change this behavior you can set the `resource_key_type` in your initializer. For example, in your `config/initializers/jsonapi_resources.rb`:
 
 ```ruby
-class CurrencyResource < JSONAPI::Resource
-  primary_key :code
-  attributes :code, :name
+JSONAPI.configure do |config|
+  # Allowed values are :integer(default), :uuid, :string, or a proc
+  config.resource_key_type = :uuid
+end
+```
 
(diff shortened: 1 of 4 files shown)


john-griffin:
Thanks, tried this out locally and it fixes the errors I encountered when working with uuids on a resource. Good stuff!

lgebhardt:
Thanks, I like it! Yes, if you could expand the coverage that would be great.

Also, I'm using uuids in our app. In the routes I do this:

``` ruby
  UUID_regex ||= /[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}(,[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12})*/

  namespace :api, defaults: {format: 'json'}  do
    namespace :v1 do
      jsonapi_resources :posts, constraints: {:id => UUID_regex}
    end
  end
end
```

This allows the router to enforce the key format for uuids. I think it would be good to get this logic baked into the `jsonapi_resources` method by looking up the key type.

Also, in `verify_key`, I think we should we support uuids with uppercase hex values. Our regexs for uuids are also different in other ways. I'm not sure at this point which is correct.

adamworrall:
Great, having the routing constraints integrated too feels good. I'll try to have this ready this weekend.

gazsp:
Is this going to be merged in to master any time soon? Thanks.

adamworrall:
Sorry, I got busy and forgot about this. I'll update it and get those tests written ASAP, sometime this week.

lgebhardt:
@adamworrall I'd like to get this PR wrapped up. Have you had a chance to get back to this? If not would you mind if I finish this up?

lgebhardt:
#420 rebases this and adds support for the uuid routing constraint.

lgebhardt:
Closed through #420

lgebhardt:
Thanks @adamworrall!

PR #164, opened 2015-04-16
Text error codes

dennisfaust:
We'd prefer to use text for errors rather than codes.


lgebhardt:
A nice addition. Thanks!

PR #165, opened 2015-04-20
Corrections to support formatting option

sirvine:
The following configuration was broken until the changes in this PR were implemented:

```
JSONAPI.configure do |config|
   config.json_key_format = :camelized_key
   config.route_format = :underscored_route
end
```


lgebhardt:
Thanks!

PR #167, opened 2015-04-20
Extract jsonapi_relationships method for nested routes in block.

barelyknown:
This makes it easier to add the default nested routes if you also have block in which other nested routes are defined.


README.md
```diff
@@ -781,7 +781,15 @@ edit_contact GET    /contacts/:id/edit(.:format) contacts#edit
 ```
 
 To manually add in the nested routes you can use the `jsonapi_links`, `jsonapi_related_resources` and
-`jsonapi_related_resource` inside the block.
+`jsonapi_related_resource` inside the block. Or, you can add the default set of nested routes using the `jsonapi_relationships` method. For example:
+
+```ruby
+Rails.application.routes.draw do
+  json_resources :contacts do
```
dgeb: s/json_resources/jsonapi_resources/

dgeb:
:+1: Seems good once we fix that typo.

dgeb:
Thanks @barelyknown!

PR #168, opened 2015-04-20
Format Fixes

lgebhardt:
Add new tests for multiword keys and resource names to ensure the formatting is working correctly.

Fixes issues with formatting of type and ids.

PR #169, opened 2015-04-20
Remove format key.

barelyknown:
I believe that the `format_key` method was added erroneously in a recent commit. This branch accidentally includes commits from another PR that I sent in a bit ago, but I decided to just keep them in (to avoid the work of removing them).


lgebhardt:
In #168 I'm restoring this back to the way it was. We need to compare the formatted type. #168 adds some tests for handling formatted keys.

PR #172, opened 2015-04-24
Ensure validation errors use correct formatter

john-griffin:
Previously all validation errors returned to the client included keys
formatted in the ruby underscore style. To ensure clients receive
consistently formatted keys they should be formatted in the error 
message before responding.

If you can think of a better way to load the `key_formatter` into the error class I'm happy to make the changes.


lgebhardt:
Thanks!

PR #173, opened 2015-04-24
Unformat the linkage type.

barelyknown:
The linkage key was not being unformatted and was causing errors if there was a multiword key (because the raw key would not match the association name). This change unformats the key which fixes that bug.


barelyknown:
GRRR. More hashrockets needed. I'll fix now.

lgebhardt:
Thanks again!

PR #174, opened 2015-04-25
Fixes formatting for type mismatch check

sirvine:
Multi-word class names will fail the type name check unless `unformat_key` is applied.


sirvine:
Oops...looks like I missed 0.3.2 release. Will close and re-submit if necessary.

PR #175, opened 2015-04-26
Check for key on association.type should reference unformatted key

sirvine:
I wasn't able to get has_many associations with multiword keys to pass `has_key?` check unless they are unformatted.  Note that line 317 uses the unformatted key, too.


lgebhardt:
Thanks! I released 0.3.3 to include this.

PR #182, opened 2015-04-29
Update resource naming to fix plural cases

leandrocp:
This fix is useful for some kinds of words, specially not english words:

```
ActiveSupport::Inflector.inflections { |inflect| inflect.irregular 'foo_bar', 'foos_bar' }

"foo_bar".pluralize => "foos_bar"
"foos_bar".singularize => "foo_bar"

"#{"FoosBarController".sub(/Controller$/, '').singularize}Resource"
=> "FoosBarResource" # wrong

"#{"FoosBarController".underscore.sub(/_controller$/, '').singularize}_resource".camelize
=> "FooBarResource" # right
```

All tests are passing, but I need some help to write new tests for this code.

### DIFF
--- a/lib/jsonapi/resource.rb
+++ b/lib/jsonapi/resource.rb
@@ -457,7 +457,7 @@ module JSONAPI
       def _resource_name_from_type(type)
         class_name = @@resource_types[type]
         if class_name.nil?
-          class_name = type.to_s.singularize.camelize + 'Resource'
+          class_name = "#{type.to_s.singularize}_resource".camelize
           @@resource_types[type] = class_name
         end
         return class_name
--- a/lib/jsonapi/resource_controller.rb
+++ b/lib/jsonapi/resource_controller.rb
@@ -143,7 +143,7 @@ module JSONAPI
     end
 
     def resource_klass_name
-      @resource_klass_name ||= "#{self.class.name.sub(/Controller$/, '').singularize}Resource"
+      @resource_klass_name ||= "#{self.class.name.underscore.sub(/_controller$/, '').singularize}_resource".camelize
     end
 
     def ensure_correct_media_type


lgebhardt:
Interesting problem and a simple fix. It looks similar to the way it's being handled in [`Resource`](https://github.com/cerebris/jsonapi-resources/blob/master/lib/jsonapi/resource.rb#L460). Can you confirm the resources are being found correctly with the type of inflection rules you show (for example in has_many associations)?

Also, maybe we should change [line 460](https://github.com/cerebris/jsonapi-resources/blob/master/lib/jsonapi/resource.rb#L460) slightly so it matches your new style:

``` ruby
class_name = "#{type.to_s.singularize}_resource".camelize
```

As far as tests go I'm not coming up with good English language examples like this. If you have a set of inflection rules for some resources I can see about adding some tests.

leandrocp:
@lgebhardt  with your suggestion, that's the result:

```
4.2.1@2.2.2 (main)> type = "foos_bar"
=> "foos_bar"
4.2.1@2.2.2 (main)> type.singularize
=> "foo_bar"
4.2.1@2.2.2 (main)> type.singularize.camelize + "Resource"
=> "FooBarResource"
4.2.1@2.2.2 (main)> "#{type.singularize}_resource".camelize
=> "FooBarResource"
```

I can give you an example in pt-BR: while you write "phone_numbers" in english, I write "numeros_telefone" in pt-BR

I'm investigating if this affects `has_many` associations.

lgebhardt:
@leandrocp Thanks!

gottfrois:
BTW @lgebhardt what's the reason for "type" to be pluralized version of the model class name?

lgebhardt:
@gottfrois The decision to have type be plural was based on the type field in [JSON API ](http://jsonapi.org). I had to choose one or the other, and making it match the spec seemed right.

PR #186, opened 2015-05-01
Allow jsonapi_related_resources to accept options.

yawboakye:
`jsonapi_resources` accepts an options hash, which can be used to
determine which controller handles a certain request instead of the
default. It'd be great if `jsonapi_related_resources` accepted options
too, for similar reasons.


lgebhardt:
This seems reasonable. We should expand the option to the other methods as well, such as `jsonapi_related_resources`, `jsonapi_links`, and `jsonapi_link`.

yawboakye:
Let me update the pull request then.

lgebhardt:
@yawboakye Thanks!

yawboakye:
Glad I could contribute to a product I :heart: so much.

PR #187, opened 2015-05-01
Save after 2nd callback in Resource#change

### LINKED ISSUE #184: Association variables not available in resource link callbacks
I have been trying to build some authorization into a resource using before callbacks but have been struggling to do so with linked/associated resources due to lack of available association variables. When any of these callbacks are fired:

``` ruby
:create_has_many_link
:replace_has_many_links
:create_has_one_link
:replace_has_one_link
:remove_has_many_link
:remove_has_one_link
```

it does not seem any variables for the link/association can be accessed.

For example, say I was trying to associate a `Comment` with a `User` on a `CommentResource`. In the `before` `create_has_one_link` callbacks there are no available variables that include which `User` is being set. But I can access the `Comment` on the `@model`.

Is this the expected behaviour or am I missing something?

brettchalupa:
Instead of saving directly after the yield in the second `run_callbacks`
call, save after the second `run_callbacks` is finished when the resource
is not `@changing`.

This allows for using callbacks like `after_replace_fields` to work with
the model with the newly set attributes before it gets saved. This is
useful when doing authorization before creation, as well as using
non-datastore persisted attributes on the model to set specific data
on the model before it gets persisted.

Fixes https://github.com/cerebris/jsonapi-resources/issues/184

### DIFF
--- a/lib/jsonapi/resource.rb
+++ b/lib/jsonapi/resource.rb
@@ -46,8 +46,9 @@ module JSONAPI
           @changing = true
           run_callbacks callback do
             yield
-            save if @save_needed || is_new?
           end
+
+          save if @save_needed || is_new?
         end
       end
     end


lgebhardt:
@brettchalupa Thanks! This order makes a lot more sense.

brettchalupa:
Awesome, my pleasure! :smile: For what it is worth, I tested the branch on the project I am working on and the tests pass and things behave as expected.

PR #189, opened 2015-05-01
Allow the use of a different base controller for JSONAPI::ResourceController

tamiswa:
I'm already using [Rails API](https://github.com/rails-api/rails-api) and I would like to continue using its `ActionController::API` controller as the base for my controllers.

Here's my first stab at a pull request for allowing the base controller for `JSONAPI::ResourceController` to be configurable from the configuration block.  This almost works perfectly except I've learned that the `JSONAPI::ResourceController`  is loaded before the initializer can be run (meaning I can't actually change the super class) so I've used the following workaround.
1. Change the line in my `Gemfile` to add `:require => false`
2. Create an initializer that looks as follows:

``` javascript
# config/initializers/jsonapi_resources.rb

require 'jsonapi/configuration'

JSONAPI.configure do |config|
  config.base_controller = "ActionController::API"
end

require "jsonapi-resources"
```

As you can see this is less than ideal (unless of course there is no other way).  Any suggestions?

Thanks

### DIFF
--- a/lib/jsonapi/configuration.rb
+++ b/lib/jsonapi/configuration.rb
@@ -10,7 +10,8 @@ module JSONAPI
                 :default_paginator,
                 :default_page_size,
                 :maximum_page_size,
-                :use_text_errors
+                :use_text_errors,
+                :base_controller
 
     def initialize
       #:underscored_key, :camelized_key, :dasherized_key, or custom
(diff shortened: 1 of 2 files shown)


mcmire:
+1 for the need, it would be great to be able to use this with Rails API. I think that the way to solve this would be to make ResourceController a module instead of a superclass. That means you wouldn't have to have a config option or constantize anything.

lgebhardt:
I've taken @mcmire's idea to make a module for the guts of `ResourceController` so you can use them in a Rails API controller. This is #201.

I haven't tested this with Rails API yet. If you could give it a try I'd appreciate it. I'm also open to renaming the module if you have a better name.

tamiswa:
Awesome thanks!

PR #192, opened 2015-05-01
Removes Active Record dependencies from the core classes.

lgebhardt:
This attempts to remove Active Record dependencies from the core `Resource` and `Operation` and classes. The AR exceptions will now be handled by the `ActiveRecordOperationsProcessor`.

This required changes to the way resources handle save (using `save` instead of `save!`). In addition the resource is now checking if the underlying model is valid instead of relying on an exception being raised. This change will break with non active record models that do not implement `save` and `valid?`.


john-griffin:
Tested this branch against our API. Everything passes apart from resources where we used wrapper model objects that hooked into the `#save!` call. They are relying on the specific AR method calls that the resource makes so no big deal, we will just update them to handle `#valid?` and `#save` instead.

lgebhardt:
@john-griffin Thanks for testing. Glad it won't be too much work for you to update your code. I updated to note for this PR to indicate the breaking changes that you have found.

PR #193, opened 2015-05-01
Protect against page[number]=0 param

jamonholmgren:
My Ember app tries to send a `page[number]=0` parameter, and jsonapi-resources blows up with a negative offset MySQL error. This protects against that.

### DIFF
--- a/lib/jsonapi/paginator.rb
+++ b/lib/jsonapi/paginator.rb
@@ -34,6 +34,7 @@ class OffsetPaginator < JSONAPI::Paginator
       validparams = params.permit(:offset, :limit)
 
       @offset = validparams[:offset] ? validparams[:offset].to_i : 0
+      @offset = 0 if @offset < 0
       @limit = validparams[:limit] ? validparams[:limit].to_i : JSONAPI.configuration.default_page_size
 
       if @limit < 1
--- a/lib/jsonapi/paginator.rb
+++ b/lib/jsonapi/paginator.rb
@@ -19,6 +19,7 @@ end
 class OffsetPaginator < JSONAPI::Paginator
   def initialize(params)
     parse_pagination_params(params)
+    verify_pagination_params
   end
 
   def apply(relation)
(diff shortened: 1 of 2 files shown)


lgebhardt:
Are you using the Paged or the Offset paginator? I ask because you modified the Offset, but your example indicates the Paged (I can see the Paged would cause this issue on line 59).

Are there any databases that support negative offsets? I looked at a few and didn't see any of the common ones allowing it. So the code is probably safe, but I don't think it's totally clear since both page 0 and 1 will return the same results.

I'm thinking the right way to handle this is an option for what the first page should be (0 or 1). It's obvious your Ember app is using 0 based pages, whereas the PagedPaginator is assuming 1 based pages (and isn't doing any range checking). So we could add a configuration option for first page and do the calculations from there. We should return an error if an invalid page is specified, rather than just shifting the offset.

Thanks for reporting this.

jamonholmgren:
> Are you using the Paged or the Offset paginator?

Ah yes, I modified the wrong one. I'm using `:paged`.

``` ruby
JSONAPI.configure do |config|
  # built in paginators are :none, :offset, :cursor, :paged
  config.default_paginator = :paged

  config.default_page_size = 25
  config.maximum_page_size = 250
end
```

> Are there any databases that support negative offsets?

I don't know of any, and it doesn't seem like there would be a use case for it.

>  I don't think it's totally clear since both page 0 and 1 will return the same results.

True. I'm mainly trying to protect against dumb clients (like mine, apparently), which send 0 when the page number isn't set.

> It's obvious your Ember app is using 0 based pages

It's actually using 1-based pages. But when the page[:number] param isn't set on the ember side, it sends zero. I will fix that, but thought that the behavior on the server side wasn't very intuitive (it blows up with a MySQL error).

> We should return an error if an invalid page is specified, rather than just shifting the offset.

On second thought, I think this is the right thing to do. I'll modify my pull request. 

Thanks for your thoughtful response!

jamonholmgren:
I've modified the code to raise exceptions, but one test is failing which makes it seem like the semantics should be to silently adjust the page number.

``` ruby
  def test_books_paged_pagination_invalid_page_format_interpret_int_text
    Api::V2::BookResource.paginator :paged

    get :index, {page: 'qwerty'}
    assert_response :success
    assert_equal 10, json_response['data'].size
    assert_equal 'Book 0', json_response['data'][0]['title']
  end
```
1. Change the test to raise an exception? This changes the behavior, of course.
2. Assume <= 0 should silently adjust to 1?
3. Make <= 0 raise an exception, but invalid alphanumeric param silently adjust to 1?

lgebhardt:
Looking at it, I think I have the test wrong. It would be better to have the request fail if an invalid page is specified, whether it's text or <1.

jamonholmgren:
@lgebhardt Updated the test, it's passing on my machine.

lgebhardt:
Thanks again!

jamonholmgren:
:+1: Thanks!

PR #194, opened 2015-05-02
Update README serializer example usage

mmartinson:
Existing docs incorrect in passing parameters to #serialize_to_hash instead of new


lgebhardt:
Thanks!

Issue #126, opened 2015-03-18
Pagination: total record/page count in meta


realityendshere:
My Ember apps use sparse arrays to load visible data on demand. The sparse array needs some means for determining the total number of objects available to show. This allows my colleagues and I to fake the array length property with record counts provided by the server and, for example, draw an appropriately tall scrolling `<div>` to mimic the appearance of a single, continuous listing.

JSON API spec suggests using the top-level `links` object to include URLs for the first, last, next, and previous available pages. But offers no suggestions for where to provide a document count. My instinct suggests sending this data in the `meta` object. After all, the record count is not necessarily useful for every app or api.

I am guessing JR first needs to support the meta object (see #89) before addressing this need. I am opening this issue for future discussion since it is something critical to the apps I work on.

xdmx:
:+1: on this. It would also be useful to have it optional, with the ability to enable it globally or on only specific resources.

This would be useful in some scenarios, like showing the total posts a user has created, and at the same time being able to disable it (avoiding the .count query) for showing for example the number of notification a user has received

realityendshere:
@xdmx Agreed! In the extreme case, a Twitter-like app, the total count of tweets in your feed could easily be massively large and essentially useless.

In contrast, a listing of followers may be a large but manageable number that requires a pagination strategy that includes a count of matching records.

nonsensery:
:+1: from me too. I don't see a place in the JSON API spec for a total count (right?), but at least supporting the pagination entries in the top-level `"links"` would be nice.

michelegera:
:+1: on this as well.

The JSON API spec details this use case in the [extension specification](http://jsonapi.org/extensions/) – take a look at the example in the “Profiles” section.

It belongs indeed to the `meta` member as @realityendshere was guessing.

bwinterling:
:+1: for meta/ pagination!

jamonholmgren:
:+1: Another vote for this! Also using Ember.

senid231:
+1
it will be better to have choice: add total_count to top level links or meta or even both

lgebhardt:
The 0.4.0 release now has support for collecting meta data during the processing of the operations, collecting the meta from all operations, and outputting the collected meta as part of the new ResponseDocument.

[...]

PR #199, opened 2015-05-07
Adding total_count to meta data (optionally) to index response

bvandenbos:
Partial implementation for #126.  This allows a resource to opt-in to returning the total count in the meta data like so:

``` ruby
      attribute :title
      attribute :isbn

      include_total_count # tells the serializer to include total_count in meta

      has_many :book_comments
    end
```

I don't love this implementation, but wanted to get some feedback.  It wasn't clear where to start constructing the meta hash or what object should be responsible for issuing the query to get the total count.  Feels like there is a larger refactor that's trying to be born to accommodate meta (and probably links) in a broader sense.


barelyknown:
Couple of comments:

1) I'd prefer the ability to set `include_total_count` to `true` or `false`, with `true` as the default argument.

``` ruby
    include_total_count false
  end
```

2) Can you add the ability to set a default configuration too? Something like:

``` ruby
JSONAPI.configure do |config|
  config.default_include_total_count = true
end
```

bvandenbos:
Great feedback.  Both items added.

slaskis:
I was looking for this myself, mainly to use for pagination. 

Maybe this API could be extended to a generic `meta` property? Something like this:

```
  attribute :title
  attribute :isbn

  meta :total_count # tells the serializer to include total_count in meta

  has_many :book_comments
end
```

Which may allow adding other future properties as meta. However this may be confusing as to if it's a top-level meta or resource level meta when I think about it. Maybe not for this feature but others. These could be separated by defining class-methods (top-level) vs instance-methods (resource-level).

And it could also allow for setting options on the meta (simply passed in to the method). Something like this:

```
  attribute :title
  attribute :isbn

  # as second argument (seems pretty rails to me)
  meta :total_count, filter: [:special_for_count]

  # or as a hash, which would allow for setting multiple meta at once
  meta total_count: {filter: [:special]}, other_meta: [:something]

  has_many :book_comments
end
```

What do you think? It seems to me these additions could easily be "plugins" by just being class or instance methods to be included into the resource in a module.

bvandenbos:
In general, I like this approach, but we probably want to account for both top-level meta and resource-level meta.  `total_count` in this case would be a top-level meta entry.  The spec also allows "meta" at the entry level.  Something like this perhaps:

``` ruby
  attribute :title
  attribute :isbn

  # Contrived resource-level meta example:
  meta :md5, -> { |resource| resource.compute_md5 }

  class << self
    # This would define top-level meta entries
    meta :total_count do |context|
      # apply filters and get the count
    end
  end

  has_many :book_comments
end
```

Additionally, the serializer will need access to the context in order to perform certain types of queries (ie: getting the total count).  Currently, the serializer is not privy to the filters that generated the query and only has access to the array of records.

lgebhardt:
Thanks for starting this. This feature has been added, though slightly differently, in other PRs.

PR #200, opened 2015-05-07
Use new Attributes hash per JSON-API changes

lgebhardt:


test/controllers/controller_test.rb
```diff
@@ -1096,6 +1143,31 @@ def test_update_missing_type
...
+    set_content_type_header!
+    javascript = Section.find_by(name: 'javascript')
+
+    put :update,
+        {
+          id: 3,
+          data: {
+            id: '3',
+            type: 'posts',
+            body: 'asdfg',
```
dgeb, line 1156: isn't `body` an attribute?

lgebhardt, line 1156: Yes, that's what this test is supposed to catch.

test/controllers/controller_test.rb
```diff
@@ -1096,6 +1143,31 @@ def test_update_missing_type
     assert_match /The required parameter, type, is missing./, response.body
   end
 
+  def test_update_unknown_key
+    set_content_type_header!
+    javascript = Section.find_by(name: 'javascript')
+
+    put :update,
+        {
+          id: 3,
```
dgeb, line 1152: extra `id`?

test/controllers/controller_test.rb
dgeb, line 1156: Ah, I see!

test/controllers/controller_test.rb
lgebhardt, line 1152: The top level `id` is provided by the route and wouldn't be part of the payload.

dgeb:
looks good @lgebhardt :+1:

PR #201, opened 2015-05-08
ActsAsResourceController Module

lgebhardt:
Move core of ResourceController to a module, ActsAsResourceController.

Alternate approach to #189


mcmire:
Thanks for the quick reply, this is great. :+1:

PR #202, opened 2015-05-10
Added a meta property

slaskis:
After writing [this comment](https://github.com/cerebris/jsonapi-resources/pull/199#issuecomment-100619435) and based upon that PR I gave it a shot and in short it allows this syntax:

```
  attribute :title
  attribute :isbn

  # pass optional options as a second argument
  meta :total_count, filter: [:special_for_count]

  has_many :book_comments
end
```

Where, in theory, a `total_count` class-method has been autoloaded into `JSONAPI::Resource` in a gem or something. In the tests I simply extended a few resources with a module.

Let me know what can be done better and why this is a bad idea please :)

### DIFF
--- a/lib/jsonapi/resource.rb
+++ b/lib/jsonapi/resource.rb
@@ -195,6 +195,7 @@ module JSONAPI
 
     class << self
       def inherited(base)
+        base._meta = (_meta || {}).dup
         base._attributes = (_attributes || {}).dup
         base._associations = (_associations || {}).dup
         base._allowed_filters = (_allowed_filters || Set.new).dup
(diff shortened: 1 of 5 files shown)


lgebhardt:
#258 introduces a configuration option to output the record count after a find operation. That uses the filters from the find operation, which is slightly different than what is presented here. If there's a need to have custom filters for just the count we can look into that in addition to #258.

henrymazza:
It doesn’t address how to add a meta to a resource. The specification says it must be at the same level as “attributes”. So no cheating usint `attribute :meta`, it has to be one level up. So, do I have to hack the serializer?

lgebhardt:
@henrymazza This is really a separate issue from this PR which is related to top-level meta. The system now supports meta at the top level, such as the total_count, but does not yet support all locations where meta is allowed by the spec (such as resource object meta). I've add #371 to address this. Please add your requirements there. I'm thinking it's a fairly simple feature, but I don't have a good feel for the type and source of the resource object meta that needs to be supported.

PR #203, opened 2015-05-11
Removed duplicated line from README.md

louisshepherd:
Removed duplicated line about PrimaryKeys from README.md


lgebhardt:
Thanks!

PR #204, opened 2015-05-11
Auto eager load included associations

slaskis:
I've been battling a bit in my app with really slow response times caused by quite deep relationships so eager loading of them was required. 

But I found that something in jsonapi-resources can cause the eager loaded data to be "lost" so it eager loaded AND loaded in N+1 queries, which was less than ideal but I finally fixed this with 012a6c2. 

Then I thought "hey, we know what is desired already, why not include them automatically!". So that's what d15853e does.

Just using this a request goes down from ~450 queries to ~18 (most of them because I'm still abusing attributes, which is out of the scope of this PR).

And it passes the tests (at least locally).

### DIFF
--- a/lib/jsonapi/request.rb
+++ b/lib/jsonapi/request.rb
@@ -4,7 +4,8 @@ require 'jsonapi/paginator'
 module JSONAPI
   class Request
     attr_accessor :fields, :include, :filters, :sort_criteria, :errors, :operations,
-                  :resource_klass, :context, :paginator, :source_klass, :source_id
+                  :resource_klass, :context, :paginator, :source_klass, :source_id,
+                  :include_directives
 
     def initialize(params = nil, options = {})
       @context = options.fetch(:context, nil)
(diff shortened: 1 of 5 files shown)


lgebhardt:
@slaskis Thanks for your work on this. In light of your new code I felt I should rename the SerializerIncludeDirectives, so I put together #205.

It seems to be working very well for the case of a single level include. However with nested includes it's still losing the eager loaded data, as #206 shows. Even with this issue it's still a vast improvement.

slaskis:
Yeah I noticed that today as well. I thought it worked pretty well yesterday but adding something like pagination or filters and it still breaks.

I don't have that much experience with the innards of the ActiveRecord eager loading and how to troubleshoot it. It's pretty much been brute force testing step by step using byebug to see when the relation is changed for now.

slaskis:
It's great that you've set up a test case though!

slaskis:
Closing this to see the rest of the magic happen in #205

PR #205, opened 2015-05-12
Auto eager load included associations + rename of SerializerIncludeDirectives

lgebhardt:
This combines #204 with a refactor of SerializerIncludeDirectives to just IncludeDirectives to clarify that it's used by more than just the Serializer now.


slaskis:
When I worked on this using monkey patches it worked even with nested includes (and we're using some nasty deep includes). But then to get the tests to pass I had to change [this line](https://github.com/cerebris/jsonapi-resources/commit/012a6c2034caf4763965e1b113da41dcc4188e75#diff-e6cf31c573bd5f2277b1f16812943df4R345) and add the `.all` part. I have no idea why though. 

But when I change that line back to just `records` the query_count tests does not fail but a lot of others do. 

Do you understand why that is?

slaskis:
Actually adding some logs it seems to be when `records` is actually just a `Model` and not a `relation`

slaskis:
Hah! 

Changing that `if`-block to 

```
        if required_includes.any?
          records.includes(required_includes)
        elsif records.respond_to? :to_ary
          records
        else
          records.all
        end
```

and now all the tests pass :)

I'm sure there's some work left to do with filters and paginations but at least it's progress

lgebhardt:
I think you may have found it. Great job!

slaskis:
Yay!

lgebhardt:
@slaskis Thanks for all your work on this.

slaskis:
No problem, just really happy to see that query count go down (and the request times with it).

PR #211, opened 2015-05-15
Primary key sorting

### LINKED ISSUE #210: Sorting by id does not use primary_key
I have a resource which uses `key` as the "primary_key". 

But when I try to sort using `?sort=+id` it uses the internal `id`-column instead of the `key`-column as expected.

I've done this little monkey patch locally for [this code](https://github.com/cerebris/jsonapi-resources/blob/master/lib/jsonapi/resource.rb#L512-L516) for now:

```
      def construct_order_options(sort_params)
        sort_params.each_with_object({}) { |sort, order_hash|
          field = sort[:field] == 'id' ? _primary_key : sort[:field]
          order_hash[field] = sort[:direction]
        }
      end
```

Not sure if that's where this should be fixed or if it should be dealt with in the request.

But if this is ok, just say the word and I'll make a PR. :)

slaskis:
This PR fixes #210 by using primary_key when trying to sort on "id". With some tests to catch regressions.

### DIFF
--- a/lib/jsonapi/request.rb
+++ b/lib/jsonapi/request.rb
@@ -15,7 +15,7 @@ module JSONAPI
       @fields = {}
       @include = []
       @filters = {}
-      @sort_criteria = []
+      @sort_criteria = [{field: 'id', direction: :asc}]
       @source_klass = nil
       @source_id = nil
       @include_directives = nil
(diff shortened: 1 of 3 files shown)


lgebhardt:
Thanks!

john-griffin:
I have found the way `@sort_criteria` has been defaulted in the request initializer makes it difficult to change the default sort order now. Do you think its best to remove that part for now?

Issue #45, opened 2014-12-01
Integrate with kaminari or will_paginate


xdmx:
I find this project the best one to implement the json api under a rails project.

The only thing I really miss is some kind of pagination integration (I know that it's still not in the json api spec, sadly), anyway, it'd be very useful some kind of integration with kaminari or will_paginate, with the ability to pass the page and page size to return.

It'd be awesome to also support different page size for included resources and to have some kind of config for a max_size (to block requests for 100000 items)

I guess this will be a placeholder until json-api/json-api#236 gets resolved

lgebhardt:
Yes pagination is a big missing feature. I plan on implementing pagination, however I've been holding off until the details are resolved in json-api.

I'm not sure about paginating related resources. This seems like a thorny problem that potentially has serious performance issues for the way JR is implemented (and I think has to be implemented to support nested includes). I've been thinking we would have a configurable cap on the number of related records. Once that is exceeded I'm not sure how the code should behave.

whatthewhat:
@lgebhardt @xdmx do you think [providing access to the Active Record relation](https://github.com/cerebris/jsonapi-resources/pull/46) could be a good option in the meanwhile?

xdmx:
I'm not sure about the pagination, since it'd also require to return some data in the response/body.

For sure it's very useful to set a context on resources like: find all orders for the current user (it'd be very bad if it returns all orders from all users!), without having to rewrite the find method (I'm still new to JS, I don't know if there is already a better way, that seems the one suggested in the readme)

whatthewhat:
I thought I saw a way to specify the response `meta` key (where the pagination data could be), but I can't find it now..

lgebhardt:
Is this it? http://jsonapi.org/extending/

whatthewhat:
Oh, I meant specifying `meta` with jsonapi-resources gem, but I don't think it can be done right now.

lgebhardt:
No, jsonapi-resources doesn't yet use the meta tag.

lgebhardt:
> do you think providing access to the Active Record relation could be a good option in the meanwhile?

@whatthewhat I like the solution you're proposing. Not only will it make implementing pagination cleaner, but it also simplifies overriding the find logic (for most apps).

whatthewhat:
@lgebhardt Great! I'll update the PR when I'll have some spare time.

It does not however completely resolve this issue, the pagination data would need to be added to the response somehow..

[...]

PR #214, opened 2015-05-18
Paginator Order Context

john-griffin:
To build paginators that rely on sort order (e.g. cursor pagination) the
sort order requested must be known when applying pagination. This adds
an additional argument to Paginator#apply to supply the current sort 
order and hooks it in through the resource finders.

See #45


lgebhardt:
Closed with #225

PR #215, opened 2015-05-18
Adds default value to filter option

old-grrt:
Adds ability to define a default value for a filter

```
filter :status, default: 'active'
```

This value is passed to filter parse, so:

```
filter :status, default: 'active,normal'
```

should work as it would come from a request parameter.


lgebhardt:
Merged with #226. Thanks!

PR #217, opened 2015-05-21
Simplifying the code that pulls the keys from associations

adomokos:
This makes the code a bit simpler, and I think more expressive as well.
All the tests are passing.


lgebhardt:
Thanks!

PR #220, opened 2015-05-27
Changes to support meta data

lgebhardt:
This set of changes lays the groundwork for meta data.
- All controller actions are now backed by an operation.
- OperationsProcessors now support callback for each operation.
  - This allows further processing to happen, such as collecting the record count.
- Adds a ResponseDocument that can serialize meta data
- OperationsProcessors can be set by a configuration option, in addition to overriding at the controller level (should probably be deprecated in the docs).

There's still four lines of error handling code that isn't covered by tests. And the Readme needs to be updated for the new features. But I want to get this in sooner rather than later.

### DIFF
--- a/lib/jsonapi/request.rb
+++ b/lib/jsonapi/request.rb
@@ -31,7 +31,10 @@ module JSONAPI
       @resource_klass ||= Resource.resource_for(params[:controller]) if params[:controller]
 
       unless params.nil?
-        send("setup_#{params[:action]}_action", params)
+        setup_action_method_name = "setup_#{params[:action]}_action"
+        if respond_to?(setup_action_method_name)
+          send(setup_action_method_name, params)
+        end
       end
     rescue ActionController::ParameterMissing => e
       @errors.concat(JSONAPI::Exceptions::ParameterMissing.new(e.param).errors)


lib/jsonapi/operation.rb
```diff
@@ -1,40 +1,166 @@
...
+
+    def apply(context)
+    end
+  end
+
+  class FindOperation < Operation
+    attr_reader :filters, :include_directives, :sort_criteria, :paginator
+
+    def initialize(resource_klass, options = {})
+      @filters = options.fetch(:filters, nil)
```
adomokos: @lgebhardt: Could you just use `options[:filters]` here? That would give you nil if the key is not found.

lgebhardt: @adomokos, now why didn't I think of that? :neutral_face:

Thanks for the suggestion.

adomokos: :+1:

slaskis:
Oh my this is a big change but I really like how it's now a very easy to follow pipeline. 

I haven't tried it out yet, but just looking at the changes I see greatness. 

Just to clarify though, the end result of this, besides having an nice and clear way override and tweak the api, is that it takes care of the top-level meta, it does not do anything for resource- or link-level metas yet, right?

lgebhardt:
@slaskis This is just a start. I hope to get support for resource level meta and link level meta in soon. But first I plan to work on bringing the project into compliance with the latest RC4 version of JSONAPI.

john-griffin:
Ran this branch in our app and tests are passing. Thanks!

barelyknown:
I tested this against our app and just saw one issue (because of an extra action that we added to one controller). I added a guard to prevent the problem in this PR: https://github.com/cerebris/jsonapi-resources/pull/222.

PR #222, opened 2015-05-28
Check if setup_*_action method exists before calling.

barelyknown:
I added a non-RESTful route that didn't have a corresponding `setup_*_action`, so I added this check to prevent calling the method if it doesn't exist.


lgebhardt:
Thanks!

PR #224, opened 2015-05-29
Use resource_class of relationship to apply filters, etc.

barelyknown:
This one took me a while to track down! :tada:

The source resource class was being used to apply filters, sorting, and pagination for relationships. So, if `apply_filters` et al were implemented in the resource class for the relationship, they weren't being used. That was causing incorrect results to be returned by the `get_related_resources` controller method.


lgebhardt:
Good catch! Thanks.

PR #227, opened 2015-05-29
Remove unneeded relationship_data method

lgebhardt:
Fixes code clarity issue brought up by @ggordon.
The code was a remnant from an earlier approach and needed to be removed.


ggordon:
@lgebhardt Awesome, thanks.

PR #230, opened 2015-06-01
Only loop through filters if #each is implemented

brettchalupa:
In Request#parse_filters, only loop through the filters if the passed in
value implements the #each method. Without this check,

```
NoMethodError: undefined method `each' for "all":String
```

gets thrown when request filters are not in a not valid format.

I am happy to add tests for this - I did not see any for `Request`,
but I most certainly can. :smile:


brettchalupa:
Writing specs at the moment, it seemed worthwhile since there is branching logic. Commit incoming. :sweat_smile:

lgebhardt:
Thanks. I think we should probably raise an error if the format is wrong.

brettchalupa:
@lgebhardt is `FilterNotAllowed` an appropriate error to raise? It feels like this may be a new error specific to the filter syntax.

lgebhardt:
I agree, it seems like a new error. How about `InvalidFilterFormat` or `InvalidFilterSyntax`?

brettchalupa:
Alright, amended, adjusted the commit to have tests, and added a new error `InvalidFiltersSyntax`. I made it plural because the other filter errors are singular, where this is more about the way the filters are applied. Should it be singular?

lgebhardt:
Thanks! I think the plural version if fine.

Issue #77, opened 2015-02-11
How should polymorphic relationships be handled?


barelyknown:
I have a class with a polymorphic relationship.

``` ruby
class Identification < ActiveRecord::Base
  belongs_to :identified, polymorphic: true
end
```

I would like to do something like this:

``` ruby
class IdentificationResource < BaseResource
  has_one :identified, polymorphic: true
end
```

This doesn't appear to be supported by `jsonapi-resources` yet (I checked `master` and `jsonapi_rc2`), but since polymorphism is supported in RC2, it seems like we should add it.

What are you thinking for this feature? I'd be happy to help.

lgebhardt:
@barelyknown I was planning on adding polymorphism support to `jsonapi-resources` in the future. If you want to tackle adding support to the `jsonapi_rc2` branch that would be nice. I haven't given the implementation details a whole lot of thought, but if you need someone to bounce some ideas off of let me know.

vrybas:
@barelyknown does #274 covers this issue?

samselikoff:
@lgebhardt I believe this can be closed

jufemaiz:
@samselikoff what's the approach to use?

vrybas:
@jufemaiz see #274

PR #231, opened 2015-06-01
Add support for has_one :x, polymorphic: true

ggordon:
**Not ready for merge!** Want to start discussion.

This patch adds some support for polymorphic associations. The main issue remaining is how to handle the related relationship links. I'd be interested in hearing ideas on how to do this, even what to do would be helpful. This should address issue #77 when completed.


lib/jsonapi/resource_serializer.rb
```diff
@@ -168,6 +168,7 @@ def relationship_data(source, include_directives)
               resource = source.send(name)
               if resource
                 id = resource.id
+                type = association.polymorphic? ? resource.class._type : type
```
nonsensery: What do you think about wrapping this logic up in a method (on Association)? It seems like lines 229-234 could benefit from that.

nonsensery:
Test coverage for polymorphic `has_many` relationships looks great, but I don't see any coverage for polymorphic `has_one` relationships. Am I missing it?

ggordon:
@nonsensery Thanks, I was focused on the other end of the polymorphic association, but I'll add tests for the has_one. I'll also try the refactor you suggested.

PR #235, opened 2015-06-02
Allow alternate response code

### LINKED ISSUE #234: 202 Accepted response on POST
I'd need to return `202 Accepted` response on some resource creation and I didn't find anything about it on the codebase. 
Is it already supported? Thanks

From the [jsonapi docs](http://jsonapi.org/format/#crud-creating-responses-202):

> **202 Accepted**
> If a request to create a resource has been accepted for processing, but the processing has not been completed by the time the server responds, the server **MUST** return a `202 Accepted` status code.

lgebhardt:
Allows resources to return a 202 instead of 201 or 200. Useful for when a save is delayed.

Fixes #234


lib/jsonapi/resource.rb
```diff
@@ -42,11 +42,16 @@ def change(callback)
...
           @changing = true
+          result = nil
           run_callbacks callback do
-            yield
+            result = yield
           end
 
-          save if @save_needed || is_new?
+          if @save_needed || is_new?
+            save
```
dgeb: Is it ok that `result` is ignored in this case?

dgeb:
Looks good but I'd prefer to return `nil` implicitly instead of `:default` explicitly. We don't want `:default` to be confused with a valid status code.

lgebhardt:
Reworked this to allow the resource methods to be either `:completed` or `:deferred`, and then the Operation determines the response code. It's more limiting, but the resource shouldn't know about response codes.

The model now no longer needs to implement save if it's not an Active Record model. The resource now checks if the save method is defined before attempting to call it.

During the rework it became obvious that if there are callbacks on the model that cancel the save we weren't accounting for that. It now raises an exception. But I'm not sure I like the flow if a callback cancels the save. It seems that shouldn't make it out as an error in the response, but I'm not sure how we should handle it. There isn't a good way that I see to pass any additional data back from the callback in the model. So right now it returns back an error with a non configurable message.

Any better ideas?

bartoszkopinski:
For future reference as this doesn't seem to be documented:


```
    # Override this on a resource to return a different result code. Any
    # value other than :completed will result in operations returning
    # `:accepted`
    #
    # For example to return `:accepted` if your model does not immediately
    # save resources to the database you could override `_save` as follows:
    #
    # ```
    # def _save
    #   super
    #   return :accepted
    # end
    # ```
```

https://github.com/cerebris/jsonapi-resources/blob/7ebba9b096c4d84d351b9c18af2971a36b97129d/lib/jsonapi/resource.rb#L182...L194

PR #237, opened 2015-06-03
Rework includes logic to JSON API 1.0 

lgebhardt:
Full include path is now serialized.
Fields control linkage.


john-griffin:
Tested this out because we were having some issues with nested `include`. Works great! The spec and the implementation requiring that _intermediate resources in a multi-part path must be returned along with the leaf nodes_ really simplifies things for us.

Thanks :+1:

barelyknown:
I didn't look closely yet, but this might fix https://github.com/cerebris/jsonapi-resources/pull/116.

dgeb:
LGTM :+1:

PR #238, opened 2015-06-05
Provide options context to apply_filter

john-griffin:
Overriding `apply_filter` is the recommended way to implement custom 
filters. However if filtering relies on the context (e.g. `current_user`)
it was not possible to implement the custom filter. This passes the 
options context to the `apply_filter` method so it is available when 
overriding in a resource.


john-griffin:
Tests passing now.

barelyknown:
Good catch.

lgebhardt:
Good catch. Thanks!

PR #239, opened 2015-06-05
add options :key and :readonly for JSONAPI::Resource.attribute

senid231:
key option allows you to prettify attribute names ( column name in model can be different from attribute name in resource).
readonly option allows you to define readonly attributes correctly (if you call something like title= it will raise NoMethodError in resource but not in model).

i think its pretty useful :)

### DIFF
--- a/lib/jsonapi/resource.rb
+++ b/lib/jsonapi/resource.rb
@@ -248,15 +248,16 @@ module JSONAPI
           ActiveSupport::Deprecation.warn('Id without format is no longer supported. Please remove ids from attributes, or specify a format.')
         end
 
+        model_attr = options.fetch(:key, attr)
         @_attributes ||= {}
         @_attributes[attr] = options
         define_method attr do
-          @model.send(attr)
+          @model.send(model_attr)
         end unless method_defined?(attr)
 
         define_method "#{attr}=" do |value|
-          @model.send "#{attr}=", value
-        end unless method_defined?("#{attr}=")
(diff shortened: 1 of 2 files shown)


barelyknown:
"Read only" is possible now by overriding the `createable_fields` and `updatable_fields` class methods. Does that give you what you need?

Regarding alternate attribute names, I'd prefer keeping that responsibility in the model class and out of this library.

lgebhardt:
I agree with @barelyknown on both of these issues. The use of `creatable_fields` and `updateable_fields` is to allow fine grained control vs. just a single read_only flag. Some fields can only be set when the resource is first created.

denkristoffer:
Looks like `creatable_fields` and `updateable_fields` are being phased out, so how are you supposed to set read only attributes now?

lgebhardt:
Just the deprecated misspelled versions are being removed. Rename `createable` to `creatable` and `updateable_fields` to `updatable_fields`.

machty:
Is the best practice still to add keys to both `creatable_fields` and `updatable_fields`? Is there any unified shortcut for excluding a field from both?

lgebhardt:
@machty It's still best practice.  There isn't a unified shortcut, but there has been some discussion around adding a `read_only` option for attributes.

hidde-jan:
I'm 👍  on a shortcut. I often employ computed attributes that are not backed by a database column and adding those to both `creatable_` and `updatable_attributes` all the time is needlessly verbose.

PR #240, opened 2015-06-05
Provide filter context to resource records query

john-griffin:
Sometimes when overriding the `records` method on a resource it is useful
to have access to the current filters. For instance if the underlying 
base query for the resource changes based on a filter param. This can 
not be done in the `apply_filter` method because compound filters would 
produce inconsistent results depending on the order in which they are 
run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -307,7 +307,7 @@ For example to allow a user to only retrieve his own posts you can do the follow
 class PostResource < JSONAPI::Resource
   attribute :title, :body
 
-  def self.records(options = {})
+  def self.records(options = {}, filters = {})
     context = options[:context]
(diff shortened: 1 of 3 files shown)


lgebhardt:
I see what you are going for, but It seems that it will make the records method a bit confusing with the filters in there. Will people try to apply the filters to the record, and then they will be applied a bit later in `filter_records`? Also, `find_by_key` won't be passing in the filters so that seems a bit inconsistent.

Would just overriding the `filter_records` method allow you to do the same thing?

john-griffin:
@lgebhardt yeah good point, agree might be confusion for new users. I'll close this any try overriding `filter_records`.

PR #243, opened 2015-06-08
Eliminate association type checks

ggordon:
This refactor removes the checks on association type like `association.is_a?(JSONAPI::Association::HasMany)`. Reduced scope to only ResourceSerializer. Pulled out AssociationSerializer.


lgebhardt:
I see what you are going for in simplifying the code that switches based on the type of the association. It is a bit messy the way it is. However I'm not so fond of mixing in the other concerns into the Association classes.

What about creating an `AssociationSerializer` class that encapsulates the serialization logic for each type of association? You could then load up the correct subclass based on the association type.

ggordon:
@lgebhardt Thanks for the feedback, I'll look into doing that.

PR #246, opened 2015-06-09
Fixed bug that ignores source_type when association is loaded eagerly

jakubrohleder:
With model schema:

```
class Activity < ActiveRecord::Base
  has_many :activity_roles
  has_many :event_invitations, through: :activity_roles
  has_many :users, through: :event_invitations, source_type: 'User'
  has_many :hosts, through: :event_invitations, source_type: 'Host'
  # ...
end

class ActivityRole < ActiveRecord::Base
  belongs_to :activity
  belongs_to :event_invitation
  # ...
end

class ActivityRole < ActiveRecord::Base
  belongs_to :activity
  belongs_to :event_invitation
  # ...
end

class EventInvitation < ActiveRecord::Base
  belongs_to :event
  belongs_to :user, polymorphic: true
  has_many :activity_roles
  # ...
end
```

Getting `Activity` with `includes(:users)` returns CollectionProxy with both `Users` and `Hosts`. For activity with 10 users and 2 hosts:

```
Activity.includes(:users).find(Activity.first.id).users.size
=> 12 
Activity.includes(:users).find(Activity.first.id).users.count
=> 10
Activity.includes(:users).find(Activity.first.id).users.all,size
=> 10
```

This bug with this model design makes resource to be applied to different classes and return wrong data (or raise exception if class have different attributes). After patch everything seems to work fine.

### DIFF
--- a/lib/jsonapi/resource.rb
+++ b/lib/jsonapi/resource.rb
@@ -602,7 +602,7 @@ module JSONAPI
               resources = []
 
               if resource_class
-                records = public_send(associated_records_method_name)
+                records = public_send(associated_records_method_name).all
                 records = resource_class.apply_filters(records, filters, options)
                 order_options = self.class.construct_order_options(sort_criteria)
                 records = resource_class.apply_sort(records, order_options)


lgebhardt:
This change introduces an N+1 queries issue. There are a few tests that count the number of queries being run, and they seem to be broken by this. It looks like the addition of `.all` is short circuiting the rails query optimization process.

jakubrohleder:
You are right. I'll try to fix this issue other way, do you have any ideas?

lgebhardt:
Can you turn your example code above into a test that shows the problem?

jakubrohleder:
Sure, but how to run tests?

```
$ bundle install
$ rake 
```

results in error:

```
LoadError: cannot load such file -- rack/test
/Users/ves/dev/jsonapi-resources/test/test_helper.rb:14:in `<top (required)>'
```

lgebhardt:
That should work. I normally run the tests along with a coverage report, which ideally should be 100%:

``` bash
COVERAGE=true rake test
```

But just running `rake` works too. What version of ruby and rails are you using?

jakubrohleder:
Rbenv with ruby 2.1.4 and rails 4.2.1.

I have noticed some time ago that my installation can be broken. I think I'll reinstall it first and then if any further problems occure I'll call for your help.

lgebhardt:
That should work. A reinstall is probably a good idea.

vrybas:
@jakubrohleder any progress with this?

scottgonzalez:
@lgebhardt This should probably be closed with all of the changes that have occurred since the PR was opened.

PR #247, opened 2015-06-09
Filter includes that are not valid associations

jakubrohleder:
When resource have has_many ot has_one that are not real associations they are still getting included in eager loading of a record. This PR fixes this issue by filtering those *directives.model_includes that are not real associations of a resource.

### DIFF
--- a/lib/jsonapi/resource.rb
+++ b/lib/jsonapi/resource.rb
@@ -307,7 +307,16 @@ module JSONAPI
       end
 
       def apply_includes(records, directives)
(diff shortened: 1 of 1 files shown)


lgebhardt:
Could you add a test that breaks without this code?

mmartinson:
I'd like to modify this address #484. Might it be better to have this explicit?

lgebhardt:
@mmartinson `apply_includes` has changed since this PR was submitted, so keep that in mind.

PR #249, opened 2015-06-10
Allow custom resource serializer

tamiswa:
A slight tweak to allow for my use case where I need to customize `JSONAPI::ResourceSerializer`.  I think it could be useful for others.

Does this work?


lgebhardt:
Thanks. This works for me.

Is your alternate serializer something that might be of use to others and thus something we should consider adding as an option?

tamiswa:
Let me refactor my serializers a bit first.  I wasn't doing much.  Just changing how links are generated in some cases.

PR #255, opened 2015-06-11
Add support for has_one :x, polymorphic: true

ggordon:
This patch adds some support for polymorphic associations. This should address issue #77 when completed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,12 @@
 # JSONAPI::Resources [![Build Status](https://secure.travis-ci.org/cerebris/jsonapi-resources.png?branch=master)](http://travis-ci.org/cerebris/jsonapi-resources)
 
-`JSONAPI::Resources`, or "JR", provides a framework for developing a server that complies with the 
+`JSONAPI::Resources`, or "JR", provides a framework for developing a server that complies with the
 [JSON API](http://jsonapi.org/) specification.
 
-Like JSON API itself, JR's design is focused on the resources served by an API. JR needs little more than a definition 
+Like JSON API itself, JR's design is focused on the resources served by an API. JR needs little more than a definition
 of your resources, including their attributes and relationships, to make your server compliant with JSON API.
 
-JR is designed to work with Rails 4.0+, and provides custom routes, controllers, and serializers. JR's resources may be 
+JR is designed to work with Rails 4.0+, and provides custom routes, controllers, and serializers. JR's resources may be
 backed by ActiveRecord models or by custom objects.
 
 ## Demo App
@@ -15,7 +15,7 @@ We have a simple demo app, called [Peeps](https://github.com/cerebris/peeps), av
 
 ## Client Libraries
 
-JSON API maintains a (non-verified) listing of [client libraries](http://jsonapi.org/implementations/#client-libraries) 
+JSON API maintains a (non-verified) listing of [client libraries](http://jsonapi.org/implementations/#client-libraries)
 which *should* be compatible with JSON API compliant server implementations such as JR.
 
 ## Installation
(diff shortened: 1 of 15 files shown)


test/unit/serializer/polymorphism_test.rb
```diff
@@ -0,0 +1,125 @@
...
+              self: '/pictures/1'
+            },
+            attributes: {
+              name: 'enterprise_gizmo.jpg'
+            },
+            relationships: {
+              imageable: {
+                links: {
+                  self: '/pictures/1/relationships/imageable',
+                  related: nil
```
lgebhardt: Is this the relationship related link you are questioning how to handle? Would `'/pictures/1/imageable'` work?

ggordon:
Yes, that's the one. The problem with `/pictures/1/imageable` is that there is no `ImageableResource` or `ImageableController`, so sometimes, in this case, the imageable route would serialize a document and sometime a product, and I haven't figured out where that fits in. The problem with using `/pictures/1/document` is that it's _very difficult_ to know the all the possible routes to create.

lgebhardt:
If we were to move the `get_related_resources` routes onto the parent resource's controller that would preclude the need for the polymorphic type controller and resource. Then `/pictures/1/imageable` could return resources of mixed type. We would need to expand support in the fields, sort, and filters to use the 'type dot field' notation and validate against the related resource types. That might require the polymorphic associations to know all the possible types, but I suspect we could find a work around to declaring that.

ggordon:
I was surprised to find out that related resources work if empty `ImageableResource` and `ImageablesController` are created, definitely some magic beyond my understanding. Can even update a polymorphic association, but there is probably a better way to determine/set type on the association.

lgebhardt:
That looks like it's working fairly well. A couple of questions:

What happens if you create a route to `imageable`? My guess is any actions directly on the ImageableController will fail as there isn't an imageable model. If that's the case I'm a bit worried it may be confusing to users.

Can we auto create the `imageable_type=` method on the resource when the association is polymorphic? Maybe with an override if the model attribute is different? I'm not that familiar with rails polymorphism, so maybe this isn't at all standard and thus a bad idea.

What else do you think needs to be added for completeness?

ggordon:
Yeah, I'm not completely happy with having to create the `ImageablesController` and the routes would probably fail. I'll add a few more tests and clean up the code and docs and then we can reevaluate.

barelyknown:
Hey @ggordon. I need to have this feature implemented very soon, so I'd be happy to help. How could I best assist you?

ggordon:
@barelyknown Thanks for the offer. Is anything missing? As it stands now it supports the functionality I need. Let me know if I'm missing something.

ggordon:
@lgebhardt My editor _cleaned up_ the Readme, and I didn't notice until after I pushed. Let me know if I should undo.

lgebhardt:
@ggordon Yes, that would probably be good to undo the readme. What else changed in your latest commit?

ggordon:
@lgebhardt Done. Added polymorphic_type method to Association, fixed the docs, and added some tests.

barelyknown:
@lgebhardt How do you feel about this PR and https://github.com/cerebris/jsonapi-resources/pull/274?

lgebhardt:
@barelyknown These look like a good start to supporting polymorphism. Sorry for the delay in merging them. I'll see if I can get them reviewed quickly once you address @dgeb's comments.

ggordon:
@lgebhardt @barelyknown Should I close this PR? It looks like PR #274 includes this and more.

barelyknown:
I think so.

PR #260, opened 2015-06-18
Allow passing routing options to jsonapi_relationships

biomancer:
This allows passing routing options to jsonapi_relationships, which is useful if you are building read-only API, and want to limit generated methods to GET-only:

```
with_options(only: [:index, :show]) do
    jsonapi_resources :contacts do
      jsonapi_relationships
    end
end
```

`jsonapi_links` and `jsonapi_related_resources` already support these options, so it seemed logical to add it here too.


lgebhardt:
Thanks!

PR #266, opened 2015-06-19
Allow passing options to attributes the same way as to attribute

biomancer:
Like this

```
attributes :starts_at, :ends_at, format: :timestamp
```

### DIFF
--- a/lib/jsonapi/resource.rb
+++ b/lib/jsonapi/resource.rb
@@ -278,8 +278,9 @@ module JSONAPI
 
       # Methods used in defining a resource class
       def attributes(*attrs)
+        options = attrs.extract_options!.dup
         attrs.each do |attr|
-          attribute(attr)
+          attribute(attr, options)
         end
       end
 


lgebhardt:
Great addition. Thanks!

butchmarshall:
I am passing..

`{..., "status-at"=>1512650542}`

with JSONAPI::Resource defining 

`attributes :status_at, format: :timestamp`

and get:

```
Internal Server Error: undefined method `unformat' for nil:NilClass /home/butch/.rvm/gems/ruby-2.4.2/bundler/gems/jsonapi-resources-3a02a4d2e149/lib/jsonapi/request_parser.rb:580:in `unformat_value'
/home/butch/.rvm/gems/ruby-2.4.2/bundler/gems/jsonapi-resources-3a02a4d2e149/lib/jsonapi/request_parser.rb:513:in `block (2 levels) in parse_params'
/home/butch/.rvm/gems/ruby-2.4.2/gems/actionpack-4.2.10/lib/action_controller/metal/strong_parameters.rb:185:in `each_pair'
/home/butch/.rvm/gems/ruby-2.4.2/gems/actionpack-4.2.10/lib/action_controller/metal/strong_parameters.rb:185:in `each_pair'
```

senid231:
@butchmarshall it's because `jsonapi-resources` doesn't provide custom value formatters
you need to create your own
something like this:
```ruby
class TimestampValueFormatter < JSONAPI::ValueFormatter
  class << self
    def format(raw_value)
      raw_value.to_i unless raw_value.nil?
    end

    def unformat(raw_value)
      Time.at(raw_value) unless raw_value.nil?
    end
  end
end
```

but I agree that error `undefined method `unformat'` is wrong in this case
we should reset to default value formatter and provide warning or raise correct error.
In any case I think it should be in a separate issue

@lgebhardt what do you think we should do in this case?

PR #267, opened 2015-06-19
Fix routing with non-array `except` value

alyssais:
A route like `jsonapi_resources :users, except: :destroy` causes an unhandled exception because jsonapi-resources doesn't allow a single non-array `except` value, which is inconsistent with Rails.

This PR fixes that, allowing both array and non-array `except` values.


lgebhardt:
Thanks!

PR #272, opened 2015-06-24
Remove unused @key attribute

pmukerji:
It seems that this @key attribute is no longer used (replaced by @foreign_key which is set in the base class from the options hash in https://github.com/cerebris/jsonapi-resources/commit/7b2459a6e83d923572e747b1089b7f7e2de7811d)


lgebhardt:
Thanks!

PR #274, opened 2015-06-25
Add support for creating resources with polymorphic has one.

barelyknown:
This simply adds to PR https://github.com/cerebris/jsonapi-resources/pull/255 from @ggordon. I added support for creating resources with a polymorphic `has_one`. I don't like the code, but I wanted to change as little as possible for now since I need this feature ASAP. We should refactor to handle the creation of all `has_one` relationships the same way (whether or not they're polymorphic), but I think that we can merge prior to that since we'll have good test coverage and the public API feels right.

### DIFF
--- a/lib/jsonapi/request.rb
+++ b/lib/jsonapi/request.rb
@@ -406,9 +406,7 @@ module JSONAPI
           when 'relationships'
             value.each do |link_key, link_value|
               param = unformat_key(link_key)
-
               association = @resource_klass._association(param)
-
               if association.is_a?(JSONAPI::Association::HasOne)
                 if link_value.nil?
                   linkage = nil
@@ -424,7 +422,12 @@ module JSONAPI
 
                 unless links_object[:id].nil?
                   association_resource = Resource.resource_for(@resource_klass.module_path + unformat_key(links_object[:type]).to_s)
-                  checked_has_one_associations[param] = association_resource.verify_key(links_object[:id], @context)
+                  association_id = association_resource.verify_key(links_object[:id], @context)
+                  if association.polymorphic?
+                    checked_has_one_associations[param] = { id: association_id, type: unformat_key(links_object[:type].to_s) }
+                  else
+                    checked_has_one_associations[param] = association_id
+                  end
                 else
                   checked_has_one_associations[param] = nil
                 end
@@ -544,8 +547,8 @@ module JSONAPI
                              context: @context,
                              resource_id: parent_key,
                              association_type: association_type,
-                             key_value: verified_param_set[:has_one].values[0],
-                             key_type: data['type']
+                             key_value: verified_param_set[:has_one].values[0][:id],
+                             key_type: verified_param_set[:has_one].values[0][:type]
                            }
                          )
       elsif association.is_a?(JSONAPI::Association::HasOne)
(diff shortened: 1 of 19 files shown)


lib/jsonapi/request.rb
```diff
@@ -536,8 +537,21 @@ def parse_add_association_operation(data, association_type, parent_key)
...
+                           resource_klass,
+                           {
+                             context: @context,
+                             resource_id: parent_key,
+                             association_type: association_type,
+                             key_value: verified_param_set[:has_one].values[0][:id],
+                             key_type: verified_param_set[:has_one].values[0][:type]
+                           }
+                         )
+      elsif association.is_a?(JSONAPI::Association::HasOne)
```
dgeb: I'd prefer to see the `HasOne` check first, then check `association.polymorphic?` within that block. We could then follow suit for `HasMany` relationships later.

dgeb:
I'd eventually like to allow restrictions on possible `type`s in a polymorphic relationship. This doesn't have to make it into this PR, but we should think about the interface. I suppose it could be a separate param from `polymorphic: true`, such as `allowed_types`.

dgeb:
All in all, this looks like a promising start on polymorphism - nice work @ggordon and @barelyknown!

lib/jsonapi/request.rb
barelyknown: Ok, I added that.

barelyknown:
I'm going to take a crack at support for `has_many` polymorphism now. We'll see how it goes.

barelyknown:
That last commit adds polymorphism to the "read" side of `has_many` relationships. I'm not sure why there are merge conflicts - I'm not seeing that on my side.

ggordon:
@barelyknown I had merge conflicts in #255 with the Readme, from PR #277, those are probably the same.

barelyknown:
Ugh. The conflicts are a mess now. I'll see if I can work it out.

lgebhardt:
Sorry, I know. I'm going through them now. So if you want to hold off I'll see if I can get it all resolved.

lgebhardt:
I pushed the new rebase/274 branch up.

I hope I got it all correct. All the tests pass, and there are the same two lines that are not covered.

barelyknown:
Oh, good. Thanks for doing that - I know that it was a pain!

lgebhardt:
@barelyknown It was my fault for not merging this earlier.

lgebhardt:
Thanks!

Merged by #288

arcreative:
@lgebhardt I'm having an issue on `0.8.3` where trying to include the polymorphic relationship results in `JSONAPI: Could not find resource 'subject'. (Class Api::V1::SubjectResource not found)`.  The resource is supposed to be variable, not understanding why it's looking for `SubjectResource` when polymorphic is set to true...  If I force the type with `has_one :subject, polymorphic: true, class_name: 'Feedback'`, the included record has a blank type (data[0].relationships.subject = `{type: "", id: "76"}`).  Either I'm doing something wrong, or...

Resource definition:
`has_one :subject, polymorphic: true` 
Fails with `JSONAPI: Could not find resource 'subject'. (Class Api::V1::SubjectResource not found`
`has_one :subject, polymorphic: true, class_name: 'Feedback' # Obviously not ideal, just testing`
Returns invalid payload when including resource, e.g. `data[0].relationships.subject = {type: "", id: "76"}`

Model definition:
`belongs_to :subject, polymorphic: true`
All polymorphic stuff seems to otherwise be working normally

PR #288, opened 2015-07-05
Rebase/274 Add support for creating resources with polymorphic has one.

lgebhardt:
This is a rebase of #274.

@barelyknown @ggordon please review to make sure I didn't miss something.

I also removed the code that wasn't being covered by the tests. If you could check that out and make sure it won't affect things in your code that would be great. If we do need it in there we should make sure we test it.


barelyknown:
I just tested on a ~large app that depends on the polymorphic functionality and the tests passed. I also reviewed the two methods that you removed and that's fine. They aren't used.

PR #290, opened 2015-07-06
Configure exception class whitelist to be reraised by operations processors.

barelyknown:
For example, if you use Pundit for authorization, you might
raise a `Pundit::NotAuthorizedError` at some point during operations
processing. If you want to use Rails' `rescue_from` macro to
catch this error and render a 403 status code, you should add
the `Pundit::NotAuthorizedError` to the `exception_class_whitelist`.

`JSONAPI.configuration.exception_class_whitelist = [Pundit::NotAuthorizedError]`

### DIFF
--- a/README.md
+++ b/README.md
@@ -272,7 +272,7 @@ end
 ```ruby
 class BookResource < JSONAPI::Resource
 
-  # Only book_admins may see unapproved comments for a book. Using 
+  # Only book_admins may see unapproved comments for a book. Using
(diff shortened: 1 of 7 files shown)


lgebhardt:
Thanks! This adds some nice flexibility that was inadvertently lost with the InternalServer error change.

barelyknown:
I agree. It's nice to gracefully rescue internal server errors, but having the option to raise/rescue some classes is pretty important in my experience.

PR #292, opened 2015-07-07
Readme: s/links/relationships/ ?

iamvery:
As you all can tell, I'm discovering a lot about JSON API as I go :). I _think_ there was a change to the standard where the generated routes on how named "relationships" rather than "links". I started to make a change to the README, but I suspect it may not be complete. I'm happy to finish it up with some direction. Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -916,9 +916,9 @@ gives the following routes
 
 ```
                      Prefix Verb      URI Pattern                                               Controller#Action
-contact_links_phone_numbers GET       /contacts/:contact_id/links/phone-numbers(.:format)       contacts#show_association {:association=>"phone_numbers"}
-                            POST      /contacts/:contact_id/links/phone-numbers(.:format)       contacts#create_association {:association=>"phone_numbers"}
-                            DELETE    /contacts/:contact_id/links/phone-numbers/:keys(.:format) contacts#destroy_association {:association=>"phone_numbers"}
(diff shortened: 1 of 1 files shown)


lgebhardt:
Good catch. The standard did change from links to relationships and I forgot to update the readme to reflect that.

What you have looks correct. What else do you think might be needed?

lgebhardt:
I guess we should also rename the `jsonapi_links` method to `jsonapi_relationships` and keep `jsonapi_links` as an alias for existing apps.

iamvery:
I'm not sure what, if anything else would be needed for this. I just didn't want to act like I was confident I covered applicable readme changes with completeness :blush:.

The docs under ["nested resources"](https://github.com/cerebris/jsonapi-resources#nested-routes) indicate that `jsonapi_relationships` is already a thing. That might make it tricky for such a change.

> Or, you can add the default set of nested routes using the jsonapi_relationships method. For example:
> 
> ```
> Rails.application.routes.draw do
>   jsonapi_resources :contacts do
>     jsonapi_relationships
>   end
> end
> ```

:confused: I'd certainly say any changes like this would be covered under a separate issue/pr. You want to write something up to track/propose such a change (if even "possible")?

lgebhardt:
You're right, I had forgotten we already used that. Thanks for this fix, and catching that.

iamvery:
:100: thank you for the excellent support! You're already earning internet points in our company Slack :joy:

PR #302, opened 2015-07-09
Pass `context` into `records_for` through options

andresberrios:
I wasn't getting the context in the options hash when overriding the `records_for` method.
I'm not sure if this is the right way to fix it, please let me know if it's ok.
Thanks!


lgebhardt:
Thanks! I think this is good for now, but it does remind me that I'd like to revisit the options/context handling throughout the code.

PR #308, opened 2015-07-13
Prefer public_send to collaborating objects

gschorkopf:
### Background

While upgrading to a recent release, we noticed due to failing tests that the resource object was accessing a private interface on our model. We traced this back to the `send` calls. The fix for our code was to specify a `:relation_name` option so that the appropriate interface is accessed. It would be nice for the relation object to complain earlier about the missing interface which was working by chance only because we had a private method matching the message it sent.
### Note

We've chosen to keep some of the `send` calls sent to the object itself. We did note that the jsonapi-resources test suite still passed even when changing these calls to `public_send`. However, we didn't feel confident enough in our knowledge of these calls to make the change here.

### DIFF
--- a/lib/jsonapi/link_builder.rb
+++ b/lib/jsonapi/link_builder.rb
@@ -58,7 +58,7 @@ module JSONAPI
 
     def engine_path_from_resource_class(klass)
       path_name = engine_resources_path_name_from_class(klass)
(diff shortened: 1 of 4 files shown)


lgebhardt:
Thanks! I think using `public_send` over `send` is a better default now that we've eliminated ruby 1.8.7 from support.

PR #311, opened 2015-07-14
Include Code Climate badge

pmukerji:
Code Climate is a tool for evaluating code quality in a repository, and also has integrations with GitHub so you can look for quality/security issues that arise within Pull Requests (See: http://docs.codeclimate.com/article/213-github-pull-request-integration)

Additionally:

There have also been some commits recently regarding the Ruby Style Guide. Code Climate has a RuboCop engine you can enable that can auto-run RuboCop on Pull Requests and to catch those details:

https://github.com/codeclimate/codeclimate-rubocop
https://github.com/bbatsov/rubocop/blob/master/config/default.yml

What do you think about using such a tool?


lgebhardt:
Thanks! We'll see how it goes. It's probably worth adding RuboCop as well. I'll look into it.

PR #315, opened 2015-07-15
Use source for errors instead of path

pmukerji:
The JSON API format specifies using a `source` key to reference the source
of the error in Error Objects (http://jsonapi.org/format/#error-objects).

Currently the key `path` is being used.

At the moment we are also getting a `path` that looks like: e.g. `/email` for an attribute `email`

The JSON API format specifies:

> `pointer`: a JSON Pointer [RFC6901] to the associated entity in the request document [e.g. "/data" for a primary data object, or "/data/attributes/title" for a specific attribute].

Should we update to specify that it is an attribute like: `/data/attributes/email`?

also I wasn't clear if we should also have `pointer` specified, to make it clear it is a JSON Pointer instead of a parameter as a string.

### DIFF
--- a/lib/jsonapi/exceptions.rb
+++ b/lib/jsonapi/exceptions.rb
@@ -312,7 +312,7 @@ module JSONAPI
                                  status: :unprocessable_entity,
                                  title: "#{format_key(element[0])} - #{message}",
                                  detail: message,
-                                 source: "/#{element[0]}")
+                                 source: { pointer: "/#{element[0]}" })
             end
           )
         end
--- a/lib/jsonapi/exceptions.rb
+++ b/lib/jsonapi/exceptions.rb
@@ -294,9 +294,10 @@ module JSONAPI
     end
 
     class ValidationErrors < Error
-      attr_accessor :messages
-      def initialize(messages)
+      attr_accessor :messages, :resource_associations
+      def initialize(messages, resource_associations)
         @messages = messages
+        @resource_associations = resource_associations
         @key_formatter = JSONAPI.configuration.key_formatter
       end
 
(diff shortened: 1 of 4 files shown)


lgebhardt:
Good catch. I didn't notice those changes to spec. As I read it `source` could either contain the source of the error as a string or a sub key of either `pointer` or `parameter` which has the details. I'd say we should use these optional keys as they remove some ambiguity.

pmukerji:
Cool I've added the `pointer` sub key, but still need to check how best to add the path to a specific attribute:

`/data/attributes/email` instead of `/email`. I'm wondering what the pointer should be for an invalid relationship? 

e.g. for creating a `BookResource` and specifying a (`has_one`) `LibraryResource` that doesn't exist

``` json
{
  "data": {
    "type": "books",
    "relationships": {
      "library": {
        "data": { "type": "libraries", "id": "4" }
      }
    }
}
```

``` json
{
  "errors": [
    {
      "title": "library - can't be blank",
      "detail": "can't be blank",
      "id": null,
      "href": null,
      "code": 100,
      "source": "/library",
      "links": null,
      "status": "unprocessable_entity"
    }
  ]
}
```

Here I guess `source` should be something like: `/data/relationships/library` or `/data/relationships/library/data/id` ?

pmukerji:
@lgebhardt - Took a shot at implementing the more specific error pointer

lgebhardt:
Thanks. Other than my one comment above I think this is ready to merge.

pmukerji:
Great, I initialized with just the resource now, and did some minor refactoring

lib/jsonapi/exceptions.rb
```diff
@@ -305,16 +307,26 @@ def format_key(key)
...
-              JSONAPI::Error.new(code: JSONAPI::VALIDATION_ERROR,
-                                 status: :unprocessable_entity,
-                                 title: "#{format_key(element[0])} - #{message}",
-                                 detail: message,
-                                 path: "/#{element[0]}")
-            end
-          )
+        error_messages.map do |attr_key, messages|
+          messages.map { |message| json_api_error(attr_key, message) }
+        end.flatten
```
fphilipe: Could replace the first `map` with `flat_map` and get rid of the `flatten`.

pmukerji: :+1: thanks! Updated.

lgebhardt:
Thanks again!

pmukerji:
No problem! Thanks for the quick merge and release

Just a small thing from the release notes:

```
Errors now use path instead of source parameter, as per the JSON API spec.
```

I guess path and source are reversed here. Not sure if you can edit the release notes, but it could read:

```
Errors now use source instead of path parameter, as per the JSON API spec.
```

lgebhardt:
@pmukerji Thanks. Updated.

PR #328, opened 2015-07-20
Init from model

### LINKED ISSUE #316: Infer class_name for associations
Consider the following model:

``` ruby
class User
  has_many :billing_addresses, class_name: 'Address'
end
```

The resource would have to be:

``` ruby
  has_many :billing_addresses, class_name: 'Address'
end
```

Would it be possible to infer the `class_name` from the model to avoid having to duplicate information?

lgebhardt:
Fixes #316

In the future I'd like to also initialize the foreign key, however this will require a bit of rework because of differences between how the model and the resource use the values.

### DIFF
--- a/README.md
+++ b/README.md
@@ -262,7 +262,7 @@ end
 
 The relationship methods (`relationship`, `has_one`, and `has_many`) support the following options:
 
- * `class_name` - a string specifying the underlying class for the related resource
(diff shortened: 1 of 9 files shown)


dgeb:
Looks good now - thanks @lgebhardt :shipit:

Issue #330, opened 2015-07-21
Custom Operations


hinshun:
As per discussed in #129, I would like to do other tasks within a CRUD operation.

For my specific use case, I would like to create related resources upon creating a parent resource. The API is still strictly CRUD and compliant with JSONAPI, but under the hood I would like to have business logic that lives on the server side rather than the client. I found that decoupling this sort of code into service objects have been great for testing and maintainability, and it looks like Operation is exactly that.

Perhaps we can refactor it such that users of this gem can write their own operations for specific resources?

debreczeni:
:+1:

PR #331, opened 2015-07-21
Refactor request to support custom operations

hinshun:
Addresses #330.

### DIFF
--- a/lib/jsonapi/operation.rb
+++ b/lib/jsonapi/operation.rb
@@ -11,318 +11,331 @@ module JSONAPI
 
     def apply
     end
-  end
 
-  class FindOperation < Operation
-    attr_reader :filters, :include_directives, :sort_criteria, :paginator
+    class << self
+      def resource_operation_for(resource_klass, operation)
+        resource_operation_name = "#{resource_klass.to_s.underscore}_#{operation}"
+        resource_operation = Operation.operation_for(resource_operation_name)
+        resource_operation || Operation.operation_for(operation)
+      end
 
-    def initialize(resource_klass, options = {})
-      @filters = options[:filters]
-      @include_directives = options[:include_directives]
-      @sort_criteria = options.fetch(:sort_criteria, [])
-      @paginator = options[:paginator]
-      @transactional = false
-      super(resource_klass, options)
-    end
-
-    def record_count
-      @_record_count ||= @resource_klass.find_count(@resource_klass.verify_filters(@filters, @context),
-                                                    context: @context,
-                                                    include_directives: @include_directives)
-    end
-
-    def pagination_params
-      if @paginator && JSONAPI.configuration.top_level_links_include_pagination
-        options = {}
-        options[:record_count] = record_count if @paginator.class.requires_record_count
-        return @paginator.links_page_params(options)
-      else
-        return {}
+      def operation_for(operation)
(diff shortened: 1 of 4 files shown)


hinshun:
Opted for refactoring `JSONAPI::FindOperation` out of the JSONAPI namespace, etc, to stay consistent with the `def class_for(type)` you have in other classes. Otherwise I will have to do something like:

``` ruby
def resource_operation_for(resource_klass, operation)
  # ...
  resource_operation || Operation.operation_for("JSONAPI/#{operation}")
end
```

The master branch's specs are passing atm, so if this looks like the right approach I will write specs next.

lgebhardt:
@hinshun I think this a good approach. It might be nice to take it even farther and move the Operation setup logic from the request to the Operations themselves (initialize taking the request object, which could contain just common parsing logic). Then the correct operation could be defined on the route, or probably better each resource could map an action to an operation (could use the context to choose the correct operation). But either way would allow you to create your custom operations without altering the request class. It would also remove a lot of the code smell from Request class, which has been on my to do list.

How are you assigning your new operations to actions?

hinshun:
@lgebhardt Could you clarify what you mean by `resource could map an action to an operation`? I'm not sure what `action` means in the context of this gem. Do you mean to move the setup_.*_action methods into the resource class?

You can override existing operations like so:

``` ruby
class EventCreateResourceOperation < CreateResourceOperation
  def apply
    # ...
    resource.my_method!
    # ...
    return JSONAPI::ResourceOperationResult.new((result == :completed ? :created : :accepted), resource)

  rescue JSONAPI::Exceptions::Error => e
    return JSONAPI::ErrorsOperationResult.new(e.errors[0].code, e.errors)
  end

  end
end
```

`Operation.resource_operation_for(resource_klass, operation)` will first search for a operation with the `resource_klass` in the front before defaulting the the base operation.

Though with this, we can refactor it so that we don't override `apply`, so that returning the result wrapper and catching the exception can be handled by the base operation class.

lgebhardt:
@hinshun I was thinking we move the setup_.*_action methods into the operations themselves. There might be some issues here with supporting multiple operations per request which we should consider.

The way it works currently is the `action` comes in from the controller (and is set by the route). The controller creates the `request` object which uses the action to know which operations to create, and initializes them. 

What I was trying to understand is where your new operations are being created and how the code knows to use EventCreateResourceOperation instead of CreateResourceOperation. I had missed that it was in prepending the resource name to the operation.

What I was proposing was allowing the resource to control which operations are used for each action. The advantage I see to this approach is you could override one of the operations and use it for more than one resource without needing to declare an operation per resource (which I know I would be prone to doing).

hinshun:
@lgebhardt I see what you mean now. Regarding multiple operations per request, in what scenario would you need to support a 1:n relationship? I think in that case you could write an operation that uses other smaller operations so that the relationship remains 1:1.

I agree with decoupling the operations from a specific resource, but I wonder if you could delve into why it should live in the resource vs its respective controller?

lgebhardt:
@hinshun When support for batch requests are added to JSON API we will want to extend JR to handle those. There may be other instances where a request will map to multiple operations.

The main reason I see adding it to the resource and not the controller is that maybe when we support some JSON API extensions, such as Batch, a request will need to have Operations for multiple different resource types.

PR #333, opened 2015-07-23
Enable exception superclasses to be whitelisted.

barelyknown:
This makes it easy to reraise any type of exception (or groups of exception classes). I needed this for integration with a application monitoring service.


lgebhardt:
Good idea. Thanks!

PR #336, opened 2015-07-24
fix validation error message attribute format

Azdaroth:
The attributes needs to be properly formatted in validation messages as well (discovered it when using Ember Data server side validation messages), this PR fixes the problem.


lgebhardt:
Thanks!

PR #337, opened 2015-07-24
fix singleton resource update action to not require id

Azdaroth:
The singleton resource doesn't have `id` part in URL for update actions, so I think there's no need to check for its' presence and do any comparison with the `id` from the `data` in that case, for normal resources (non-singleton) the `id` is required on URL level anyway, so I think it's safe and shouldn't cause any regression.

### DIFF
--- a/lib/jsonapi/request.rb
+++ b/lib/jsonapi/request.rb
@@ -104,7 +104,7 @@ module JSONAPI
     def setup_update_action(params)
       parse_fields(params[:fields])
       parse_include_directives(params[:include])
-      parse_replace_operation(params.require(:data), params.require(:id))
+      parse_replace_operation(params.require(:data), params[:id])
     end
 
     def setup_destroy_action(params)
@@ -563,7 +563,7 @@ module JSONAPI
       end
     end
 
-    def parse_single_replace_operation(data, keys)
+    def parse_single_replace_operation(data, keys, id_key_presence_check_required: true)
       fail JSONAPI::Exceptions::MissingKey.new if data[:id].nil?
 
       type = data[:type]
@@ -572,7 +572,7 @@ module JSONAPI
       end
 
       key = data[:id]
-      unless keys.include?(key)
+      if id_key_presence_check_required && !keys.include?(key)
         fail JSONAPI::Exceptions::KeyNotIncludedInURL.new(key)
       end
 
@@ -596,7 +596,8 @@ module JSONAPI
           parse_single_replace_operation(object_params, keys)
         end
       else
-        parse_single_replace_operation(data, [keys])
+        parse_single_replace_operation(data, [keys],
+                                       id_key_presence_check_required: keys.present?)
       end
 
     rescue JSONAPI::Exceptions::Error => e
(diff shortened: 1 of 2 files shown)


lgebhardt:
I don't see the JSON API spec supporting singleton resources. I'd like to see support for this in the spec before we add official support in this project.

Azdaroth:
Hmm, I don't see it either. Any idea if it's going to be supported in JSON API?

dgeb:
So, the spec isn't opinionated about URLs. A resource could exist at `/preferences/` for instance, and some would consider that a "singleton".

What the spec doesn't support is a resource without a `type` and `id`. For example, a "preferences" resource would still need a `type` and `id`, even if only one preferences resource could ever be accessed by a particular user. These requirements have benefits though, such as consistent parsing on the client side and easier tracking of a user's particular "singleton" resource.

Azdaroth:
So in that case I think this PR complies with the JSON API - the `type` and `id` is still required in sent data, it's only about the `id` in the URL, which is properly generated with `jsonapi_resource` for update action, but the current `id` verification makes it impossible use.

lgebhardt:
It seems odd to have the id in the payload, but not the url. But I'll take a second look at it then in that light.

lib/jsonapi/request.rb
```diff
@@ -104,7 +104,8 @@ def setup_update_relationship_action(params)
     def setup_update_action(params)
       parse_fields(params[:fields])
       parse_include_directives(params[:include])
-      parse_replace_operation(params.require(:data), params.require(:id))
+      parse_replace_operation(params.require(:data), params[:id],
+                              id_key_presence_check_required: params[:id].present?)
```
lgebhardt: I think we could skip the `id_key_presence_check_required` parameter and do the check in `parse_replace_operation`. And since it's only for singletons the check could just be done when data isn't an array.

lib/jsonapi/request.rb
```diff
@@ -588,15 +589,17 @@ def parse_single_replace_operation(data, keys)
...
 
-    def parse_replace_operation(data, keys)
+    def parse_replace_operation(data, keys, id_key_presence_check_required: true)
       if data.is_a?(Array)
         fail JSONAPI::Exceptions::CountMismatch if keys.count != data.count
 
         data.each do |object_params|
-          parse_single_replace_operation(object_params, keys)
+          parse_single_replace_operation(object_params, keys,
+                                         id_key_presence_check_required: id_key_presence_check_required)
```
lgebhardt: You could omit `id_key_presence_check_required` here since this case will never be a singleton, and `parse_single_replace_operation` provides a default value of true.

lib/jsonapi/request.rb
Azdaroth: Makes sense, fixed it.

lib/jsonapi/request.rb
Azdaroth: Done.

lgebhardt:
Looks good. Thanks for making those changes.

PR #339, opened 2015-07-24
update to support non-standard primary key has many relationships

### LINKED ISSUE #332: Custom primary key issue
Hello, I recently started playing around with jsonapi-resources for a project. I am hoping to convert an existing Rails 4 service to use JSON API. While I was experimenting with jsonapi-resources I ran into issues with a model that has a `primary_key` other than `id`. After some debugging I believe I tracked down the issues but before I take it too far I wanted to get some feedback to make sure I am not just doing something wrong.

This is all based on master where the latest commit was 4f2e4592efeb34d18a37fab66ba84688b087bf1a

Consider this simplified example:

```
# config/routes.rb
jsonapi_resources :users
jsonapi_resources :organizations
```

```
# app/resources/organization_resource.rb 
class OrganizationResource < JSONAPI::Resource
  attribute :name

  has_many :members, class_name: "User"
end
```

```
# app/resources/user_resource.rb
  primary_key :uuid
  attributes :uuid, :email

  has_many :orgs, class_name: "Organization"

  def self.verify_key(key, context = nil)
    key && String(key)
  end
end
```

The controllers are empty so I did not include them here.

As I worked though getting this up and running I ran into the following issues, all related to the UserResource uuid primary key.
1. Users routes were generated using `id` and `user_id` and jsonapi-resources was expecting `uuid` and `user_uuid`.
2. `organizations/:id/members` was trying to order by `users.id`
3. `users/:uuid/orgs` was trying to sort by `orgs.uuid`
4. `users/:uuid` added `users.uuid IS NULL` to the where clause

To solve these problems I made the following changes to jsonapi-resources locally:

To solve the first issue:

```
# ActionDispatch::Routing::Mapper#jsonapi_resources

options = resources.extract_options!.dup
options[:controller] ||= @resource_type
options.merge!(res.routing_resource_options)

# options[:param] = :id
options[:param] = res._primary_key

options[:path] = format_route(@resource_type)
```

To solve 2nd and 3rd issues:

```
# JSONAPI::Resource#_add_relationship

sort_criteria =  options.fetch(:sort_criteria, {})
unless sort_criteria.nil? || sort_criteria.empty?
  # order_options = self.class.construct_order_options(sort_criteria)
  order_options = relationship.resource_klass.construct_order_options(sort_criteria)
  records = resource_klass.apply_sort(records, order_options)
end

```

To solve final issue:

```
# JSONAPI::Request

def setup_show_action(params)
  parse_fields(params[:fields])
  parse_include_directives(params[:include])
  # @id = params[:id]
  @id = params[@resource_klass._primary_key]
  add_show_operation
end
```

Are these actual issues with jsonapi-resources or am I just doing something wrong?

Thanks!

macasek:
Fixes: https://github.com/cerebris/jsonapi-resources/issues/332

This PR fixes some issues when you have a model that specifies a non-standard primary key. See test/fixtures/active_record.rb for an example model.

The first commit fixes the get_related_resources where the order by cause of the sql query was using the wrong primary key.

The second commit resolve an issue where the code is incorrectly looking for a param postfixed with the primary key of the parent model (`_as_parent_key`) but the router generates routes that always end in _id.

### DIFF
--- a/lib/jsonapi/resource.rb
+++ b/lib/jsonapi/resource.rb
@@ -596,7 +596,7 @@ module JSONAPI
       end
 
       def _as_parent_key
-        @_as_parent_key ||= "#{_type.to_s.singularize}_#{_primary_key}"
+        @_as_parent_key ||= "#{_type.to_s.singularize}_id"
       end
 
       def _allowed_filters
--- a/lib/jsonapi/resource.rb
+++ b/lib/jsonapi/resource.rb
@@ -744,7 +744,7 @@ module JSONAPI
 
               sort_criteria =  options.fetch(:sort_criteria, {})
               unless sort_criteria.nil? || sort_criteria.empty?
-                order_options = self.class.construct_order_options(sort_criteria)
+                order_options = relationship.resource_klass.construct_order_options(sort_criteria)
                 records = resource_klass.apply_sort(records, order_options)
               end
 
--- a/test/controllers/controller_test.rb
+++ b/test/controllers/controller_test.rb
@@ -2645,7 +2645,7 @@ class Api::V1::MoonsControllerTest < ActionController::TestCase
                                   id: "1",
                                   type: "moons",
                                   links: {self: "http://test.host/moons/1"},
-                                  attributes: {name: "Titan", "description": "Best known of the Saturn moons."},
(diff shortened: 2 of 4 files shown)


lgebhardt:
Thanks!

PR #347, opened 2015-07-29
Fix camel case scope converting inside link_builder

Fredar:
Currently camelcase scopes are incorrectly evaluated, so that `AdminApi` becomes `adminapi` instead of `admin_api`. 
As a result `url_helper` receives invalid method and crashes the app.


lgebhardt:
Thanks!

PR #350, opened 2015-07-29
Fix syntax error on postgres custom select

### LINKED ISSUE #348: Autocount is breaking on custom SELECTs
The method `JSONAPI::Resource#find_count` should be changed to:

``` ruby
 def self.find_count(filters, options = {})
  filter_records(filters, options).count(:all)
 end
```

so the SELECT is cleared. This is how `ActiveRecord::Relation#size` is implemented, BTW.

This problem emerges as a syntax error on any custom select, specially the ones you make just to please Postgres’s GROUP BY clauses.

lgebhardt:
Change count to count(:all)

fixes #348

### DIFF
--- a/lib/jsonapi/resource.rb
+++ b/lib/jsonapi/resource.rb
@@ -483,7 +483,7 @@ module JSONAPI
       end
 
       def find_count(filters, options = {})
-        filter_records(filters, options).count
+        filter_records(filters, options).count(:all)
       end
 
       # Override this method if you have more complex requirements than this basic find method provides


lib/jsonapi/resource.rb
```diff
@@ -483,8 +483,8 @@ def sort_records(records, order_options)
       end
 
       def find_count(filters, options = {})
-        filter_records(filters, options).count
-      end
+        filter_records(filters, options).count(:all)
+     end
```
dgeb: did you drop a space here accidentally?

dgeb:
Looks good otherwise :+1:

lgebhardt:
Fixed

dgeb:
Thanks! :shipit:

PR #351, opened 2015-07-31
Return HTTP 400 Bad Request when include is present and not allowed

koppen:
From http://jsonapi.org/format/#fetching-includes:

> If an endpoint does not support the include parameter, it must respond with
> 400 Bad Request to any requests that include it.

This change makes jsonapi-resources adhere to cases where
`allowed_request_params` has been configured to not allow the `include`
parameter. Previously, this configuration option was silently ignored and
resources were included in the output regardless.

### DIFF
--- a/lib/jsonapi/request.rb
+++ b/lib/jsonapi/request.rb
@@ -196,6 +196,10 @@ module JSONAPI
     def parse_include_directives(include)
       return if include.nil?
 
+      unless JSONAPI.configuration.allowed_request_params.include?(:include)
+        fail JSONAPI::Exceptions::ParametersNotAllowed.new([:include])
+      end
+
       included_resources = CSV.parse_line(include)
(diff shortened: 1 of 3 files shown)


lgebhardt:
You have uncovered a bigger issue. The `allowed_request_params` configuration option is no longer being used (except in your new code). The original intent wasn't to allow includes (or the other params) to be excluded, but to provide a master list, that could be added to, of params to enforce with strong parameters. This use of strong params was removed a while ago, but I never took this configuration option out as I should have.

So I reworked this feature based on your PR, see #352. I think a boolean in the configuration for each of the optional parameters is clearer. 

``` ruby
JSONAPI.configure do |config|
  config.allow_include = false
  #config.allow_sort = true
  #config.allow_filter = true
end
```

koppen:
That's awesome and much better, cheers! :)

PR #353, opened 2015-08-02
Rails generator for resources

kevintraver:
Initial attempt at creating a Rails generator for JSONAPI resource files

### DIFF
--- a/lib/generators/jsonapi/USAGE
+++ b/lib/generators/jsonapi/USAGE
@@ -1,0 +1,8 @@
+Description:
+    Generator for JSONAPI Resources
+
+Example:
+    rails generate jsonapi:resource Post
+
+    This will create:
+        app/resources/post_resource.rb
--- a/lib/generators/jsonapi/templates/jsonapi_resource.rb
+++ b/lib/generators/jsonapi/templates/jsonapi_resource.rb
@@ -1,0 +1,4 @@
+<% module_namespacing do -%>
+class <%= class_name %>Resource < JSONAPI::Resource
+end
+<% end -%>
--- a/test/lib/generators/jsonapi/resource_generator_test.rb
+++ b/test/lib/generators/jsonapi/resource_generator_test.rb
@@ -1,0 +1,15 @@
+require 'test_helper'
+require 'generators/jsonapi/resource_generator'
+
+module Jsonapi
+  class ResourceGeneratorTest < Rails::Generators::TestCase
+    tests ResourceGenerator
+    destination Rails.root.join('../resources')
+    setup :prepare_destination
+
+    test "resource is created" do
+      run_generator ["post"]
+      assert_file 'app/resources/post_resource.rb'
+    end
+  end
+end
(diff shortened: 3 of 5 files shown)


lib/generators/jsonapi/templates/jsonapi_resource.rb
```diff
@@ -0,0 +1,4 @@
+<% module_namespacing do -%>
+class <%= class_name %>Resource > JSONAPI::Resource
```
lgebhardt: Should use <

``` ruby
class <%= class_name %>Resource < JSONAPI::Resource
```

lgebhardt:
Cool, thanks. Looks like a good start. This is now creating a resources directory in the project. We should add this to the .gitignore file or clean up the files after the tests.

Also it seems the resources are replaced if the generator is run again. I don't use the Rails generators much so I don't know if this is standard. But it seems risky as you could overwrite some of your work if it accidentally run it again (which I'm sure I would do at times with a missed keystroke in selecting command history).

I don't know if generators usually go this far, but it would be cool to interrogate the model and setup the default relationships and attributes. That's certainly not needed at first, but it seems useful.

kevintraver:
Thanks for the feedback. If you are inheriting from the standard rails generator it looks like get a confirmation to overwrite.

I agree it would be cool to have more integration. I think the next step would be to add parameters for attributes.

lgebhardt:
I just tested it out in a real project and it does warn you if the file exists. So no issue there. However it created the resource directory in the root of the project, and by convention the directory should be `app/resources`.

barelyknown:
Nice idea @kevintraver. I don't use generators all that often, but I think that this is a friendly feature for new users of `jsonapi-resources`.

lib/generators/jsonapi/USAGE
```diff
@@ -0,0 +1,8 @@
+Description:
+    Generator for JSONAPI Resources
+
+Example:
+    rails generate jsonapi:resource Post
+
+    This will create:
+        resources/post_resource.rb
```
lgebhardt: app/resources/post_resource.rb

test/lib/generators/jsonapi/resource_generator_test.rb
```diff
@@ -0,0 +1,15 @@
+require 'test_helper'
+require 'generators/jsonapi/resource_generator'
+
+module Jsonapi
+  class ResourceGeneratorTest < Rails::Generators::TestCase
+    tests ResourceGenerator
+    destination Rails.root.join('../resources')
```
lgebhardt, line 7: Could you put the generated files into a temp directory. Maybe `../test/temp/generators`. And then we can add that to .gitignore.

kevintraver, line 7: Im sure theres a proper way to teardown, I just need to find out how to do it.

kevintraver:
I've fixed the teardown/cleanup issue, and improved the tests.

lgebhardt:
Thanks!

jondkinney:
@kevintraver @lgebhardt any plans to add this generator into the Rails config.generators hooks so that it happens automatically (if configured to do so) when generating a rails resource? `rails g resource user` for example, would also generate a jsonapi resource.

PR #359, opened 2015-08-05
switch detail and title values in validation error

### LINKED ISSUE #358: ValidationErrors. why is detail message smaller than title?
current code is:

``` ruby
module JSONAPI
  module Exceptions
    class ValidationErrors < Error
      ...
      def json_api_error(attr_key, message)
        JSONAPI::Error.new(code: JSONAPI::VALIDATION_ERROR,
                           status: :unprocessable_entity,
                           title: "#{format_key(attr_key)} - #{message}",
                           detail: message,
                           source: { pointer: pointer(attr_key) })
      end
      ...
    end
  end
end
```

i think that title should contain only message. maybe we should switch it like:

```
title: message,
detail: "#{format_key(attr_key)} - #{message}",
```

senid231:
fixes #358


lgebhardt:
Thanks!

PR #362, opened 2015-08-06
Fix for when ActiveRecord is not loaded.

### LINKED ISSUE #360: ActiveRecord coupling fix
Just updated to latest master, and came across this error complaining about missing AR

```
  1) Supports API GET /supports?filter[proposal_id]=:proposal_id&filter[participant_id]=:participant_id support resource
     Failure/Error: subject { get "/supports?#{filter_params}", {}, headers  }
     NameError:
       uninitialized constant ActiveRecord
     # /Users/work/.rvm/gems/ruby-2.2.2/bundler/gems/jsonapi-resources-72168924065f/lib/jsonapi/resource.rb:689:in `block in _add_relationship'
     # /Users/work/.rvm/gems/ruby-2.2.2/bundler/gems/jsonapi-resources-72168924065f/lib/jsonapi/resource.rb:685:in `each'
     # /Users/work/.rvm/gems/ruby-2.2.2/bundler/gems/jsonapi-resources-72168924065f/lib/jsonapi/resource.rb:685:in `_add_relationship'
     # /Users/work/.rvm/gems/ruby-2.2.2/bundler/gems/jsonapi-resources-72168924065f/lib/jsonapi/resource.rb:346:in `has_one'
     # ./app/resources/support_resource.rb:4:in `<class:SupportResource>'
     # ./app/resources/support_resource.rb:3:in `<top (required)>'
```

This app is using mongodb/mongoid, and so far I had been fixing the AR dependency like this:

``` ruby
# app/controllers/my_resource_controller.rb
class MyResourceController < ApplicationController
  include JSONAPI::ActsAsResourceController

  def create_operations_processor
    TempMongoidOperationsProcessor.new
  end
end

# lib/temp_mongoid_operations_processor.rb
class TempMongoidOperationsProcessor < JSONAPI::OperationsProcessor

  private

    def transaction
      # if @transactional
      #   ActiveRecord::Base.transaction do
      #     yield
      #   end
      # else
      yield
      # end
    end

    def rollback
      # raise ActiveRecord::Rollback if @transactional
    end

    def process_operation(operation)
      # operation.apply(@context)
      operation.apply
    # rescue ActiveRecord::DeleteRestrictionError => e
    #   record_locked_error = JSONAPI::Exceptions::RecordLocked.new(e.message)
    #   return JSONAPI::ErrorsOperationResult.new(record_locked_error.errors[0].code, record_locked_error.errors)

    rescue Mongoid::Errors::DocumentNotFound
      record_not_found = JSONAPI::Exceptions::RecordNotFound.new(operation.associated_key)
      return JSONAPI::ErrorsOperationResult.new(record_not_found.errors[0].code, record_not_found.errors)
    end
end
```

Is my fix outdated? Is there a new recommended fix to get non-AR ORM/ODMs going?

lgebhardt:
Check if model is derived from ActiveRecord::Base by name, not class.

Hopefully fixes #360

### DIFF
--- a/lib/jsonapi/resource.rb
+++ b/lib/jsonapi/resource.rb
@@ -686,7 +686,7 @@ module JSONAPI
           check_reserved_relationship_name(attr)
 
           # Initialize from an ActiveRecord model's properties
-          if _model_class && _model_class < ActiveRecord::Base
+          if _model_class && _model_class.ancestors.collect{|ancestor| ancestor.name}.include?('ActiveRecord::Base')
             model_association = _model_class.reflect_on_association(attr)
             if model_association
               options[:class_name] ||= model_association.class_name


oliverbarnes:
:+1: works! thanks for the speedy fix

lib/jsonapi/resource.rb
```diff
@@ -686,7 +686,7 @@ def _add_relationship(klass, *attrs)
           check_reserved_relationship_name(attr)
 
           # Initialize from an ActiveRecord model's properties
-          if _model_class && _model_class < ActiveRecord::Base
+          if _model_class && _model_class.ancestors.collect{|ancestor| ancestor.name}.include?('ActiveRecord::Base')
             model_association = _model_class.reflect_on_association(attr)
```
dgeb, line 690: how about testing for the presence of `reflect_on_association` on the class instead?

lgebhardt, line 690: I was worried about the possibility that a poro might implement `reflect_on_association`, but the results may not support the `class_name` method. This seemed safer to me.

dgeb, line 690: I'd like poros to be able to support interfaces such as `reflect_on_association`, but this is fine. It's mostly an academic argument at this point, since we'd have to clearly document these specific interfaces for it to be practical.

PR #366, opened 2015-08-07
Add related resources pagination links

### LINKED ISSUE #303: Paginating nested resources
I'm running into an issue where pagination is ignored when you have nested resources. Is this a known issue?

For example, the following query returns 2 change logs:

> curl http://localhost:3000/v1/change-logs?page[size]=2

But if I curl the same resource nested within its parent, it returns all change logs:

> curl http://localhost:3000/v1/network-hosts/1/change-logs?page[size]=2

Another thing I noticed that's probably related is that the first request includes the pagination "links" attribute at the end of the response JSON, but the second request does not.

antonversal:
Currently there are not any pagination links when nested resources are requested. 
fixes #303

### DIFF
--- a/lib/jsonapi/link_builder.rb
+++ b/lib/jsonapi/link_builder.rb
@@ -31,8 +31,10 @@ module JSONAPI
       "#{ primary_resources_url }?#{ query_params.to_query }"
     end
 
-    def relationships_related_link(source, relationship)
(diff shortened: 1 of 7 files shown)


lgebhardt:
Thanks! Very much appreciated.

PR #372, opened 2015-08-10
fix bug when no related resource

antonversal:
Unfortunately my PR #366 has a bug. When there arn't any related resources 500 error is raised.
So this PR is going to fix the error. 
Additionally the fix removes pagination links from all responses with empty resources, which I think is good )

### DIFF
--- a/lib/jsonapi/operation.rb
+++ b/lib/jsonapi/operation.rb
@@ -185,7 +185,7 @@ module JSONAPI
                                               sort_criteria: @sort_criteria,
                                               paginator: @paginator)
 
-      return JSONAPI::RelatedResourcesOperationResult.new(:ok, source_resource, related_resource, options)
+      return JSONAPI::RelatedResourcesOperationResult.new(:ok, source_resource, @relationship_type, related_resource, options)
 
     rescue JSONAPI::Exceptions::Error => e
       return JSONAPI::ErrorsOperationResult.new(e.errors[0].code, e.errors)
--- a/lib/jsonapi/operation_result.rb
+++ b/lib/jsonapi/operation_result.rb
@@ -43,10 +43,11 @@ module JSONAPI
   end
 
   class RelatedResourcesOperationResult < ResourcesOperationResult
-    attr_accessor :source_resource
+    attr_accessor :source_resource, :_type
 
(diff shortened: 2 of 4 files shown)


lgebhardt:
I hadn't noticed the 500 error. Looks like a test for a valid response with an empty resultset was missing.

I think we want the `links` for empty result sets, both for related and primary resources. I think only the `next` link should be excluded and the `first` and `last` would both be the same. For example we should add a test like this:

``` ruby
  def test_pagination_empty_results
    Api::V2::BookResource.paginator :offset
    Api::V2::BookCommentResource.paginator :offset
    get '/api/v2/books?filter[id]=2000&page[limit]=10'
    assert_equal 200, status
    assert_equal 0, json_response['data'].size
    assert_nil json_response['links']['next']
    assert_equal 'http://www.example.com/api/v2/books?filter[id]=2000&page%5Blimit%5D=10&page%5Boffset%5D=0', json_response['links']['first']
    assert_equal 'http://www.example.com/api/v2/books?filter[id]=2000&page%5Blimit%5D=10&page%5Boffset%5D=0', json_response['links']['last']
  end
```

My reasoning for including the links is that the response is valid and the links should be able to reproduce the resultset, even if it's empty.

The `test_pagination_related_resources_without_related` should have a similar set of asserts:

``` ruby
  def test_pagination_related_resources_without_related
    Api::V2::BookResource.paginator :offset
    Api::V2::BookCommentResource.paginator :offset
    get '/api/v2/books/10/book_comments'
    assert_equal 200, status
    assert_nil json_response['links']['next']
    assert_equal 'http://www.example.com/api/v2/books/10/book_comments?page%5Blimit%5D=10&page%5Boffset%5D=0', json_response['links']['first']
    assert_equal 'http://www.example.com/api/v2/books/10/book_comments?page%5Blimit%5D=10&page%5Boffset%5D=0', json_response['links']['last']
  end
```

It looks like there's a bug in the offset and paged paginator with small and empty resultsets, so the above tests won't pass even if you emit the `links`. I'll put a PR together for this now.

antonversal:
ok, I'll update the pr

antonversal:
updated, tests are not passed because of #373.

lgebhardt:
Thanks!

PR #373, opened 2015-08-10
Fix pagination links for small and empty result sizes.

lgebhardt:
When looking at #372 that the paginators were generating the invalid parameters with small and empty sized results.


antonversal:
:+1: I noticed this too before ) and wanted to fix it. Thanks!

PR #388, opened 2015-08-15
Add entry about including related resources

oliverbarnes:


barelyknown:
I like having this in the README. Can you add examples of each?

oliverbarnes:
Np, I added an example for the include params, but the linked `serialize_to_hash` section seems to be covered?

barelyknown:
Great. Thanks.

Issue #364, opened 2015-08-06
STI - mixed endpoint


adamworrall:
I have a group of STI child models that I'd like to be able to return from a single endpoint. Right now I have a resource set up for the parent model, and inherited resources for each child model, and each of those endpoints is working as expected.

What I'd like to do, though, is to have one endpoint that uses the appropriate resource for each model returned in an aggregate query. For example, the parent resource is `Articles` and I'd like a call to `GET Articles` to return the various children, serialized according to their ActiveRecord `type`. Is there any way to accomplish this? The codebase has changed a lot since I looked at it in April, but it seems like the assumption of one resource -> one route is pretty locked in. 

The ability to at least get an index of all the different types under one endpoint is fairly important, otherwise I could end up having to have 10+ calls made for a client to fetch recent articles.

adamworrall:
I was able to get this working with a small change to `JSONAPI::Resource`.

I added a flag similar to `abstract`, `sti_parent_resource`. 

The `find` method gets a small change at the end:

```
  (...)
  records.map do | model |
    if @sti_parent_resource
      resource_for("#{self.name.deconstantize}::#{model.type}").new(model, context)
    else
      new(model, context)
    end
  end
```

The rest of the index operation now applies the resource based on each model's STI type, and the serializer seems to handle everything correctly. I haven't tested this with any other operations - the main use case for me is just getting a list of recent models with one query.

Is there any interest in supporting this kind of STI response in the lib? I'd investigate this a bit further and start working on a PR if so.

barelyknown:
Yes, I'd like to see it investigated. I've hacked similar solutions, and I think that making support for STI first class would be great.

lgebhardt:
:+1: on a support for STI.

utilityboy:
I'd love to see support for STI as well. Perhaps I'll close #197 in favour of this.

antonversal:
:+1:

PR #393, opened 2015-08-21
support STI

antonversal:
can't wait util #364 is implemented. 

caching of `module_path` is removed because it works incorrectly for inherited resources.

```
(byebug) self
Api::V7::PurchaseOrderResource
(byebug) self.module_path
"api/v6/" #should be api/v7/
```


lgebhardt:
Merged by way of #507

lgebhardt:
@antonversal Thanks!

PR #394, opened 2015-08-23
Relation name

### LINKED ISSUE #386: Creating relationship that differs in name from type
In the json:api spec there is this example, suggesting that is is possible to post a people resource identifier object to the author relationship. I am trying to so something similar and am getting some different behaviour than I would expect from JR.

```
PATCH /articles/1/relationships/author HTTP/1.1
Accept: application/vnd.api+json

{
  "data": { "type": "people", "id": "12" }
}
```

Using this example, I would expect JR to look up the article and person objects and call something like this `article.author = person`. This assumes that the article model implements an author association that represents an instance of Person.

The example above using the to_one, but assume for example's sake it's a too_many authors, since that's where I've been having the problem.

In `Resource#_create_to_many_links` this code `@model.public_send(relationship.type) << related_resource.model` would call `article.people` instead of `article.author(s)`, since `relationship.type` is derived from the `class_name` option of the association in the resource class.

Not sure if this is a bug or a feature and my problem stems from a misunderstanding of the spec. Should the association be modified so that some attribute of the relationship object other than type can be used to make associations, or is there a way to make this work with the options that are currently implemented?

Just to be clear, what I would like to achieve is for JR to call the name of the relationship in the url as the relationship method on the model, and not the class_name option from the relationship in the JR Resource class. Changing the relationship class name in the resource class to make it work but then have my relationship pointing to a class that doesn't actually exist.

lgebhardt:
Uses the relation_name instead of the type for _create_to_many_links.

May fix #386

### DIFF
--- a/lib/jsonapi/relationship.rb
+++ b/lib/jsonapi/relationship.rb
@@ -24,7 +24,7 @@ module JSONAPI
       @resource_klass ||= Resource.resource_for(@module_path + @class_name)
     end
 
-    def relation_name(options = {})
+    def relation_name(options)
       case @relation_name
         when Symbol
           # :nocov:
--- a/lib/jsonapi/resource.rb
+++ b/lib/jsonapi/resource.rb
@@ -165,10 +165,11 @@ module JSONAPI
       relationship_key_values.each do |relationship_key_value|
         related_resource = relationship.resource_klass.find_by_key(relationship_key_value, context: @context)
 
+        relation_name = relationship.relation_name(context: @context)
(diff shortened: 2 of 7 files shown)


barelyknown:
@lgebhardt Is this good to merge? I wasn't sure given the "may fix" comment and didn't have time to take a very close look.

lgebhardt:
@barelyknown I think these changes are right, but it would probably be good to wait until we know if it helps with #386.

mmartinson:
Fixes #386

barelyknown:
Thanks for verifying.

PR #395, opened 2015-08-24
Use `records_for_relation_type` method in ShowRelatedResourcesOperation

barelyknown:
The change from using `records_for` to `public_send(@relation_type)` caused errors because the `records` method in `ShowRelatedResourcesOperation` returned an `Array` of `RelatedResource` instead of an `ActiveRecord::Relation` of the records. The change also fixes an untested bug related to record counts for related resources.


lgebhardt:
Thanks!

PR #396, opened 2015-08-24
underscore JSONAPI::Resource#module_path when deriving from name

camhine:
Using downcase does not have the intended effect. For example, a 
resource with name 'MyModuleNameSpace::NamespacedResource' previously 
had a module_path of 'mymodulenamespace'. Using underscore the 
module_path for 'MyModuleNameSpace::NamespacedResource' is 
'my_module_name_space'.

### DIFF
--- a/lib/jsonapi/resource.rb
+++ b/lib/jsonapi/resource.rb
@@ -642,7 +642,7 @@ module JSONAPI
       end
 
       def module_path
-        @module_path ||= name =~ /::[^:]+\Z/ ? ($`.freeze.gsub('::', '/') + '/').downcase : ''
+        @module_path ||= name =~ /::[^:]+\Z/ ? ($`.freeze.gsub('::', '/') + '/').underscore : ''
       end
 
       def construct_order_options(sort_params)
--- a/test/unit/resource/resource_test.rb
+++ b/test/unit/resource/resource_test.rb
@@ -46,6 +46,11 @@ class PersonWithCustomRecordsForErrorResource < PersonResource
   end
 end
 
+module MyModule
+  class MyNamespacedResource < JSONAPI::Resource
+  end
+end
+
 class ResourceTest < ActiveSupport::TestCase
   def setup
     @post = Post.first
@@ -57,6 +62,10 @@ class ResourceTest < ActiveSupport::TestCase
 
   def test_model
     assert_equal(PostResource._model_class, Post)
+  end
+
+  def test_module_path
+    assert_equal(MyModule::MyNamespacedResource.module_path, 'my_module/')
   end
 
   def test_base_resource_abstract


barelyknown:
Thanks! Can you add a test that would fail without this change?

camhine:
I've added a test that fails without this change. I haven't used ActiveSupport::TestCase before so please let me know if I haven't done it right.

barelyknown:
Looks good! Thanks for adding the test.

barelyknown:
@lgebhardt How do you feel about this PR? This could break existing routes, and we're not providing an easy way to avoid that. I agree with @camhine that underscoring the module name seems like the better option, but I wanted to hear your take about the trade off before making the change.

lgebhardt:
I'd like to see a test that demonstrates the need for this PR. I'm not yet seeing where the existing method is breaking things. Of course I'm also not seeing an issue with the changes either. As far as I remember, and can find, the module_path method isn't used for the routes, or other publicly visible strings. Could you provide a few more details so I can better understand the issue.

camhine:
This method's current behaviour breaks when I try to create certain
namespaced resources.

If I have a resource such as MyModule::AResource that 'has_one'
MyModule::SomeOtherResource creating/updating a MyModule::AResource breaks
because it cant find resource 'mymodule/some_other_resource'. The resource
is there, its just in my_module not mymodule.

It's something along those lines anyway. I'll try to get some time this
weekend to post some code that really demonstrates the behaviour.

camhine:
Retrieving namespaced has_one resources is also broken without this fix.

Say we have a pair of namespaced resources:

```
module MyNamespace
  class ParentResource < JSONAPI::Resource
    has_one :child
  end

  class ChildResource < JSONAPI::Resource
    has_one :parent
  end
end
```

Routes like:

```
namespace :my_namespace do
    jsonapi_resources :parents
    jsonapi_resources :childs
  end
```

Then we try to get the child of the parent resource, where parent with id of 1 exists:

```
GET "/my_namespace/parents/1/child"
```

The result is this:

```
Internal Server Error: JSONAPI: Could not find resource 'mynamespace/Child'. (Class Mymodule::ChildResource not found) /usr/local/lib/ruby/gems/2.2.0/gems/jsonapi-resources-0.5.7/lib/jsonapi/resource.rb:277:in `resource_for'
```

I tracked the source this error down to the JSONAPI::Resource.module_path method. It downcases the resources name rather than underscoring it. Leading to the 500 error above.

I figured downcasing is not how this method is meant to behave. Changing it to underscore fixed the 500 on my end and didn't cause any other issues. I added a unit test which fails for the downcase case, and passes in the underscore case.

I guess adding an integration test for this would further demonstrate the issue I just described. I actually spent an hour or so trying to figure out a way to do it in the existing test suite. But it seems like it would take more effort than is worthwhile.

This is really a unit/class level issue. So far I think this method has been working by coincidence - namespaces without camel casing (i.e Api, or Namesapce) work using either downcase or underscore.

Please let me know if I'm missing something. I can write an integration test if somebody convinces me of it's necessity.

barelyknown:
Thanks for the detail. Sounds like we should merge to me. Do you agree @lgebhardt?

lgebhardt:
Agreed. Thanks!

camhine:
Thanks heaps @lgebhardt @barelyknown. We're getting real value out of this gem :)

PR #403, opened 2015-08-27
Refactor request

mmartinson:
@lgebhardt this introduces change to make a method easier to override and does a small refactor to make Request#parse_params easier to read. See commit message for some more details

### DIFF
--- a/lib/jsonapi/request.rb
+++ b/lib/jsonapi/request.rb
@@ -414,7 +414,6 @@ module JSONAPI
                 checked_to_many_relationships[param] = result_val
               end
             end
-
           end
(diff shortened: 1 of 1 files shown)


lgebhardt:
Thanks! That's much cleaner.

mmartinson:
I've been using this library a lot for a new project and am starting to get
to know the codebase pretty decently. I'd like to contribute some more and
would be happy to hear some feedback or direction as far as things that are
working or not or on parts of the codebase that need attention. I don't
have much experience with open source projects so let me know if it would
be more or less helpful to do things in certain ways.

PR #410, opened 2015-08-31
Add some additional info about raising errors

### LINKED ISSUE #408: Update docs for raising appropriate error codes
Apparently it's possible to get `jsonapi-resources` to raise appropriate error codes for different situations (see #407), but this is not documented well. We should show an example that does not rely on some other library and we should show enough code to help people understand exactly what's happening top to bottom.

jagthedrummer:
This addresses the subject of raising errors with all the relevant
info that a user would need to know how to handle their own application
level errors.

I placed it right after the first mention of raising custom errors, but
I'm happy to move it elsewhere if that would be better.

Resolves #408

### DIFF
--- a/README.md
+++ b/README.md
@@ -427,7 +427,8 @@ end
 
 ```
 
-For example, you may want raise an error if the user is not authorized to view the related records.
+For example, you may want raise an error if the user is not authorized to view the related records. See the next
+section for addtional details on raising errors.
 
 ```ruby
 class BaseResource < JSONAPI::Resource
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -427,7 +427,8 @@ end
 
 ```
 
-For example, you may want raise an error if the user is not authorized to view the related records.
+For example, you may want raise an error if the user is not authorized to view the related records. See the next
+section for addtional details on raising errors.
```
dgeb: s/addtional/additional/

README.md
```diff
@@ -444,6 +445,42 @@ class BaseResource < JSONAPI::Resource
 end
 ```
 
+
+###### Raising Errors
+
+Inside the finder methods (like `records_for`) or inside of resource callbacks
+(like `before_save`) you can `raise` an error to halt processing. JSONAPI::Resources
+has some built in errors that will return appropriate error codes. By
+default an other error that you raise will return a `500` status code
```
dgeb: s/an/any/

dgeb:
Just a couple typos, but otherwise LGTM. Thanks @jagthedrummer!

@barelyknown @lgebhardt please review.

README.md
```diff
@@ -444,6 +445,42 @@ class BaseResource < JSONAPI::Resource
...
+
+```ruby
+# config/initializer/jsonapi-resources.rb
+JSONAPI.configure do |config|
+  config.exception_class_whitelist = [NotAuthorizedError]
+end
+```
+
+Handling the error and rendering the appropriate code is now the resonsiblity of the
+appliation and could be handled like this:
```
lgebhardt, line 472: spelling: `appliation` --> `application`

lgebhardt:
Other than the spelling typo it looks good to me.

dgeb:
Just fixed that last typo and did a little other README cleanup too. :shipit:

jagthedrummer:
Hooray! :smiley:

PR #414, opened 2015-08-31
Status codes Take 2

### LINKED ISSUE #412: error objects status value
I was looking at the JSON-API documentation and I think the error object that JSONAPI-Resources is generating are incorrect. Looking at the docs it says the 
- status:  the HTTP status code applicable to this problem, expressed as a string value
- code: an application-specific error code, expressed as a string value.

On a simple little app I created to tryout jsonapi-resource the following error object is generate when attempting to get a post that doesn't exist.

```
$ curl -X GET -H "Accept: application/vnd.api+json" http://test-json-api.dev/api/posts/7

$ {"errors":[{"title":"Record not found","detail":"The record identified by 7 could not be found.","id":null,"href":null,"code":404,"source":null,"links":null,"status":"not_found"}]}[amkirwan@boba-fett:~/code/rails/test-json-api]
```

From reading the JSON-API docs it seems as though `status` should be the string `404` not `not_found`. Therefore the `code` attribute should be some error code specific to the given API.

Maybe I'm totally wrong reading the JSON-API docs but that is how it appears to me the error object should be. Therefore the previous GET request would return the following. The "1100" in the response would be the app specific identifier code for the error.

```
$ {"errors":[{"title":"Record not found","detail":"The record identified by 7 could not be found.","id":null,"href":null,"code":"1100","source":null,"links":null,"status":"404"}]}[amkirwan@boba-fett:~/code/rails/test-json-api]
```

lgebhardt:
Moving back to symbols internally and converting them to the stringified numeric HTTP status code.

Fixes #412 and #413

### DIFF
--- a/lib/jsonapi/error.rb
+++ b/lib/jsonapi/error.rb
@@ -14,7 +14,8 @@ module JSONAPI
                         end
       @source         = options[:source]
       @links          = options[:links]
-      @status         = options[:status]
+
+      @status         = Rack::Utils::SYMBOL_TO_STATUS_CODE[options[:status]].to_s
     end
   end
 
--- a/lib/jsonapi/exceptions.rb
+++ b/lib/jsonapi/exceptions.rb
@@ -11,7 +11,7 @@ module JSONAPI
 
       def errors
         [JSONAPI::Error.new(code: JSONAPI::INTERNAL_SERVER_ERROR,
-                            status: 500,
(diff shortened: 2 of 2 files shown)


dgeb:
We have a winner :100:

amkirwan:
Looks good. Makes sense to use symbols and convert them.

PR #415, opened 2015-09-01
Exception callback

mmartinson:
Refactor of operations processors to reduce some duplication, and adding of feature to the ResourceController for passing callbacks to the general error catcher. 

I needed a way to have control of logging and place to put airbrake handling for exceptions that are being caught and rendered. The controller seemed like a good candidate, since the same handling will be used across resources and error logic seems more the business of a controller than a resource to me. 

I did manual testing of the feature in my own app. I had a look at the test suite and wasn't sure if and where it would make sense to add tests. It seems that the controller tests are all integration tests. I'm happy to add some unit tests if that's deemed important, but could use some direction if that's the case. I'm also happy to hear general feedback if there's something that doesn't seem right.

### DIFF
--- a/lib/jsonapi/acts_as_resource_controller.rb
+++ b/lib/jsonapi/acts_as_resource_controller.rb
@@ -182,5 +182,37 @@ module JSONAPI
         # :nocov:
       end
     end
+
+    def add_error_callbacks(callbacks)
+      @request.server_error_callbacks = callbacks || []
+    end
+
+    # Pass in a methods or a block to be run when an exception is 
+    # caught that is not a JSONAPI::Exceptions::Error
+    # Useful for additional logging or notification configuration that 
+    # would normally depend on rails catching and rendering an exception.
+    # Ignores whitelist exceptions from config
+
+    class_methods do 
+      def on_server_error(*args, &callback_block)
+        callbacks = []
+
+        if callback_block 
+          callbacks << callback_block
+        end
+
+        method_callbacks = args.map do |method|
+          ->(error) do 
+            if self.respond_to? method
+              send(method, error)
+            else
+              Rails.log.warn("#{method} not defined on #{self}, skipping error callback")
+            end
+          end
+        end.compact
+        callbacks += method_callbacks
+        append_before_filter { add_error_callbacks(callbacks) }
+      end
+    end
   end
 end
--- a/lib/jsonapi/acts_as_resource_controller.rb
+++ b/lib/jsonapi/acts_as_resource_controller.rb
@@ -206,7 +206,7 @@ module JSONAPI
             if self.respond_to? method
               send(method, error)
             else
-              Rails.log.warn("#{method} not defined on #{self}, skipping error callback")
+              Rails.logger.warn("#{method} not defined on #{self}, skipping error callback")
             end
           end
         end.compact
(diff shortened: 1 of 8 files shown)


barelyknown:
I had this problem on the to-do list for an application that I'm working on - thanks for getting to it.

I just tested it with my app, and things look pretty good. A few comments:
- `Rails.log` is being called instead of `Rails.logger` in one place. That needs to be fixed.
- It wasn't clear to me why the callback should be a class method instead of an instance method. It's not a big deal, but why did you go that way?

To test the feature in my application, I did something like the following:

``` ruby
it "calls report_to_appsignal when an exception occurs" do
  begin
    JSONAPI::CreateResourceOperation.class_eval do
      alias_method :original_apply, :apply
      def temporary_apply
        raise StandardError, "just a temporary error"
      end
      alias_method :apply, :temporary_apply
    end
    expect_any_instance_of(JSONAPI::CreateResourceOperation).to receive(:apply).and_call_original
    expect(ApplicationController).to receive(:report_to_appsignal)
    post :create, data: data
  ensure
    JSONAPI::CreateResourceOperation.class_eval do
      alias_method :apply, :original_apply
      remove_method :temporary_apply
    end
  end
end
```

A similar test seems like a good idea for this PR. Let me know if you need any help figuring that out.

lib/jsonapi/acts_as_resource_controller.rb
```diff
@@ -182,5 +182,37 @@ def handle_exceptions(e)
...
+        if callback_block 
+          callbacks << callback_block
+        end
+
+        method_callbacks = args.map do |method|
+          ->(error) do 
+            if self.respond_to? method
+              send(method, error)
+            else
+              Rails.log.warn("#{method} not defined on #{self}, skipping error callback")
```
barelyknown: This is the incorrect call to `Rails.log`.

mmartinson:
@barelyknown let me know if those tests make sense. Your RSpec example makes a lot of sense to me, but i'm unfamiliar with minitest/fixtures and their best practices. Fixed the logging error. 

Re: controller class methods. I recognize the utility of using instance methods for both access to the current request and familiarity to Rails users, but it was unclear to me how to implement that. I wasn't sure how the get a reference of the controller instance to pass into the block that calls the method. Does that OperationsProcessor that triggers the callback have access to its controller instance through `@request`?

barelyknown:
In my apps I usually provide the instance of the controller in the `context` like this:

``` ruby
def context
  {
    current_user: current_user,
    controller: self,
  }
end
```

But, since we can't count on having access to the instance of the controller, I'm fine with the class method approach.

The tests look good to me.

@lgebhardt What do you think? I like this feature addition quite a bit - it's important for production apps.

lgebhardt:
Looks pretty good to me. We do need to get some tests into there to get the coverage back to 100% when we test it:

``` bash
COVERAGE=true rake test
```

```
5847 / 5855 LOC (99.86%) covered
```

The existing tests are mostly integration as you have mentioned. But adding some simple unit tests is just fine (and preferable in most cases). If some lines simply can't be tested it's OK to exclude them from the coverage calculation, but that should be rare.

mmartinson:
tested

lgebhardt:
@mmartinson Thanks!

PR #416, opened 2015-09-01
Use the ':git' source when bundling Rails master

jashank:
As per http://bundler.io/git.html:

> `http://` and `git://` URLs are insecure. A man-in-the-middle attacker
> could tamper with the code as you check it out, and potentially supply
> you with malicious code instead of the code you meant to check
> out. Because the `:github` shortcut uses a `git://` URL in Bundler 1.x
> versions, we recommend using using HTTPS URLs or overriding the
> `:github` shortcut with your own HTTPS git source.

Also, as Rails master (currently 5.0.0.alpha) has a dependency on Arel
master (currently 7.0.0.alpha), we must pull this in too.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,12 +11,13 @@ platforms :jruby do
 end
 
 version = ENV['RAILS_VERSION'] || 'default'
-rails = case version
-        when 'master'
-          { github: 'rails/rails' }
-        when 'default'
-          '>= 4.2'
-        else
-          "~> #{version}"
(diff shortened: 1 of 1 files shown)


dgeb:
I would hope no one's pulling from master for anything but development (i.e. probably not a realistic concern for production apps). But regardless, seems like a good precaution. :+1:

lgebhardt:
Thanks!

PR #420, opened 2015-09-02
Key type config option

lgebhardt:
Rebased #160 with some slight rework. 
Added constraint for routes if the key_type is `:uuid`.


dgeb:
Looks great @lgebhardt :shipit:

karlfreeman:
Lovely stuff here. Thanks so much @lgebhardt!

PR #421, opened 2015-09-02
Exceptions for invalid has_many post and patch

wernull:
**Issue:** Post/patch of a has_many would throw 500 instead of 404 like the invalid post/patch of a has_one.

Example exception response with one valid and one invalid has_many:

```
{
    "errors": [{
        "title": "Record not found",
        "detail": "Couldn't find all Users with 'id': (1, 9999) (found 1 results, but was looking for 2)",
        "id": null,
        "href": null,
        "code": 404,
        "source": null,
        "links": null,
        "status": "not_found"
    }]
}
```

or with only one invalid has_many:

```
{
    "errors": [{
        "title": "Record not found",
        "detail": "Couldn't find User with 'id'=9999",
        "id": null,
        "href": null,
        "code": 404,
        "source": null,
        "links": null,
        "status": "not_found"
    }]
}

```

### DIFF
--- a/lib/jsonapi/exceptions.rb
+++ b/lib/jsonapi/exceptions.rb
@@ -45,6 +45,20 @@ module JSONAPI
       end
     end
 
+    class HasManyRecordNotFound < Error
+      attr_accessor :message
+      def initialize(message)
+        @message = message
+      end
+
+      def errors
+        [JSONAPI::Error.new(code: JSONAPI::RECORD_NOT_FOUND,
+                            status: :not_found,
+                            title: 'Record not found',
+                            detail: "#{message}")]
+      end
+    end
+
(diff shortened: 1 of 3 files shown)


lgebhardt:
By catching the ActiveRecord::RecordNotFound exceptions in the Operation it's tying them to ActiveRecord. We've been trying to keep the Operation independent of ActiveRecord by catching ActiveRecord exceptions in ActiveRecordOperationsProcessor. This allows the project to be more easily used with AR models, as well as plain ruby classes.

I'm not sure why the ActiveRecord::RecordNotFound exception isn't being caught by the ActiveRecordOperationsProcessor in your case, which I think should result in a 404.

Issue #422, opened 2015-09-03
Removing relationships via relationship URL


timabdulla:
Currently, it appears that JR does not comply with the spec with regard to how it handles the removing of relationships via a relationship URL. The spec dictates that relationships are to be removed by sending a request with the DELETE method to the relationship URL with a request body that specifies the resources for which the relationships should be removed.[0] JR allows for this operation by accepting a comma-separated list of IDs of the resources for which the relationships should be removed at the end of the relationship URL. It is not only incongruent with the spec, but it also forecloses the possibility of allowing for a heterogeneous collection of relationships to exist at the URL, as it assumes that primary key alone is enough to uniquely identify a resource, when only the tuple of type and ID are truly enough to uniquely identify a resource.

Was there a reason for this decision, or is this simply an oversight? It looks fairly easy to correct if it's the latter, and I'd be happy to step in and help!

[0] = http://jsonapi.org/format/#crud-updating-to-many-relationships

lgebhardt:
That's an oversight. It looks like I missed that as the spec changed. I would greatly appreciate the help in fixing this. Thanks!

timabdulla:
we @paysavvy are pretty heavy users of JR, so always happy to help out! we'll get on this right away. thanks for this great project!

timabdulla:
cc @mmartinson

vrybas:
@timabdulla fixed by #423?

Issue #404, opened 2015-08-27
Related source with `relation_name` throws undefined method error on create / update actions


apraditya:
I have this resource

```
class PostResource < JSONAPI::Resource
  has_many  :categories,   relation_name: 'internal_categories'
end
```

and I get this error on POST create and PUT update calls

```
Internal Server Error: undefined method `category_ids=' for class `Post' /usr/local/opt/rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/jsonapi-resources-0.5.7/lib/jsonapi/resource.rb:705:in `method'
```

Shouldn't it run based on `relation_name` options? Probably similar to #378 (in a way). Thanks

jcarlson:
Yeah I just noticed this too and left a comment in a commit somewhere: https://github.com/cerebris/jsonapi-resources/commit/494e869ab3739ddb149b3dd6ed897e280a0ee960#commitcomment-13165776

This may be related to #378 but is definitely a different issue.

lgebhardt:
I'll see if I can get a fix for this soon.

lgebhardt:
#447 fixes `_remove_to_many_link` to use the relation_name. This won't fix the error on the POST and PUT calls. For that I think setting the `foreign_key` option on the relationship is going to be required. Can you check if this solves the problem?

PR #423, opened 2015-09-03
Fix delete to many

mmartinson:
Fixes what I think is a bug in deleting a to_many relationship. I issue is similar to what was addressed in #394 when creating a to_many. I had a look at the other relationship creation/destruction code in Resource and I'm wondering there maybe a similar problem in other methods. I fixed this one but don't have time at the moment to write tests for the others. 

Here's the gist of what I was getting:

My client is calling `DELETE /teams/1/relationships/team_leads/7`

Team has a relationship to both `employees` and `team_leads`, which can be the same record.

Before the change, JR was deleting the relationship to the employee with this id while not modifying the team lead relationship. It was doing so because `team_lead` has (I believe correctly) the type `employees`. It seems to me that JR should be using the value from the url here and not the type of underlying model?

### DIFF
--- a/lib/jsonapi/resource.rb
+++ b/lib/jsonapi/resource.rb
@@ -167,7 +167,6 @@ module JSONAPI
 
         relation_name = relationship.relation_name(context: @context)
         # TODO: Add option to skip relations that already exist instead of returning an error?
-        # binding.pry if relation_name == :special_tags
         relation = @model.public_send(relation_name).where(relationship.primary_key => relationship_key_value).first
         if relation.nil?
           @model.public_send(relation_name) << related_resource.model
--- a/lib/jsonapi/resource.rb
+++ b/lib/jsonapi/resource.rb
@@ -167,6 +167,7 @@ module JSONAPI
 
         relation_name = relationship.relation_name(context: @context)
         # TODO: Add option to skip relations that already exist instead of returning an error?
+        # binding.pry if relation_name == :special_tags
         relation = @model.public_send(relation_name).where(relationship.primary_key => relationship_key_value).first
         if relation.nil?
           @model.public_send(relation_name) << related_resource.model
@@ -180,7 +181,6 @@ module JSONAPI
 
     def _replace_to_many_links(relationship_type, relationship_key_values)
       relationship = self.class._relationships[relationship_type]
-
       send("#{relationship.foreign_key}=", relationship_key_values)
       @save_needed = true
 
@@ -208,9 +208,7 @@ module JSONAPI
     end
 
     def _remove_to_many_link(relationship_type, key)
-      relationship = self.class._relationships[relationship_type]
-
-      @model.public_send(relationship.type).delete(key)
+      @model.public_send(relationship_type).delete(key)
 
       :completed
     end
--- a/test/controllers/controller_test.rb
+++ b/test/controllers/controller_test.rb
@@ -1208,7 +1208,7 @@ class PostsControllerTest < ActionController::TestCase
     p = Post.find(14)
     assert_equal [2, 3], p.tag_ids
 
-    delete :destroy_relationship, {post_id: 14, relationship: 'tags', keys: '3'}
+    delete :destroy_relationship, {post_id: 14, relationship: 'tags', data: [{type: 'tags', id: 3}]}
 
     p.reload
     assert_response :no_content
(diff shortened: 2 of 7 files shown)


lgebhardt:
We need to coordinate this with #422

mmartinson:
Ya, definitely. @timabdulla will tackle this together. We're working on the same project with JR. Will either add to this PR or remove it once there is another up.

lib/jsonapi/resource.rb
```diff
@@ -208,9 +207,7 @@ def _replace_polymorphic_to_one_link(relationship_type, key_value, key_type)
     end
 
     def _remove_to_many_link(relationship_type, key)
-      relationship = self.class._relationships[relationship_type]
-
-      @model.public_send(relationship.type).delete(key)
+      @model.public_send(relationship_type).delete(key)
```
mmartinson, line 210: @timabdulla. The problem I identified was with the Resource using the type key of the relationship being changed, rather than the relationship from the path. This change addresses it, but may be unnecessary if code will be changed to extract the type from the resource identifier object. 
I'm not sure how a relationship being polymorphic affects this. The to-many code uses relation_name on the model (which is configurable as a option in the resource relationship declarations). It seems for that reason that the relationship to be deleted should still be the polymorphic type given in the path.

test/controllers/controller_test.rb
```diff
@@ -1215,6 +1215,24 @@ def test_delete_relationship_to_many
...
+
+    delete :destroy_relationship, {post_id: 14, relationship: 'special_tags', keys: '2'}
+    assert_equal 0, Post.find(14).special_tags.count, "Relationship that matches URL relationship not destroyed"
+
+    #check that the tag association is not affected
+    assert_equal Post.find(14).tags.count, before_tags
+  ensure
+    PostResource.instance_variable_get(:@_relationships).delete(:special_tags)
+  end
+
```
mmartinson, line 1235: @timabdulla this test could be reworked slightly for the changes you're making.

timabdulla:
added commits to fix #422 

edit: also did some refactoring to reduce duplication in parsing of relationship actions

timabdulla:
@mmartinson it looks like my commits are causing PostsControllerTest#test_delete_relationship_to_many_with_relationship_url_not_matching_type to fail. looking into it now.

test/fixtures/active_record.rb
```diff
@@ -1014,6 +1033,7 @@ module V2
 
     class BookResource < JSONAPI::Resource
       attribute :title
+      attribute :title
```
lgebhardt: Why is this duplicated?

lib/jsonapi/resource.rb
lgebhardt, line 210: Yes, this should use the `relation_name` from the relationship. This was brought up in https://github.com/cerebris/jsonapi-resources/commit/494e869ab3739ddb149b3dd6ed897e280a0ee960#commitcomment-13165776 and #404

test/fixtures/active_record.rb
mmartinson: not for any reason. fixed

lgebhardt:
Thanks! I'm going to work on fixing #404, which will also fix the one note above regarding relation_name, in a separate PR.

jcarlson:
An unfortunate downside to this change is that it makes using Angular as a client near impossible, as the Angular team has apparently decided that `DELETE` requests shall not have bodies. So I'm a little stuck on how I'm going to upgrade from 0.5.9 to 0.6.0 without completely breaking my JavaScript client's ability to delete relationships.

I get it, the spec says "do it this way", but here's a case where strict adherence to an arbitrary spec is causing a negative impact on my product/team/business.

Would it be possible to allow some kind of configuration or backwards compatibility mode to make working with Angular clients at least possible?

FWIW, here's the problem code in Angular's ng-resource library: 
https://github.com/angular/bower-angular-resource/blob/master/angular-resource.js#L531

jcarlson:
Also, I would be happy to lobby the Angular team to make changes, but a little Googling suggests that others have tried and lost that war.

timabdulla:
that is unfortunate, but there were good reasons other than strict adherence to implement this. please read my motivating description at the top of #422 for my argument for the change. we use Angular as well, but we have our own in-house API client that is specifically built to work with JSON API.

I think that having a config option to support deviations from the spec only encourages partial adherence to the spec, which in turn encourages a multiplicity of deviant implementations that somewhat undo the point of having a spec in the first place.

my personal recommendation, and I don't speak for this project, would be to just monkey-patch this gem to revert the change. it will make upgrades a bit more painful, but I don't think introducing extra complexity to support your specific edge case is worth it. here is the commit that you would want to work around: https://github.com/cerebris/jsonapi-resources/commit/19249d137c4b389a3b38309994a7548c3ed7d394#diff-a32d87f955d51d1e7270b1a974529f7e

PR #426, opened 2015-09-07
Use `module ClassMethods` instead of `class_methods do` to support Rails 4.0

davidstosik:
This issue was introduced with 01814602920ede0620e4420cb5c553d503933186, released with 0.5.9.


lgebhardt:
Thanks!

Issue #129, opened 2015-03-22
Application structure question


elucid:
This is a very interesting project and executes well something that many other projects have attempted in the past.

I have seen the sample contacts application, read the documentation, and I am currently reading through the gem source but there are still some things which are unclear to me. The most significant of which is where non-CRUD behaviour should be included.

For example, if I were building an application to sell T-shirts using Stripe, I might have an `Order` class with various `LineItem`s that would consist of `Product`s with corresponding quantities. Now, I could easily manage to create all of the relevant records via a create request to the `OrdersController`. However, where would I put the call to create my Stripe charge, schedule fulfilment, etc.?

Normally, I would leave my `Order` and corresponding models fairly simple, and organize charging and fulfilment via services. I would have, say, a `PurchaseService` which would be instantiated by a `PurchaseController`. This would arrange for the various models to be created, trigger the Stripe charge, and if successful, schedule fulfilment (possibly by delegating to another service).

The code in the `PurchaseController` would be very minimal and would not do anything more than pass parameters to the `PurchaseService`, make a single call on it, and then take care of response and error handling.

It seems like the intention with JR would be to leave the default implementation for the resource controllers, and add charging and purchasing behaviour to the resource code. For example, I would do:

``` ruby
class PurchaseController < JSONAPI::ResourceController
end
```

and

``` ruby
class PurchaseResource < JSONAPI::Resource
  def create
    # call to Stripe, etc. goes here
  end
end
```

Does this make sense? Would you structure things differently?

One the one hand this seems like a good place to include the behaviour without needing to reproduce a lot of error and response handling code if I were to override the corresponding controller method. On the other, I've seen the callback support in `JSONAPI::Resource` but I'm not clear if/when the corresponding lifecycle methods are actually called within the resource code. Additionally, if the code were structured this way, the `PurchaseResource` would also have to coordinate validating and creating `Order`s, `LineItem`s, etc. rather than relying on the default `JSONAPI::Resource` functionality to do so.

Anyhow, apologies for the long-winded and naive question but I feel like I am missing something obvious. Any help in clarifying would be appreciated.

kellysutton:
@elucid 

In the past, I've found anything that departs from CRUD should be looked at a little bit more carefully, especially in the case of purchasing. By creating a few extra resources, you can maintain a consistent pattern within your application.

[...]

PR #427, opened 2015-09-07
Use @model.id.nil? to determine resource#is_new? instead of resource#id.nil?

debreczeni:
Fixes create callback and resource saving when id is overridden to something else, like an UUID.
This happens when I want to keep integer id’s internally for the database relations but use an UUID on the public JSON interface.
Since UUID creation happens on resource initialization create callbacks never run.

Example resource:

```
class Api::V1::MyResource < JSONAPI::Resource
  def id
    @model.uuid
  end
end
```

### DIFF
--- a/lib/jsonapi/resource.rb
+++ b/lib/jsonapi/resource.rb
@@ -32,7 +32,7 @@ module JSONAPI
     end
 
     def is_new?
-      id.nil?
+      @model.new_record?
     end
 
     def change(callback)
--- a/lib/jsonapi/resource.rb
+++ b/lib/jsonapi/resource.rb
@@ -32,7 +32,7 @@ module JSONAPI
     end
 
     def is_new?
-      @model.new_record?
+      model.id.nil?
     end
 
     def change(callback)


debreczeni:
Using a separate ID and UUID in the database also needs some tweaking, like overriding:

```
class Api::V1::MyResource < JSONAPI::Resource
  def self._primary_key
    :uuid
  end
end
```

On a sidenote it would be cool if one could define setters not just getters for the resource, and be able to define different required attributes for show and create. So if I want to be able to create an :email attribute but not include :email for the show operation, I'd have to return nils, which can be confusing:

```
class Api::V1::MyResource < JSONAPI::Resource
  attribute :email # Allows write

  def email; end # disables read
end
```

Should I open a new issue with these? I am also having trouble running the specs, they fail with all sorts error before running the suite... Shouldn't a `bundle install && bundle exec rake` be enough? UPDATE: it looks like the test suite fails very early (in the Rakefile) on travis too: https://travis-ci.org/cerebris/jsonapi-resources/jobs/79101541#L240

lgebhardt:
This PR fails three tests on my local build, all related to using `@model.new_record?` when the underlying model isn't based on ActiveRecord.

The early failure you are seeing is apparently due to an update to minitest-reporters 1.1.0 (#428). While that is being fixed you could test with the old 1.0.11 version.

lgebhardt:
#428 is now fixed (by removing the gem, since it was no longer used).

lgebhardt:
Back to your PR:

> This happens when I want to keep integer id’s internally for the database relations but use an UUID on the public JSON interface.
> Since UUID creation happens on resource initialization create callbacks never run.

This project originally tried to allow for this sort of alternate id behind the scenes, however we ran into quite a few issues and support was dropped. If I remember correctly the problems were with the relationships.

debreczeni:
Thanks for your prompt reply! I will do a rebase and check the tests.

Re: having both IDs and UUIDs, I guess I will be good until I add on some relationships, by then I will eighter move to a 128bit integer in the DB for the primary ID or we see if it's worth adding support for this to this gem.

debreczeni:
I have pushed a new version which checks on the model's id attribute, this should make it work in all scenarios.

lib/jsonapi/resource.rb
```diff
@@ -32,7 +32,7 @@ def id
     end
 
     def is_new?
-      id.nil?
+      model.id.nil?
```
lgebhardt, line 35: By directly checking `model.id` you are losing the ability to have an alternate primary key on the model as provided by the `id` function:

``` ruby
def id
  model.public_send(self.class._primary_key)
end
```

debreczeni, line 35: Thanks for your feedback. 

I think currently there is no clear separation of the public interface (resource) and the underlying data (records).
Which would be essential if different types of data backends are to be supported, and a non-1:1:1 mapping between the client, interface and backend. (As mentioned in #129)
I think we are forcing too much opinion if we are to be expecting the id to be nil for new records and not nil if the record is persisted. In my case I am initializing an uuid on record instantiation (what about UUIDs generated on the client side?), and hence the create callbacks never run on my new records, because uuid is already present for new records (whereas id is not).

Let me try to explain. On the resource I am supposed to specify the interface. If I override the id attribute on the resource with an uuid with:

``` ruby
class MyResource
  def id
    @model.uuid
  end
end
```

then I also have to define the following lines if I want the finder to be able to find records:

``` ruby
# ...
  def self._primary_key
    :uuid
  end
# 
```

Which kind of makes the first overriding unnecessary, knowing that:

``` ruby
def id
  model.public_send(self.class._primary_key)
end
```

But isn't JSONAPI::Resource._primary_key supposed to be a "private"ish method?

I understand that I am forcing a feature that is not supported by this gem (to be able to use an internal id and an external one (uuid), and in my special case the the uuid field on my records are initilized on instantiation. So in my particular case it is rather preventing me from using a different field for id then helping it - unless I do some further tweaks:

As a solution to my current problem, swapping from 32bit integer ids to 128 bit could work too (maybe with the help of the activeuuid gem: https://github.com/jashmenn/activeuuid), or simply just overriding the is_new method. 
That's how I am working around the issue at the moment, knowing I use activerecord as my persistence layer:

``` ruby
  def is_new?
    @model.new_record?
  end
```

A possible long term solution would be to make it clear what types of persistence layers are supported, and determine whether a record is new by checking which interface to use based on the persistence layer of the resource.

Or a completely different solution to my current problem would be to always run create callbacks on create actions, without checking the nilness of the id field… But I guess that is not possible since the record is not aware of the current action?

lgebhardt, line 35: Ideally I'd like to add the ability for resources to have both a private (primary) and public key, as you are trying to do here. It's been a common request, and a feature we initially tried to add. I think we are going to need to add an option on the resource to identify the public_key. Without starting the project I'm not sure what issues we'll encounter.

I added issue #434 as a place to discuss this.

debreczeni, line 35: Thank you, much appreciated! :smiley:

scottgonzalez:
Based on the fact that support for other types of models has progressed and `is_new?` is a public method that can be safely overridden in the application, I'd recommend closing this and just overriding `is_new?` in your resource(s).

PR #433, opened 2015-09-08
Allow fields param for resources without explicit relationship to primary resource.

ggordon:
The fields parameter checked that the type matched either the resource
or one of the relationships on the resource. This prevented allowing
a resource when its name didn't match the relationship and relationships
with more than one level of nesting.


ggordon:
Closed since included in #435

PR #435, opened 2015-09-09
Allow fields param when name does not match resource

ggordon:
The fields parameter checked that the type matched either the resource
or one of the relationships on the resource. This prevented allowing
a resource when its name didn't match the relationship and relationships
with more than one level of nesting.

Also when checking for the fields param for a resource, if not
specified, recursively check for field of parent classes until
JSONAPI::Resource is reached.


lgebhardt:
Thanks!

ggordon:
Thanks for merging! @lgebhardt

Issue #424, opened 2015-09-04
Nested attributes for relationships


garytaylor:
If I were updating / creating a "post" and I also wanted to create / update tag id 3 in the example below, should that be possible (i.e. similar to nested attributes in active record).

Probably a bit of a bad example, but hopefully you get the idea ?

I am finding that each entry in "data" in the relationships must only contain id and type.  Nowhere have I yet seen how to do this in json_api in general, not just in this gem so maybe it is not possible.  In which case, I guess I would need to create / update the tag seperately, but then I may need to worry about rolling it back if the creation / update of the post did not succeed etc...  

```
post '/posts',
  {
    'posts' => {
      'attributes' => {
        'title' => 'A great new Post'
      },
      'relationships' => {
        'tags' => {
          'data' => [
              {type: 'tags', id: 3, attributes: {name: "New Tag Name"}},
              {type: 'tags', id: 4}
            ]
        }
      }
    }
  }
```

lgebhardt:
No, that's not possible in the current version of JSON API. There's a Bulk Extension that is currently being worked on. Once it reaches a beta state (expected at the end of September) we will work to get it implemented in JR. See https://github.com/cerebris/jsonapi-resources/issues/183#issuecomment-132786791 for a little more detail.

Also, I'm pretty sure the structure of the payload will be different from what you have shown in your example.

KronicDeth:
The nested format @garytaylor posted is supported by Ember [according](https://github.com/json-api/json-api/issues/795#issuecomment-159927027) to @lolmaus.  @andruby came to that format on his [own](https://github.com/json-api/json-api/issues/795#issuecomment-159926461) as did [I](https://github.com/json-api/json-api/issues/795#issuecomment-168732087), so maybe it will make it into the specs ¯_(ツ)_/¯

bf4:
@lgebhardt @dgeb looks like this behavior has been rejected by the Ember-Data team https://github.com/rails-api/active_model_serializers/pull/1614#issuecomment-202670830

PR #438, opened 2015-09-10
test creating a resource with nested attributes

danielma:
This is an example of using Rails approved nested_attributes to create multiple resources at once. I want this for my application and I know others have asked about it as well (#424 and others).

I realize that this form doesn't necessarily conform to the JSON API spec, so I doubt you'll want to merge it. But I put the test together for my own sake and wanted to make the code available for others.

### DIFF
--- a/Guardfile
+++ b/Guardfile
@@ -1,0 +1,42 @@
+# A sample Guardfile
+# More info at https://github.com/guard/guard#readme
+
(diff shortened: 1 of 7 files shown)


lgebhardt:
Thanks for posting this. As you say it's nonconforming so we won't merge it, but it still may help others. I'll leave it open for a bit until we get a better way of sharing code and tips setup.

PR #443, opened 2015-09-12
Link Builder Not Working when used within a rails engine

garytaylor:
When I used the gem as part of a rails engine, with the models inside the namespace of the engine, the method used to generate the path of the resource has a "_" at the beginning, therefore causing method_missing errors. (.e.g. _user_path instead of user_path)
This was due to the engine_resource_path_name_from_source method calculating the method name by using the base_path_name followed by an underscore.
However, this does not allow for when base_path_name is an empty string.

So, this pull request instead uses an array of the components of the method name, filtering out any blanks and joining them with underscores.

I have not fully checked out the project - this change was done within github, so unfortunately no tests checked.  If you cant accept it without, then I can get on to doing these but not until mid week.

Thanks

Gary

### DIFF
--- a/lib/jsonapi/link_builder.rb
+++ b/lib/jsonapi/link_builder.rb
@@ -81,7 +81,7 @@ module JSONAPI
       scopes         = module_scopes_from_class(source.class)[1..-1]
       base_path_name = scopes.map { |scope| scope.underscore }.join("_")
       end_path_name  = source.class._type.to_s.singularize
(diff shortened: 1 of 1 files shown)


lgebhardt:
Thanks!

garytaylor:
No worries.

Im really impressed with this gem saves lots of time :-)
If you ever want any help with any features I've 9 years ruby experience
and heavily into json api. Just give me a shout.

Cheers

Gary

PR #451, opened 2015-09-21
Remove Rails 4.0.x testing from Travis

lgebhardt:
This removes Travis testing for Rails 4.0.x. This is being done because using the in memory database sometimes crashes ruby. Using the file based database fails under Rails 4.0 with a readonly message.

I will still periodically test under Rails 4.0 using the in memory database, and if someone finds a solution to either of the above issues we can add Travis support back for Rails 4.0. It's just no worth any more of my time trying to track this down.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: ruby
 sudo: false
 env:
-  - "RAILS_VERSION=4.0.0"
   - "RAILS_VERSION=4.1.0"
   - "RAILS_VERSION=4.2.0"
 rvm:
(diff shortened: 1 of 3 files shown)


dgeb:
Fine by me. @barelyknown any objections or ideas to workaround the segfaults we've been seeing testing 4.0 with Travis?

barelyknown:
No, that makes sense to me. I'd like us to focus on 4.2 and 5.0 from here on.

dgeb:
> I'd like us to focus on 4.2 and 5.0 from here on.

Sounds good to me :+1:

PR #457, opened 2015-09-24
enable config.always_include_to_many_linkage_data

danielma:
The README specifically states that

> always_include_to_many_linkage_data is not currently implemented

However, config is already set up to receive that configuration parameter, and adding only one line to the code (as well as enabling it in the config) makes this feature start working. 

Given that this feature is so close to being implemented, but you left it out, I have to assume there is a good reason. However, I couldn't find any reason that it shouldn't be allowed. Can you help me understand why this isn't enabled right now?

This is more of a question than anything. It might have been good as an issue, but a pull request lets me easily attach the code I'm referring to.

Thanks!

### DIFF
--- a/lib/jsonapi/resource_serializer.rb
+++ b/lib/jsonapi/resource_serializer.rb
@@ -256,6 +256,7 @@ module JSONAPI
     end
 
     def link_object_to_many(source, relationship, include_linkage)
+      include_linkage = include_linkage | @always_include_to_many_linkage_data | relationship.always_include_linkage_data
       link_object_hash = {}
       link_object_hash[:links] = {}
       link_object_hash[:links][:self] = self_link(source, relationship)


lgebhardt:
The main point of contention for me is that the option will potentially result in large results if there are many related records. This is currently also an issue if you use the include option, but with that the client at least has an option to turn off the include on the request. 

I'm afraid we would be giving people a foot gun without implementing pagination of the linkage data. That's why I held off on the implementation. My plan has been to enable it once we implement pagination of linkage data.

danielma:
@lgebhardt got it. Thanks for the quick response. It would help me a lot to have this enabled, so I think I'll use a patched version for now.

As always, thanks for this great gem

lgebhardt:
At least the patch is small and unlikely to conflict with other changes. Hopefully I find some time soon to get pagination added for the linkage data.

danielma:
@lgebhardt yesterday I thought I needed this, today I realized I shot myself in the foot. 

I'm still using the patched version because there are _some_ resources where I want to `always_include_linkage_data`, but I've turned `always_include_to_many_linkage_data` off

wvteijlingen:
I needed this functionality and patched it in as follows, just in case anyone needs this :).

```ruby
module JSONAPI
  module AlwaysIncludeToManyLinkageData
    def link_object_to_many(source, relationship, include_linkage)
      include_linkage = include_linkage | @always_include_to_many_linkage_data | relationship.always_include_linkage_data
      super(source, relationship, include_linkage)
    end
  end

  class ResourceSerializer
    prepend AlwaysIncludeToManyLinkageData
  end
end
```

PR #459, opened 2015-09-25
Context cleanup

### LINKED ISSUE #458: Trying to interject authorization, but context is always nil
Following the README, I'm trying to interject an authorization step across all resources. I've got a base resource that looks like this:

```
    module Api
      module V1
          def records_for(relation_name, options={})
            context = options[:context]
            records = model.public_send(relation_name)

            # just logging options hash for the time being
            Rails.logger.debug options

            records
          end
        end
      end
```

All of my other resources inherit from this `BaseResource` class. I'm just passing the records through and logging the `options` hash for now. `options` is always `{:context => nil}`.

I also have a base controller called `API::V1::ApplicationController`, which inherits from `JSONAPI::ResourceController`, and which includes the following method:

```
def context
  {current_user: current_user}
end
```

All of my other API controllers inherit from `API::V1::ApplicationController`. I can confirm that `current_user` is set properly when the `context` method is called, but it never makes its way to the options hash in `BaseResource`.

lgebhardt:
Hopefully this PR adds a little more rigor to the initialization of resources. By requiring the context to be set for initializing a resource we can prevent cases where users inadvertently create resources without the associated context. It would be an easy mistake to make when, for example, the find_by_key method is overridden to forget to pass in the context from the options hash.

This will require a minor version bump because it's potentially a breaking change.

May fix #458

### DIFF
--- a/README.md
+++ b/README.md
@@ -462,12 +462,12 @@ class PostResource < JSONAPI::Resource
 
   # def record_for_author
   #   relation_name = relationship.relation_name(context: @context)
-  #   records_for(relation_name, context: @context)
+  #   records_for(relation_name)
   # end
 
   # def records_for_comments
   #   relation_name = relationship.relation_name(context: @context)
-  #   records_for(relation_name, context: @context)
+  #   records_for(relation_name)
   # end
 end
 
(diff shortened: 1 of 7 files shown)


barelyknown:
This will probably require many users to update their tests, but it's worth it IMO to avoid resources that inadvertently have a context. Good to merge from my perspective.

PR #462, opened 2015-09-25
Explanation of filter chain / context issue

beechnut:
I'm not sure this is the right place for it, but this note in this place might have saved @sirvine and I some time.


lgebhardt:
@beechnut Good idea. Thanks.

PR #465, opened 2015-09-28
Switch from `_filter` to `_action`

bdmac:
Rails 5 includes a deprecation warning about using the `_filter`
versions of controller callbacks. Switch to the newere `_action`
versions of the same callbacks to avoid the deprecation warning. All
versions of Rails being tested include support for this variant on the
callback names.


lgebhardt:
Thanks!

PR #472, opened 2015-10-03
added setter example within getter example in 'Formatting' section of README

Jeehut:
Since I was quite a bit helpless and confused about **if there is or isn't a setter option available too**, as there is a getter option available which can be seen within the README **I decided to provide a setter example** which is included in this pull request.

Since the given getter example **didn't seem to be a good fit for a setter** (or at least I don't know how) I decided to include my real world example (a bit simplified) from [this commit](https://github.com/Flinesoft/MovieAssistor/commit/a83bfa59f9de963a960ba3bcc61a0574da3ba268).

I hope it helps some people in the future recognize the setter option of this great gem more quickly!

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,6 +227,41 @@ end
 The system will lookup a value formatter named `DateWithTimezoneValueFormatter` and will use this when serializing and
 updating the attribute. See the [Value Formatters](#value-formatters) section for more details.
 
+##### Flattening a Rails relationship
+
+It is possible to flatten Rails relationships into attributes by using getters and setters. This can become handy if a relation needs to be created alongside the creation of the main object which can be the case if there is a bi-directional presence validation. For example:
+
+```ruby
+# Given Models
+class Person < ActiveRecord::Base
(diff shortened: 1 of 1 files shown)


lgebhardt:
@Dschee Sorry about the omitted setter example. I feel your setter/getter example may be a bit more involved than is necessary for this small section of the readme, though it looks very useful as an example for flattening a rails relationship into an attribute. What do you think of breaking the spoken language example out either into the wiki or as a subsection of the attributes section? We could then maybe use the email as a simple setter example where the email is just converted to lowercase before storing in the model.

Jeehut:
Yeah, I totally agree with you. I even thought about lowercasing the email but then decided that it wouldn't be a good example as lowercasing the email should be done via a before_validation and not within the resource definition. But maybe that doesn't matter as this is just an example and we can keep it until someone else comes up with something better.

Okay than, let me do the changes, give me a moment ...

lgebhardt:
@Dschee Thanks!

Issue #449, opened 2015-09-16
custom OperationsProcessor not set in config initializer


MetricMike:
![image](https://cloud.githubusercontent.com/assets/311365/9914102/3902b4c6-5c7e-11e5-999b-07c19cdd18b3.png)

![image](https://cloud.githubusercontent.com/assets/311365/9914525/5106d5c8-5c80-11e5-8506-5c05872f7585.png)

Calling `operations_processor=` before `create_operations_processor` in a controller fixes the symptom, but the initializer should be doing that, right?

MetricMike:
The issue I was having was due to load order. Initializers are run before the app/ directory is loaded, so custom classes need to be required into `config/jsonapi_resources.rb` in order for the README documentation to work.

I don't think a change to the README is necessary, but I'll pull request if someone thinks it would be helpful.

lgebhardt:
Thanks for the explanation. It might be good to add a note to the README. I can see this biting others.

jakubrohleder:
I had same problem today, adding this to the documentation would be a good idea!

PR #475, opened 2015-10-05
Adds requires to custom class examples in README

MetricMike:
Addresses confusion about initializers and load order in #449


lgebhardt:
Thanks!

Issue #301, opened 2015-07-09
Cursor Pagination missing?


andresberrios:
Hi! I'm sorry if this is a silly question but I can't seem to find the Cursor Paginator, even though it's mentioned in the Readme. I really want to use it if it's there somewhere, before spending time rolling up my own.
If there's no cursor paginator, I've been thinking of using https://github.com/Kukunin/cursor_pagination (although it seems a bit old) to integrate it as a resource paginator.
What do you think?

lgebhardt:
You are correct that cursor pagination isn't implmented yet. Some work was done, but it didn't work correctly so I pulled it and haven't gone back to it. It should have been removed from the readme (until the feature is finished). Sorry for the confusion.

andresberrios:
Ok @lgebhardt , thanks for the answer!
By the way, do you know of any good, reliable gem that implements cursor pagination?

lgebhardt:
@andresberrios Sorry, I don't know of a good gem at this point.

xdmx:
@andresberrios if you find a way / work out a way to have cursor pagination with jsonapi-resources it'd be great to have it as a pull request / included in the gem itself along with the other 2 existing ways (paged & offset) :smiley:

andresberrios:
@xdmx sorry, I was going to integrate another gem as a custom paginator, but now I've struggled for several days trying to have custom actions and overriding default behaviour in JR and I've found no remotely simple way, so I'm going to have to go back to normal rails for now. I really want to see how JR evolves, but for now I have to move on.
Still, thanks for the awesome work.

PR #477, opened 2015-10-05
Remove reference to cursor paginator

thomassnielsen:
Removes reference to the non-existing cursor paginator.

See #301


dgeb:
Good call - that should minimize confusion. Thanks @thomassnielsen :+1:

PR #478, opened 2015-10-05
Add example for 'ensure_correct_media_type' in 'Controllers' section

tiagopog:
In my real world application, I needed to make sure that clients were requesting content with the correct media type "application/vnd.api+json", for some custom actions other than create, update, create_relationship and update_relationship. Then I figured out that the way to do so was by setting "before_action" with the callback "ActsAsResourceController#ensure_correct_media_type". Currently there's a lack of documentation for this use case, so I thought that would be great moment to do it.

Hope it helps. Cheers!


lgebhardt:
Thanks!

PR #483, opened 2015-10-08
Fix get_related_resources when record_for_related_resources has join

### LINKED ISSUE #482: Related resource that joins other association breaks
So I have the following models setup

```
class Article < ActiveRecord::Base
  has_and_belongs_to_many :internal_categories
end

class InternalCategory < ActiveRecord::Base
  has_and_belongs_to_many :articles
  has_one                 :internal_category_stat
end

class InternalCategoryStat < ActiveRecord::Base
  belongs_to :internal_category
end
```

and this resource:

```
class Api::V1::ArticleResource < JSONAPI::Resource
  has_many  :categories,   relation_name: 'internal_categories'

  attributes :title, :summary, :content, :published_at

  def records_for_categories(options = {})
    @model.internal_categories
      .joins(:internal_category_stat)
      .select('internal_categories.*, internal_category_stats.total_used')
  end
end
```

The reason I select additional fields from the stat model is to be provide sort by them. However, even this call breaks `/api/v1/articles/2/categories`.

```
PG::UndefinedFunction: ERROR:  function count(internal_categories, integer) does not exist
LINE 1: SELECT COUNT(internal_categories.*, internal_category_stats....
               ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
: SELECT COUNT(internal_categories.*, internal_category_stats.total_used) FROM "internal_categories" INNER JOIN "internal_category_stats" ON "internal_category_stats"."internal_category_id" = "internal_categories"."id" INNER JOIN "articles_internal_categories" ON "internal_categories"."id" = "articles_internal_categories"."internal_category_id" WHERE "articles_internal_categories"."article_id" = $1
Internal Server Error: PG::UndefinedFunction: ERROR:  function count(internal_categories, integer) does not exist
LINE 1: SELECT COUNT(internal_categories.*, internal_category_stats....
               ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
: SELECT COUNT(internal_categories.*, internal_category_stats.total_used) FROM "internal_categories" INNER JOIN "internal_category_stats" ON "internal_category_stats"."internal_category_id" = "internal_categories"."id" INNER JOIN "articles_internal_categories" ON "internal_categories"."id" = "articles_internal_categories"."internal_category_id" WHERE "articles_internal_categories"."article_id" = $1 /usr/local/opt/rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/activerecord-4.2.4/lib/active_record/connection_adapters/postgresql_adapter.rb:641:in `prepare'
```

 I think the problem lies in [`JSONAPI::ShowRelatedResourcesOperation#record_count`](https://github.com/cerebris/jsonapi-resources/blob/d0c68d605993e133ea4f96393cbbf7dfdd3a7aea/lib/jsonapi/operation.rb#L153) due to Rails [issue #15138](https://github.com/rails/rails/issues/15138). What do you think of passing `:all` to the `#count` method?

apraditya:
`#get_related_resources` breaks when the resource has `#records_for_` that joins to other association and selects some columns.

This PR will fix #482.

### DIFF
--- a/lib/jsonapi/operation.rb
+++ b/lib/jsonapi/operation.rb
@@ -150,7 +150,7 @@ module JSONAPI
     end
 
     def record_count
-      @_record_count ||= records.count
+      @_record_count ||= records.count(:all)
     end
 
     def source_resource
--- a/test/controllers/controller_test.rb
+++ b/test/controllers/controller_test.rb
@@ -2989,6 +2989,18 @@ class Api::V1::MoonsControllerTest < ActionController::TestCase
                                 }
 
    end
+
+   def test_get_related_resources_with_select_some_db_columns
+     PlanetResource.paginator :paged
+     original_config = JSONAPI.configuration.dup
+     JSONAPI.configuration.top_level_meta_include_record_count = true
+     JSONAPI.configuration.json_key_format = :dasherized_key
+     get :get_related_resources, {planet_id: '1', relationship: 'moons', source: 'api/v1/planets'}
+     assert_response :success
+     assert_equal 1, json_response['meta']['record-count']
+   ensure
+     JSONAPI.configuration = original_config
+   end
 end
 
 class Api::V1::CratersControllerTest < ActionController::TestCase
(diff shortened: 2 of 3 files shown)


lgebhardt:
Thanks!

Issue #464, opened 2015-09-27
How to add a custom link


ceritium:
Hi,

I have the resource `Specification` and I want add a link to the raw file like:

``` json
{
  "data": [
  {
    "id": "26",
    "type": "specifications",
    "links": {
      "self": "http://localhost:3000/api/specifications/26",
      "raw": "http://localhost:3000/api/specifications/26/raw"
    }
},
```

How I could do it?

Thanks

lgebhardt:
There isn't yet an official way to add additional links. Eventually I'd like to be able to add custom links to the resource and the relationships, but as of now we haven't started work on that.

I'm not sure the best way to accomplish this. Probably the simplest way would be to patch the `ResourceSerializer.relationship_links` and test for the source being equal to the specifications class and if so adding the additional link. It's ugly, but should work until we implement a better way.

If you, or anyone else, want to tackle adding this as a PR that would be even better. Eventually I'd like to have a `link` method on the resource to add a new link and provide a lambda to generate the link.

ceritium:
Thank you very much

I'm not sure when I could try the patch, if works I will reply about that.

For my could be closed this issue :D

akharris:
@ceritium +1 I had the same need for links that looked like the following

![screenshot 2015-09-30 12 48 27](https://cloud.githubusercontent.com/assets/5504772/10201329/71c18ca0-6779-11e5-9b86-fc26fd61a81b.png)

@lgebhardt I had the same need and have a [branch](https://github.com/cerebris/jsonapi-resources/compare/master...AKHarris:feature/custom-links-defined-per-resource) that implements a `custom_link` method that's defined on the Resource class. For example if you wanted to generate @ceritium's example you would define the SpecificationResource class as follows.

``` ruby
class SpecificationResource
  attributes :title, :description #etc

  custom_link :raw, :self
end
```

where the first argument `raw` is the name of the key and `self` is the 'kind' of url you want to generate, which in this case is an extension of the self url. While not yet implemented I, I wanted to leave open the possibility of defining custom links that hang off of the various relationship urls by passing in `:relationship` instead of `:self`.

if you wanted a longer path extended from the self link you could do that as follows. You could also add a custom extension

``` ruby
class SpecificationResource
  attributes :title, :description #etc

  custom_link :raw, :self, path: "super/duper/custom/path", extension: :xml
end
```

which would generate the following

```
{
  "data": [
  {
    "id": "26",
    "type": "specifications",
[...]

PR #486, opened 2015-10-12
Feature/custom links defined per resource

akharris:
based on conversation in #464 and some code I managed to extract from a recent spike using this library.

cc @ceritium
### Overview

I've made some modifications since my initial PR. I've simplified the custom link building interface so that now the `custom_link` method simply takes a `JSONAPI::Resource` instance and a `LinkBuilder` object. **I'm open to reverting to an older version of the PR**. This current implementation does depend on knowing how to use internal parts of the gem since it largely relies on the passed-in link builder object.

I've also backtracked on having code that adds a link with an if condition. I think it's much cleaner to return a null for a custom link than it is to remove the entire node, and we can always re-add the capability to completely add/remove custom links based on Resource instance-level business logic.
### Examples

In this example, the url simply hangs off the the normal 'self' link.

``` ruby
class SpecificationResource < JSONAPI::Resource
  attributes :title, :description #etc

  custom_link :raw, ->(source, link_builder) { link_builder.self_link(source) + "/raw" }
end
```

generates the following

```
{
  "data": [
  {
    "id": "26",
    "type": "specifications",
    "links": {
      "self": "http://example.com/api/specifications/26",
      "raw": "http://example.com/api/specifications/26/raw.xml"
    }
}
```

Slightly more complicated custom link

``` ruby
class SpecificationResource < JSONAPI::Resource
  attributes :title, :description #etc

  custom_link :raw, ->(source, link_builder) { link_builder.self_link(source) + "/super/duper/path.xml" }
end
```

generates

```
{
  "data": [
  {
    "id": "26",
    "type": "specifications",
    "links": {
      "self": "http://example.com/api/specifications/26",
      "raw": "http://example.com/api/specifications/26/super/duper/path.csv"
    }
}
```

``` ruby
class ProjectProposalResource < JSONAPI::Resource
  attributes :title, :description, :approved, :zoning_code

  has_one :department
  has_many :tasks

  # make the `approved?` method available to the Resource instance
  delegate :approved? to: :model

  custom_link :some_weird_gov_metadata_standard, ->(source, link_builder)  do
    if source.approved?
      link_builder.self_link(source) + "/some/metadata/standard.xml"
    end
  end
end
```

```
# example.com/api/project-proposals
# 
# ProjectProposal 12 is approved but ProjectProposal 13 is not
{
  "data": [
  {
    "id": "12",
    "type": "projectProposals",
    "links": {
      "self": "http://example.com/api/project-proposals/12",
      "some_weird_gov_metadata_standard": "http://example.com/api/project-proposals/12/some_weird_gov_metadata_standard.xml"
    },
    "attributes": {
      ...
    }
  {
    "id": "13",
    "type": "projectProposals",
    "links": {
      "self": "http://example.com/api/project-proposals/13",
      "some_weird_gov_metadata_standard": null
    },
    "attributes": {
      ...
    }
  }
}
```

Here are a couple of contrived examples that use lambdas.

``` ruby
class ArticleResource < JSONAPI::Resource
  attributes :title, :body, :published_at

  # make the method `original_source` available to the model
  delegate :original_source, to: :model

  custom_link :original_source, ->(source, link_builder) do 
    source.original_source # url stored on the model
  end
end
```

```
{
  "data": [
  {
    "id": "2",
    "type": "articles",
    "links": {
      "self": "http://example.com/api/articles/2",
      "original_source": "http://www.cerebris.com/blog/2015/06/04/jsonapi-1-0/"
    }
}
```

``` ruby
class ArticleResource < JSONAPI::Resource
  attributes :title, :body, :published_at

  custom_link :v1_api_link, ->(source, link_builder) do 
    Figaro.env.domain + "/v1/some-other-part-of-your-api"
  end
end
```

```
{
  "data": [
  {
    "id": "2",
    "type": "articles",
    "links": {
      "self": "http://example.com/api/v2/articles/2",
      "v1_api_link": "http://example.com/v1/some-other-part-of-your-api"
    }
}
```

### DIFF
--- a/lib/jsonapi/link_builder.rb
+++ b/lib/jsonapi/link_builder.rb
@@ -54,12 +54,21 @@ module JSONAPI
       custom_links = {}
 
       link_instructions.each do |link|
-        key = link.first
         link_info = link.last
-        type = link_info[:type]
-        ext = link.last[:ext]
 
-        custom_links[key] = send("build_custom_#{type}_link", key, source, link_info)
+        key  = link_info[:name]
+        type = link_info[:type]
+        ext  = link_info[:ext]
+        if_condition = link_info[:if]
+        next if if_condition && if_condition.call(source) == false
+
+        case type.to_sym
+        when :self
+          custom_links[key] = self_link_extension(key, source, link_info)
+        when :custom
+          custom_links[key] = build_custom_link(key, source, link_info)
+        end
+
       end
 
       custom_links
@@ -67,9 +76,9 @@ module JSONAPI
 
     private
 
-    def build_custom_self_link(key, source, link_info={})
+    def self_link_extension(key, source, link_info={})
       extension = link_info[:ext]
-      path = link_info[:relative_path] || key
+      path = link_info[:path] || key
 
       url = "#{ self_link(source) }/#{ path }"
       url += ".#{extension}" if extension
@@ -77,8 +86,12 @@ module JSONAPI
       url
     end
 
-    # method for building custom links from relationships
-    def build_custom_relationship_link(key, source, link_info={})
+    def build_custom_link(key, source, link_info)
+      extension = link_info[:ext]
+      url = link_info[:with].call(source)
+      url += ".#{ extension }" if extension
+
+      url
     end
 
     def build_engine_name
--- a/lib/jsonapi/link_builder.rb
+++ b/lib/jsonapi/link_builder.rb
@@ -49,7 +49,29 @@ module JSONAPI
       end
     end
 
+    def build_custom_links(source)
+      link_instructions = source.custom_links
+      custom_links = {}
+
+      link_instructions.each do |link|
+        key = link.first
+        type = link.last[:type]
+        ext = link.last[:ext]
+
+        custom_links[key] = send("build_custom_#{type}_link", key, source, ext)
+      end
+
+      custom_links
+    end
+
     private
+
+    def build_custom_self_link(key, source, ext=nil)
+      url = "#{ self_link(source) }/#{ key }"
+      url += ".#{ext}" if ext
+
+      url
+    end
 
     def build_engine_name
       scopes = module_scopes_from_class(primary_resource_klass)
--- a/lib/jsonapi/link_builder.rb
+++ b/lib/jsonapi/link_builder.rb
@@ -55,10 +55,11 @@ module JSONAPI
 
       link_instructions.each do |link|
         key = link.first
-        type = link.last[:type]
+        link_info = link.last
+        type = link_info[:type]
         ext = link.last[:ext]
 
-        custom_links[key] = send("build_custom_#{type}_link", key, source, ext)
+        custom_links[key] = send("build_custom_#{type}_link", key, source, link_info)
       end
 
       custom_links
@@ -66,11 +67,18 @@ module JSONAPI
 
     private
 
-    def build_custom_self_link(key, source, ext=nil)
-      url = "#{ self_link(source) }/#{ key }"
-      url += ".#{ext}" if ext
+    def build_custom_self_link(key, source, link_info={})
+      extension = link_info[:ext]
+      path = link_info[:relative_path] || key
+
+      url = "#{ self_link(source) }/#{ path }"
+      url += ".#{extension}" if extension
 
       url
+    end
+
+    # method for building custom links from relationships
+    def build_custom_relationship_link(key, source, link_info={})
     end
 
     def build_engine_name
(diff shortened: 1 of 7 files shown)


lib/jsonapi/resource.rb
```diff
@@ -279,7 +288,33 @@ def resource_for(type)
...
+        link_manifest = { name: name.to_sym, type: type.to_sym }
+
+        ext  = options[:ext]  || options[:extension]
+        path = options[:path] || options[:relative_path]
+        url  = options[:url] 
+        with = options[:with]
+        if_condition   = options[:if] 
+
+        if type == :custom && with.nil?
+          warn "#{ self.class.name }'s custom link #{name} needs a a lambda passed-in as the 'if' option, like `if: ->(instance) { # do stuff here }`"
```
lgebhardt: The warning refers to the `if` option but the check is for the `with` option. Is this intended?

Also, this line isn't being covered in the tests.  We should add a test for it, or wrap it with `# :nocov:` directives.

beechnut:
This example

``` ruby
class ArticleResource < JSONAPI::Resource
  attributes :title, :body, :published_at

  custom_link :original_source, :custom, with: ->(instance) { instance.original_source }
end
```

doesn't work, in my experience, unless `#original_source` is an attribute on the Resource. Since `instance` is the resource itself, and not the model, it doesn't have access to methods defined on the model that are not defined on the resource.

In order to get my custom next-record and previous-record links to work, it was necessary to call methods on the model directly:

``` ruby
class ProjectResource < JSONAPI::Resource
  u = Rails.application.routes.url_helpers
  custom_link :next,     :custom, with: ->(i) { u.project_url(i.model.next.presence) if i.model.next }
  custom_link :previous, :custom, with: ->(i) { u.project_url(i.model.previous) if i.model.previous }
end
```

lib/jsonapi/resource.rb
```diff
@@ -115,6 +115,14 @@ def records_for(relation_name)
       model.public_send relation_name
     end
 
+    def custom_links?
+      custom_links.size > 0
+    end
+
+    def custom_links
```
lgebhardt: Do we need the instance level `custom_links` and `custom_links?` methods? It seems we could just directly access the class level methods.

lib/jsonapi/link_builder.rb
```diff
@@ -49,8 +49,51 @@ def self_link(source)
...
+
+      url = "#{ self_link(source) }/#{ path }"
+      url += ".#{extension}" if extension
+
+      url
+    end
+
+    def build_custom_link(key, source, link_info)
+      extension = link_info[:ext]
+      url = link_info[:with].call(source)
```
lgebhardt, line 77: Would it make sense to pass in the link_builder to the lambda? That way the methods it provides could be used.

lib/jsonapi/resource.rb
akharris: yeah that's a typo, one that would've been caught with the spec I didn't write :smiley: 

I'll add a spec that catches when a custom link type is made with no `with` lambda.

lib/jsonapi/link_builder.rb
akharris, line 77: :+1: that's an awesome idea and could make `custom_link` simpler. I'll see where I can take it in the next few days.

akharris:
@beechnut good catch, I updated the top comment with your corrections. :+1: 

I think the following should work as well.

``` ruby
class ProjectResource < JSONAPI::Resource
  u = Rails.application.routes.url_helpers
  delegate :presence, to: :model
  delegate :next, to: :model
  delegate :previous, to: :model

  custom_link :next,     :custom, with: ->(i) { u.project_url(i.next) if i.next }
  custom_link :previous, :custom, with: ->(i) { u.project_url(i.previous) if i.previous }
end
```

lib/jsonapi/resource.rb
akharris: probably not. In `ResourceSerializer#relationship_links` I call `source.custom_links?` but that could just as easily be changed to `source.class.custom_links?`

lgebhardt: Thanks. Any methods we put at the instance level limit the available names for attributes.

akharris:
sorry I've been taking some time off lately and will be offline the next two weeks. I'll resume work on this PR when I get back (unless someone else has added the same feature).

lgebhardt:
In the new resource level meta feature I'm passing in `custom_generation_options` (https://github.com/cerebris/jsonapi-resources/blob/master/lib/jsonapi/resource_serializer.rb#L133) which contains the serializer. From there you can get the link builder. It would be nice if we could reuse the `custom_generation_options` function as the parameter for the custom links.

akharris:
@lgebhardt I'm back from my vacation. I'll look into incorporating `custom_generation_options` over the next week.

PR #487, opened 2015-10-13
Rename pagination "previous" link to "prev" per JSON API spec

willcosgrove:
We were running into this issue when we were using a JSON API client library that expected the pagination link for the previous page to be named `prev`, rather than `previous`.  It looks like that is what the official spec says it should be: http://jsonapi.org/format/#fetching-pagination


lgebhardt:
Good catch. Thanks!

PR #489, opened 2015-10-13
Making sure the module_path takes a namespaced module_name into account

nathanpalmer:
I'm submitting this to start the conversation. We're having trouble getting this to work properly with our namepaced model names. Over half of them are namespaced and I'm experiencing inconsistencies. See this example:

`/api/project_resource.rb`

``` ruby
class API::ProjectResource
  has_many :people
end
```

`/api/project/people_resource.rb`

``` ruby
class API::Project::PeopleResource
  has_one :main_product, class_name: 'Product'
  has_one :secondary_product, class_name: 'Product'
end
```

`/api/product_resource.rb`

``` ruby
class API::ProductResource
end
```

1) In order for it to find the resource for one of our models it must be in a sub directory (which I prefer) as you can see from the /api/project/ directory

2) Sometimes it's looking for the product_resource under /api/project and sometimes it's not. The easiest way for me to replicate it is to include the main_product and then right afterward change it to include the secondary product. It usually fails right then with this error

`JSONAPI: Could not find resource 'api/project/Product'. (Class API::Project::ProductResource not found)`

After digging in it seems it's because of two things

  1) We're demodularizing the model name for `_model_name` (something I haven't addressed yet)

  2) We're constructing the module_name based on the full name which doesn't properly represent the api path considering when the model name has multiple parts.


lib/jsonapi/resource.rb
```diff
@@ -678,7 +682,8 @@ def _allowed_filter?(filter)
       end
 
       def module_path
-        @module_path ||= name =~ /::[^:]+\Z/ ? ($`.freeze.gsub('::', '/') + '/').underscore : ''
+        n = /::[^:]+\Z/
+        @module_path ||= name.gsub(_model_name =~ n ? "#{$`}::" : '','') =~ n ? ($`.freeze.gsub('::', '/') + '/').underscore : ''
```
lgebhardt: This line will be in conflict when you rebase against master. It should now be:

``` ruby
        name.gsub(_model_name =~ n ? "#{$`}::" : '','') =~ n ? ($`.freeze.gsub('::', '/') + '/').underscore : ''
```

lgebhardt:
The current project assumes the models are not namespaced. It also assumes the referenced resources all exist in the same namespace.

I think it makes sense to allow namespaced models. In hindsight it would have been ideal to implement that as the default. I would like to see some tests that show this working.

I'm a little unsure of the intent of `model_type` method you've added. Is it to get around the resources all must be in the same namespace assumption?

nathanpalmer:
@lgebhardt I've just merged in the latest master to this branch. It took us a while to circle back to this but carving out some time to see if we can get it to work in this project. 

> I'm a little unsure of the intent of model_type method you've added. Is it to get around the resources all must be in the same namespace assumption?

Yes, there isn't a good way that I know of to dynamically determine whether it's a real namespace or a sub-model. So instead opting to simply have an override where you can set the `model_type` and it will be used both in determining the module path as well as returning it as the type for that model.

To give some more visibility most of the models have related children:

``` ruby
class User < ActiveRecord::Base
end

class User::Contact < ActiveRecord::Base
end

class Order < ActiveRecord::Base
end

class Order::Address < ActiveRecord::Base
end
```

We need to maintain the parent model name as part of the type. Normally it would get chopped off and assume that the parent model is simply a namespace. 

This becomes an even bigger issue when we have several child model names that are the name:

``` ruby
class Email < ActiveRecord::Base
end

class Order::Email < ActiveRecord::Base
end
```

Both get identified as `email` when the second should be `order/email`. This change allows us to put a `model_type` property on the second class so it gets identified properly.

``` ruby
class EmailResource < JSONAPI::Resource
end

class Order::EmailResource < JSONAPI::Resource
  model_name 'Order::Email'
  model_type 'order/email'
end
```

nathanpalmer:
I'm going to close this pull request out for now. This is proving to be more difficult than I thought as the resolution of names -> models is in quite a few different places. Some of abstraction over this might help but I'll re-think it and submit again once I have a good idea.

cesarizu:
I've been actually using this (with some other commits) for some time on a production app. You can see my commits here:

https://github.com/GovSciences/jsonapi-resources/commits/master

Issue #490, opened 2015-10-14
Support for sorting by relationship attributes


ghost:
Seems like this isn't supported right now.

The spec says:

> Note: It is recommended that dot-separated (U+002E FULL-STOP, ".") sort fields be used to request sorting based upon relationship attributes. For example, a sort field of author.name could be used to request that the primary data be sorted based upon the name attribute of the author relationship.

Will see if I can whip up a PR.

davidgoli:
:+1: also for filtering

aldhsu:
Closed by #494

PR #494, opened 2015-10-15
[#490] Add support for sorting relationship fields.

aldhsu:
#490

Spike of sorting on relationship fields. 

Some caveats at the moment:
1. Doesn't support nested relationships
2. Haven't tested with multiple joins with the same name
3. You must `include` the relationship in the request if you want to sort on it
4. It only supports belongs to 
5. Request should be in the form `model.field` as per the JSONAPI spec, not the plural form 
6. The relationships field needs to be explicitly included in the `sortable_fields`


ghost:
Would appreciate an ack on this PR. We've been using it in production already without any problems. Any changes we need to make to get this merged?

lgebhardt:
Sorry, I thought I had responded (but I guess I never finished). In general I like the PR, but I'm not so fond of the requirement that you must `include` the relationship. It also doesn't take into account the relationship's resource name and instead uses the model name. For example if we have:

``` ruby
class PersonResource < BaseResource
  attributes :name, :email
  has_many :posts
end

  attributes :title, :body, :subject
end
```

I would expect to sort on `author.name` and not `person.name`, which is what I think the code will currently require to work. The PR has the following test:

``` ruby
def test_sorting_by_relationship_field
  get :index, {include: 'author', sort: 'person.name'}
  assert_response :success
  assert_equal '17', json_response['data'][0]['id']
end
```

I'd like to see that test fail and the following succeed:

``` ruby
def test_sorting_by_relationship_field
  get :index, {sort: 'author.name'}
  assert_response :success
  assert_equal '17', json_response['data'][0]['id']
end
```

I think if we parse the sort parameters that contain `.`s we can determine the resource the request is for, From there we could automatically do a join in the `records` (even nested) and we could use the resource name instead of the underlying model name.

DavidMikeSimon:
It would also be fantastic if filtering on relationship attributes worked the same way.

aldhsu:
I have now implemented sorting by relationship attributes as @lgebhardt has specified.

lgebhardt:
@aldhsu Looking good. Just a few issues. The following code is not getting tested by the current test suite:

https://github.com/kinesisptyltd/jsonapi-resources/blob/0bf3d61cd01e99f34f852c0b819ffe6d650c185e/lib/jsonapi/resource.rb#L542-L544

https://github.com/kinesisptyltd/jsonapi-resources/blob/0bf3d61cd01e99f34f852c0b819ffe6d650c185e/test/fixtures/active_record.rb#L954

You can test the coverage with:

``` bash
COVERAGE=true bundle exec rake test
```

vrybas:
@aldhsu wow that's great! Much needed feature

aldhsu:
We have made some improvements to the PR. It now supports reflexive relationships. There is a few more things to clean up before this PR is ready.
1. Clashes with our filtering on reflexive relationships is a small problem. @lgebhardt I agree that sorting doesn't imply that you want to `include` the model but it would be great to have one true source of inclusions. ie. `records` is not manipulated in each filter and sort with `includes` or `joins`
2. I will address the testing

achan:
Are there plans on merging this to master? Or is there currently a way to fetch relationships in a sorted order?

jcrispw:
Hi @lgebhardt,

@aldhsu and I have been working to clean up and add additional testing coverage for the PR, plus full support for self-referential relationships. We think it's ready to go. Can you let us know if there's anything else you need?

Thanks
James

aldhsu:
@achan this PR is about sorting primary resources by relationships not sorting included resources. I don't know if that is something that is addressed in the JSONAPI spec.

lgebhardt:
@aldhsu @jcrispw et al. Thanks for this very useful feature!

Issue #480, opened 2015-10-06
Attribute named model renders the actual AR model


jimmybaker:
![2015-10-06 at 1 24 am](https://cloud.githubusercontent.com/assets/50668/10301549/0e5b7802-6bc9-11e5-91bf-317ab16eb40f.png)

Is this a bug? If not, how do I get around this without renaming this attribute?

lgebhardt:
Yes, this is a bug. It'll be a day or so before I can look into a solution.

lgebhardt:
I'm changing my stance that this is a bug. I forgot we had reserved the attribute name of `model` but only emit a warning, not explicitly prohibiting the method. We should probably look at renaming the internal attribute named `model` to allow it to be used as an attribute. Until that happens I'm not sure there's a great workaround.

ggordon:
I can look into this, I need it for my app.

lgebhardt:
Thanks again @ggordon

ggordon:
You're welcome.

PR #495, opened 2015-10-15
Allow attribute named 'model'

ggordon:
Accessing the the underlying model uses the `_model_` method instead of
`model`. Accesing by variable still remains `@model`

Fix for #480


lgebhardt:
Thanks for fixing this. One request though: could you rename the `_model_` method to `_model`? This will be more consistent with the other similar methods.

ggordon:
@lgebhardt  Done.

lgebhardt:
@ggordon Thanks!

PR #496, opened 2015-10-15
Silence schema_load messages during test

ggordon:
The schema load messages cluttered the output and made it harder to see actual issues.


lgebhardt:
@ggordon Thanks! That's much better.

PR #498, opened 2015-10-15
Allow 'href' and 'link' as field names

ggordon:
Allow 'href' and 'link' as fields names. Also add test relationships named 'type' or 'id'.

### DIFF
--- a/lib/jsonapi/resource.rb
+++ b/lib/jsonapi/resource.rb
@@ -705,13 +705,13 @@ module JSONAPI
       def check_reserved_attribute_name(name)
         # Allow :id since it can be used to specify the format. Since it is a method on the base Resource
         # an attribute method won't be created for it.
-        if [:type, :href, :links].include?(name.to_sym)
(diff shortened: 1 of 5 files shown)


lgebhardt:
I was actually going to look into whether we could get rid of the reserved field names. I know we need to keep `type` reserved, but I think we can now allow `href` and `links` since we moved all the attributes and relationships under the `attributes` and `relationships` keys.

ggordon:
Let me know what you'd like to do, I was initially just trying to get rid of the deprecation warnings during testing.

lgebhardt:
That would be nice to get rid of the warnings. I assume there's also a simple way of turning off the deprecation warnings, but I've learned to ignore them. 

If you're up for it you could remove the reserver fields except for type, add a test using them and see if it breaks.

ggordon:
@lgebhardt I think this addresses your comment. Please let me know if I misinterpreted something.

lgebhardt:
@ggordon Thanks again!

Merged by #510.

PR #499, opened 2015-10-15
Simplify `verify_key` method in resource

galiat:
- The switch statement returns a proc, only to `.call` it. Remove the call statement and run the code directly in the switch statement.
- Remove rescue block in `:integer` case. The method will rescue appropriately without it.
- Rather than `return key if nil` just return

This also removes rspec warnings (x3):
warning: shadowing outer local variable - key 
warning: shadowing outer local variable - context

### DIFF
--- a/lib/jsonapi/resource.rb
+++ b/lib/jsonapi/resource.rb
@@ -554,40 +554,28 @@ module JSONAPI
 
       def verify_key(key, context = nil)
         key_type = resource_key_type
-        verification_proc = case key_type
 
+        case key_type
         when :integer
-          -> (key, context) {
-            begin
-              return key if key.nil?
-              Integer(key)
-            rescue
-              raise JSONAPI::Exceptions::InvalidFieldValue.new(:id, key)
-            end
-          }
+          return if key.nil?
+          Integer(key)
         when :string
-          -> (key, context) {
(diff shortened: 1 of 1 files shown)


lib/jsonapi/resource.rb
```diff
@@ -554,40 +554,28 @@ def resource_key_type
...
-              raise JSONAPI::Exceptions::InvalidFieldValue.new(:id, key)
-            end
-          }
+          return if key.nil?
+          if key.to_s.match(/^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$/)
+            key
+          else
+            raise JSONAPI::Exceptions::InvalidFieldValue.new(:id, key)
+          end
         else
```
galiat, line 576: @adamworrall Would you expect that `key` should be a proc if we end up in this `else` block? Based on the old code, and the tests, I suspect so, since you were calling `.call` on the result of the switch statement.

Did you consider adding another `when` block for the case that key_type is a proc? It would be more clear and free up the else block for error handling.

lgebhardt:
@galiat Thanks! That does clean things up.

PR #500, opened 2015-10-15
Test deprecated id attribute

ggordon:
Test setup was showing deprecation warnings on `attribute :id`, fixed those and added an explicit test since I didn't see one elsewhere.


lgebhardt:
@ggordon Thanks, that's a nice improvement.

Issue #504, opened 2015-10-20
Underlying validation error messages lost if save fails in an after_save block


theirishpenguin:
Not best practice but it is possible to have validation errors set in an ActiveRecord after_save block raising an ActiveRecord::RecordInvalid exception - as outlined at http://tech.taskrabbit.com/blog/2013/05/23/rollback-after-save. A conventional rails web application will correctly handle this and display the correct errors on the UI.

However, when using jsonapi-resources this does not happen because jsonapi-resources assumes that if `@model.save` is false then it is not worth polling the object for errors. The code is here...

https://github.com/cerebris/jsonapi-resources/blob/master/lib/jsonapi/resource.rb#L148

theirishpenguin:
Potential solution... https://github.com/cerebris/jsonapi-resources/pull/505/files

vrybas:
@theirishpenguin fixed by #505 ?

theirishpenguin:
Indeed it is, closing now. Thanks @vrybas

PR #505, opened 2015-10-20
Fix for #504 - surface correct validation error messages if save…

theirishpenguin:
... fails in an after_save block.

Note: Need to add tests, if this is deemed to be in the right direction.

### DIFF
--- a/test/unit/resource/resource_test.rb
+++ b/test/unit/resource/resource_test.rb
@@ -21,10 +21,6 @@ end
 class ArticleWithBadAfterSaveResource < JSONAPI::Resource
   model_name 'PostWithBadAfterSave'
   attribute :title
-
-  def self.records(options)
-    options[:context].posts
-  end
 end
 
 class NoMatchResource < JSONAPI::Resource
--- a/test/unit/resource/resource_test.rb
+++ b/test/unit/resource/resource_test.rb
@@ -8,6 +8,25 @@ class ArticleResource < JSONAPI::Resource
   end
 end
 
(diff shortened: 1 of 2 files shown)


lgebhardt:
I'm ok with this change even though it's a non-normal situation. Could you please rebase against master and add a test for the new code (check for 100% coverage using `COVERAGE=true rake test` please).

theirishpenguin:
Great @lgebhardt. I hope to get some time this week to complete this and will do the actions listed in your comment. Thanks for the quick reply.

theirishpenguin:
Hi @lgebhardt, I have rebased and updated this PR with a test.

I'm not familiar with 100% of the jsonapi-resource codebase so I may not have implemented the test in the correct style - is it okay?

test/unit/resource/resource_test.rb
```diff
@@ -8,6 +8,25 @@ def self.records(options)
...
+  end
+end
+
+class ArticleWithBadAfterSaveResource < JSONAPI::Resource
+  model_name 'PostWithBadAfterSave'
+  attribute :title
+
+  def self.records(options)
+    options[:context].posts
+  end
```
lgebhardt: This method isn't being called in the tests. Could you please remove it?

lgebhardt:
Other than the one uncovered line things look good. Thanks.

theirishpenguin:
Cool. Done, thanks.

lgebhardt:
Thanks!

theirishpenguin:
Great, thanks!

PR #507, opened 2015-10-20
STI (rebase #393) and Immutable option

lgebhardt:
Rebases #393.

Adds an immutable option to allow a resource to be read only. Allows an STI base resource to return a heterogeneous collection, but not modify the collection.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,7 +108,7 @@ end
 
 ##### Immutable Resources
 
-Resources that are immutable should be declared as such with the `immutable method`. Immutable resources will only
+Resources that are immutable should be declared as such with the `immutable` method. Immutable resources will only
 generate routes for `index`, `show` and `show_relationship`.
 
 ###### Immutable for Readonly
--- a/README.md
+++ b/README.md
@@ -108,7 +108,7 @@ end
 
 ##### Immutable Resources
 
-Resources that are immutable should be declared as such with the immutable method. Immutable resources will only
+Resources that are immutable should be declared as such with the `immutable method`. Immutable resources will only
 generate routes for `index`, `show` and `show_relationship`.
 
 ###### Immutable for Readonly
--- a/README.md
+++ b/README.md
@@ -108,19 +108,18 @@ end
 
 ##### Immutable Resources
 
-Resources that are immutable should be declared as immutable. Immutable resources will only generate routes for 
-`index`, `show` and `show_relationship`.
+Resources that are immutable should be declared as such with the immutable method. Immutable resources will only
+generate routes for `index`, `show` and `show_relationship`.
 
 ###### Immutable for Readonly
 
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -106,6 +106,51 @@ class ContactResource < BaseResource
 end
 ```
 
+##### Immutable Resources
+
+Resources that are immutable should be declared as immutable. Immutable resources will only generate routes for
```
dgeb: How about: "Resources that are immutable should be declared as such with the `immutable` method."

README.md
```diff
@@ -106,6 +106,51 @@ class ContactResource < BaseResource
...
 ```
 
+##### Immutable Resources
+
+Resources that are immutable should be declared as immutable. Immutable resources will only generate routes for 
+`index`, `show` and `show_relationship`.
+
+###### Immutable for Readonly
+
+Some resources are readonly and are not to be modified through the API. Declaring a resource as immutable prevents
```
dgeb: read-only

dgeb: Also: "Declaring a resource as immutable prevents creation of routes that allow modification of the resource."

README.md
```diff
@@ -106,6 +106,51 @@ class ContactResource < BaseResource
...
+the routes that allow modification of the resource from being created.
+
+###### Immutable for Heterogeneous Collections
```
dgeb: Simply: "Immutable Heterogeneous Collections"

README.md
```diff
@@ -106,6 +106,51 @@ class ContactResource < BaseResource
...
+
+Immutable resources can be used to create a common resource that can be used for fetching heterogeneous collections
+of resources. Manipulation of these resources will still follow the rules setup by the specific resource types.
+
```
dgeb, line 120: "Immutable resources can be used as the basis for a heterogeneous collection. Resources in heterogeneous collections can still be mutated through their own type-specific endpoints."

README.md
```diff
@@ -106,6 +106,51 @@ class ContactResource < BaseResource
...
+# routes
+  jsonapi_resources :vehicles
+  jsonapi_resources :cars
+  jsonapi_resources :boats
+
+```
+
+In the above example vehicles are immutable. A call to `/vehicles` or `/vehicles/1` will return vehicles with types
+of either `car` or `boat`. But calls to put a new `car` must be made to `/cars`. The rails models backing the above code
```
dgeb: s/put/post/ a new `car`

README.md
```diff
@@ -106,6 +106,50 @@ class ContactResource < BaseResource
 end
 ```
 
+##### Immutable Resources
+
+Resources that are immutable should be declared as such with the immutable method. Immutable resources will only
```
dgeb: sorry to nit, but I would use backticks around `immutable`

dgeb:
Looks good now - thanks @lgebhardt :shipit:

PR #511, opened 2015-10-22
Resource meta

lgebhardt:
Adds resource meta support.

Adds serialization options to be used in generating resource level meta and resource level links through procs.

Comments on the `custom_generation_options` method in the serializer would appreciated. I'm thinking this will also be used in the generation of custom links (see #486). As it's just a hash we should be able to easily change it, but I'd like to get it right to start.

### DIFF
--- a/README.md
+++ b/README.md
@@ -274,7 +274,9 @@ updating the attribute. See the [Value Formatters](#value-formatters) section fo
 
 ##### Flattening a Rails relationship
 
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -793,6 +796,31 @@ Will get you the following payload by default:
 }
 ```
 
+#### Resource Meta
+
+Meta information can be included for each resource. The `meta` information to include with each resource as it is 
+serialized is set using the `meta` method in the resource declaration. For example:
```
dgeb, line 802: How about:

> Meta information can be included for each resource using the `meta` method in the resource declaration. For example:

lgebhardt:
Closing in favor of another approach.

Issue #419, opened 2015-09-02
how to manage field aliases in errors


senid231:
for example we have such model:

``` ruby
class Product < ActiveRecord::Base
  validate_uniqueness_of :product_name
  alias_attribute :name, :product_name
end
```

and resource:

``` ruby
class ProductResource < JSONAPI::Resource
  attribute :name
end
```

on create/update resource we can get such response:

``` ruby
{
  erorrs: [{
    code: 100
    status: :unprocessable_entity,
    title: message,
    detail: "product_name - has already been taken",
    source: { pointer: "/data/attributes/product_name" }
  }]
}
```

but client don't know anything about `product_name`.
if model used not only in for `JSON API` we can't just change validator or rename db column ( some other code or client of old version API could check this name).

lgebhardt:
That name is coming from the model's `errors.messages` call. I'm not sure if there's a way to get rails to use the alias instead.

A few ideas that I have not tried:
- put the validation on the alias (no idea if it will still work)
- override the `errors` method on the model and replace any keys that == product_name with just name

senid231:
PR #515 solves the issue

PR #515, opened 2015-10-25
add model_error_messages to resource

senid231:
related to #419 
we will have ability to  change model error keys/values


lgebhardt:
Thanks, looks like a good solution. We should write this up in the readme.

senid231:
conflicts has been resolved :)

lgebhardt:
Thanks!

PR #518, opened 2015-10-27
Check that destroys actually succeed.

ghost:
Original implementation did not check return value from calls to `destroy`, allowing failed destroys to return to the client as successful. Use existing exceptions to return errors in cases where it does fail.

### DIFF
--- a/lib/jsonapi/resource.rb
+++ b/lib/jsonapi/resource.rb
@@ -174,8 +174,9 @@ module JSONAPI
     end
 
     def _remove
-      @model.destroy
-
+      unless @model.destroy
+        fail JSONAPI::Exceptions::ValidationErrors.new(self)
+      end
       :completed
     end
 
--- a/test/controllers/controller_test.rb
+++ b/test/controllers/controller_test.rb
@@ -1653,6 +1653,14 @@ class PostsControllerTest < ActionController::TestCase
     assert_response :bad_request
   end
 
+  def test_delete_with_validation_error
+    post = Post.create!(title: "can't destroy me", author: Person.first)
+    delete :destroy, { id: post.id }
+
+    assert_equal "can't destroy me", json_response['errors'][0]['title']
+    assert_response :unprocessable_entity
+  end
+
   def test_delete_single
     initial_count = Post.count
     delete :destroy, {id: '4'}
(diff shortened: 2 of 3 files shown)


lib/jsonapi/resource.rb
```diff
@@ -157,8 +157,9 @@ def _save
     end
 
     def _remove
-      @model.destroy
-
+      unless @model.destroy
+        fail JSONAPI::Exceptions::ValidationErrors.new(self)
```
lgebhardt, line 178: It seems this should be its own custom Exception type and not a ValidationError. Or am I misunderstanding the ways `destroy` could fail?

ghost, line 178: I see 2 ways that a destroy could fail:
- A runtime exception, caused by a DB constraint or something else like lock or connection error. This should work _without_ this PR as an exception should be thrown from the database adapter
- A validation error on the model, added to the model via a `before_destroy` callback.

Our use case is we enforce some business rules in a `before_destroy` callback. This callback is similar to this:

``` ruby
def check_some_business_rule
   errors.add(:base, "You can't delete this record because of xyz")
   false
end
```

We could create another exception class but I do believe the second type is a form of validation error, at least within the context of Rails.

lgebhardt, line 178: That makes sense. We could check the model for errors and if there are any do the fail as you are. If there are not we could fail with a new error (or errors if we can figure out what went wrong).

thibaudgg:
:+1: I exactly need that now :smile:.

barelyknown:
Good catch.

thibaudgg:
In our use-case we would like to require the User#current_password to be send on delete.
Would it be possible to set `data.attributes` on `Resource#_model, maybe with something like`self.destroyable_fields`?

``` http
DELETE /users/1 HTTP/1.1

{
  "data": {
    "id": "1",
    "type": "users",
    "attributes": {
      "current_password": "1234"
    }
  }
}
```

jcrispw:
@lgebhardt This patch has worked well for us in production for a few months.We've updated it with latest changes from master. Are you happy to merge?  We're keen to move back to mainline rather than our fork. Thanks

lgebhardt:
@Soliah Thanks! Merged by way of #604.

PR #522, opened 2015-10-29
Rework request processing

### LINKED ISSUE #458: Trying to interject authorization, but context is always nil

lgebhardt:
By avoiding the before and after actions any derived controllers will avoid conflicts with actions/filters.

Fixes #458

### DIFF
--- a/lib/jsonapi/acts_as_resource_controller.rb
+++ b/lib/jsonapi/acts_as_resource_controller.rb
@@ -6,52 +6,69 @@ module JSONAPI
 
     included do
       before_action :ensure_correct_media_type, only: [:create, :update, :create_relationship, :update_relationship]
-      append_before_action :setup_request
-      after_action :setup_response
     end
 
     def index
-      process_request_operations
+      setup_and_process_request
     end
 
     def show
-      process_request_operations
+      setup_and_process_request
     end
 
     def show_relationship
-      process_request_operations
+      setup_and_process_request
     end
 
     def create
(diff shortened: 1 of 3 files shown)


lib/jsonapi/acts_as_resource_controller.rb
```diff
@@ -6,52 +6,69 @@ module ActsAsResourceController
...
-      process_request_operations
+      setup_and_process_request
     end
 
     def get_related_resources
-      process_request_operations
+      setup_and_process_request
+    end
+
+    def setup_and_process_request
```
dgeb: perhaps just `process_request` ?

Issue #502, opened 2015-10-16
Internal Server Errors in non-production modes


iamvery:
It's pretty common for us to use errors to drive a feature. Unfortunately the catchall "Internal Server Error" gobbles up a lot of context of what went wrong, and that makes it hard to figure out what to do next in TDD. I'd like to come up with a way to expose that context in non-prod environments so that we can debug problems.

To be sure, the backtrace is printed to the console if the error occurs while the server is running locally, but if you're driving a feature with something like Capybara it's not easily visible (or least idk where to find this backtrace?). When I think about how this works with a run-of-the-mill acceptance test of some server-rendered Rails page, we often lean on the fact that Rails renders a "special" error page in non-prod environments which includes information about the error that occurred.. This got me to the conclusion that perhaps we should extend the internal server error (and maybe others?) to include the exception in the response in non-prod environments.

What do you think?

iamvery:
That issue number... 5xx :heart_eyes: :smirk:

lgebhardt:
Seems like a good addition to me. :+1:

iamvery:
Nice! I'll make myself and note and try to come up with an implementation :soon:

lgebhardt:
Thanks

iamvery:
See #524

PR #524, opened 2015-11-05
Include exception message and backtrace in response

iamvery:
Following #502, this is a simple potential solution to communicating exception information in non-prod.

It bothers me a bit to submit a change that doesn't have a test, but wasn't clear where such a test would exist as the response of internal server errors doesn't appear to be tested. To that end, this is something of a development support feature and presumably as long as it doesn't break production functionality it should be safe to ship.
### TIL

One thing I did realize while I was working on this is that the backtrace produced by the error _is_ logged to `log/test.log`. So one could conceivably monitor this file if they encountered an unexpected error in test. I must have failed to realize this before simple because it's not my typical workflow. Maybe that's not true for others?

---

So, what do you think? Is this useful?

### DIFF
--- a/lib/jsonapi/error.rb
+++ b/lib/jsonapi/error.rb
@@ -1,6 +1,6 @@
 module JSONAPI
   class Error
-    attr_accessor :title, :detail, :id, :href, :code, :source, :links, :status
+    attr_accessor :title, :detail, :id, :href, :code, :source, :links, :status, :meta
 
     def initialize(options = {})
       @title          = options[:title]
@@ -16,6 +16,7 @@ module JSONAPI
       @links          = options[:links]
 
       @status         = Rack::Utils::SYMBOL_TO_STATUS_CODE[options[:status]].to_s
+      @meta           = options[:meta]
     end
   end
 
--- a/lib/jsonapi/exceptions.rb
+++ b/lib/jsonapi/exceptions.rb
@@ -10,10 +10,17 @@ module JSONAPI
       end
 
       def errors
+        unless Rails.env.prod?
+          meta = Hash.new
+          meta[:exception] = exception.message
+          meta[:backtrace] = exception.backtrace
+        end
+
         [JSONAPI::Error.new(code: JSONAPI::INTERNAL_SERVER_ERROR,
                             status: :internal_server_error,
                             title: 'Internal Server Error',
-                            detail: 'Internal Server Error')]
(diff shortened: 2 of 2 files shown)


lgebhardt:
This does seem useful. However it doesn't look like the [spec](http://jsonapi.org/format/#error-objects) allows for the `exception` and `backtrace` keys as valid members of the error object. @dgeb can you provide some guidance? We could certainly stuff them into the `meta` key.

iamvery:
Ah good call. The meta key would probably suffice.

Interestingly since I created this PR I've been tailing `log/test.log` somewhat happily... Worth fixing up?

lgebhardt:
@iamvery I see it as worthwhile. It's sometimes slightly difficult to tail a log on a qa server.

iamvery:
:+1: I'll try to get an update in soon. Thanks for the feedback!

dgeb:
Just chiming in here to support using `meta`. The spec doesn't allow custom members in the top-level of an error object.

iamvery:
:heart:

iamvery:
I just pushed an update that moves this information to `meta`. Let me know if you'd like anything else done.

lgebhardt:
@iamvery Thanks for patching this up. But I have a few suggestions. I want to avoid having the meta fields output if they are nil. I also think by passing the `meta` option into `Error` we can reuse it for other exceptions easily.

No need to do this now, but I'm thinking I may want to create a configuration option to control whether these are output. I haven't decided if an option is overkill at this point.

iamvery:
Thanks for the feedback! The PR has been updated. I'm thinking that meta being `nil` in the case of prod will omit it from serialization due to https://github.com/cerebris/jsonapi-resources/blob/master/lib/jsonapi/resource_serializer.rb#L134. Is that right?

lgebhardt:
Thanks @iamvery. The errors do not run through that code. They are actually just output as a hash in https://github.com/cerebris/jsonapi-resources/blob/master/lib/jsonapi/response_document.rb#L102 so the nil parameters come through as nils. So by adding `meta` to the base `Error` object you are outputting `meta` to every error, which is what I wanted.

xdmx:
Shouldn't it be `Rails.env.production?` instead of `Rails.env.prod?`

iamvery:
Hah! Good catch. Tests, amitite? I'll send a patch unless someone else gets to it 😆

Issue #523, opened 2015-11-02
Namespaced model not working after STI fix added


ggordon:
If my path for the model and resource differ, the resource cannot be found. 

```
module Legacy
  class FlatPost < ActiveRecord::Base
    self.table_name = "posts"
  end
end

class FlatPostResource < JSONAPI::Resource
  model_name "::Legacy::FlatPost"
  attribute :title
end

class FlatPostsController < JSONAPI::ResourceController
end 
```

and a route

```
jsonapi_resources :flat_posts
```

then 

```
get '/flat_posts' 
```

Gives me
**Could not find resource 'legacy/flat_post'.**

Where did that `legacy` come from?

```
Started GET "/flat_posts" for 127.0.0.1 at 2015-11-02 14:51:52 -0500
Processing by FlatPostsController#index as HTML
  Legacy::FlatPost Load (0.3ms)  SELECT "posts".* FROM "posts"  ORDER BY "posts"."id" ASC
Internal Server Error: JSONAPI: Could not find resource 'legacy/flat_post'. (Class Legacy::FlatPostResource not found) /Code/jsonapi-resources/lib/jsonapi/resource.rb:300:in `resource_for'
//Code/jsonapi-resources/lib/jsonapi/resource.rb:525:in `block in find
```

ggordon:
It looks like `JSONAPI::Resource#resource_type_for(model)` is confusing model path with resource path, most of the time these are the same, but not always. I'm at a loss as to how to fix this, any ideas @lgebhardt ?

lgebhardt:
@ggordon Would it help to make it a requirement than models and resources must reside in the same namespace? It's been something I've been thinking about since seeing #513. This would be a breaking change, but I think it might be for the best. But I haven't done any testing of the idea so I don't know exactly what will break if we go that route.

ggordon:
@lgebhardt Unfortunately that won't work for me. I have some legacy namespaced models and the resource that wraps it is in a different namespace. This was all working before. I think I might need to create a mapping like `@@resource_types`, but mapping models to resources, or maybe just overrides when there are differences. (How did I miss #513 it's exactly the same issue?)

lgebhardt:
Would it be possible to create a model in the resource's namespace that is derived from the legacy model?

barelyknown:
[...]

Issue #513, opened 2015-10-23
Namespaced resources


garytaylor:
Since the upgrade to 0.6.1, my test suite for my has broken and it was working for 0.6.0 - the reason is an internal server error when doing most operations and I have debugged it down to line 506 of resource.rb - well, this is where the error starts to happen.

The error is :-

```
Internal Server Error: JSONAPI: Could not find resource 'finance_manager/finance_manager/bank_account'. (Class FinanceManager::FinanceManager::BankAccountResource not found) /Users/garytaylor/.rvm/gems/ruby-2.2.2/gems/jsonapi-resources-0.6.1/lib/jsonapi/resource.rb:281:in `resource_for'
/Users/garytaylor/.rvm/gems/ruby-2.2.2/gems/jsonapi-resources-0.6.1/lib/jsonapi/resource.rb:506:in `block in find'
/Users/garytaylor/.rvm/gems/ruby-2.2.2/gems/activerecord-4.2.4/lib/active_record/relation/delegation.rb:46:in `each'
/Users/garytaylor/.rvm/gems/ruby-2.2.2/gems/activerecord-4.2.4/lib/active_record/relation/delegation.rb:46:in `each'
/Users/garytaylor/.rvm/gems/ruby-2.2.2/gems/jsonapi-resources-0.6.1/lib/jsonapi/resource.rb:505:in `find'
/Users/garytaylor/.rvm/gems/ruby-2.2.2/gems/jsonapi-resources-0.6.1/lib/jsonapi/operation.rb:45:in `apply'
/Users/garytaylor/.rvm/gems/ruby-2.2.2/gems/jsonapi-resources-0.6.1/lib/jsonapi/active_record_operations_processor.rb:23:in `block in process_operation'
/Users/garytaylor/.rvm/gems/ruby-2.2.2/gems/jsonapi-resources-0.6.1/lib/jsonapi/operations_processor.rb:90:in `with_default_handling'
/Users/garytaylor/.rvm/gems/ruby-2.2.2/gems/jsonapi-resources-0.6.1/lib/jsonapi/active_record_operations_processor.rb:21:in `process_operation'
/Users/garytaylor/.rvm/gems/ruby-2.2.2/gems/jsonapi-resources-0.6.1/lib/jsonapi/operations_processor.rb:53:in `block (5 levels) in process'
[...]

PR #528, opened 2015-11-11
Handle namespaced models

ggordon:
Fix for #523 (possible fix for #513)

Resource lookup was failing unless the model namespace was the same as
the resource namespace, excluding the module path. So while it worked
for the resource, `/api/v2/posts` having a model `Post`. It didn't
reliably work for a model `Legacy::Post`.

Overriding `resource_for_type` did solve part of the problem, but when
the mismatched resource was a relationship, `module_path` was wrong
and the resource could not be found.

I tried creating `/api/v2/legacy/posts` as a copy of `/api/v2/posts` and
this got me further, but still there were some cases were the resource
could not be found. (So far I am unable to generalize these cases, as
I could see no differences between these and the ones where this
technique worked.)

I had to resort to storing the `model` and `path` in the
`@@resource_types` class variable. And use that information in the few
`resource_for` calls that tried to lookup by model.

### DIFF
--- a/lib/jsonapi/resource.rb
+++ b/lib/jsonapi/resource.rb
@@ -293,13 +293,37 @@ module JSONAPI
         check_reserved_resource_name(base._type, base.name)
       end
 
-      def resource_for(type)
-        resource_name = JSONAPI::Resource._resource_name_from_type(type)
-        resource = resource_name.safe_constantize if resource_name
-        if resource.nil?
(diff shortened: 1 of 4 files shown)


nlsrchtr:
Works for me very good. Thanks for your work! :+1:

adamworrall:
This fixed the issue for me, as well.

lgebhardt:
@ggordon Thanks for tackling this!

PR #529, opened 2015-11-16
Make params parsing behave more like Rails

reidab:
Specifically, ensure that the parsed params end up as a HashWithIndifferentAccess instead of a vanilla Hash.

### DIFF
--- a/lib/jsonapi/mime_types.rb
+++ b/lib/jsonapi/mime_types.rb
@@ -5,5 +5,7 @@ end
 Mime::Type.register JSONAPI::MEDIA_TYPE, :api_json
 
 ActionDispatch::ParamsParser::DEFAULT_PARSERS[Mime::Type.lookup(JSONAPI::MEDIA_TYPE)] = lambda do |body|
-  JSON.parse(body)
+  data = JSON.parse(body)
+  data = {:_json => data} unless data.is_a?(Hash)
+  data.with_indifferent_access
 end


lib/jsonapi/mime_types.rb
```diff
@@ -5,5 +5,7 @@ module JSONAPI
 Mime::Type.register JSONAPI::MEDIA_TYPE, :api_json
 
 ActionDispatch::ParamsParser::DEFAULT_PARSERS[Mime::Type.lookup(JSONAPI::MEDIA_TYPE)] = lambda do |body|
-  JSON.parse(body)
+  data = JSON.parse(body)
+  data = {:_json => data} unless data.is_a?(Hash)
```
lgebhardt, line 9: This isn't ever reached in the current test suite, and I think `JSON.parse` will raise an error if it can't convert to a hash. It seems like we would also want to raise an error if we ever get back a result that isn't a Hash. Is there something I'm missing here?

reidab, line 9: While the JSON API spec requires the top-level element to be a hash, JSON itself is fine with top-level arrays. I pulled the `{:_json => data}` wrapping from the Rails implementation of this parser, with the expectation that existing code further down the pipeline would raise `ParameterMissing` or something along those lines. I avoided raising the exception here because it wouldn't  have the advantage of being rescued and packed up in a nice JSON response. 

The root issue is that JSON.parse returns a vanilla `Hash` instead of a `HashWithIndifferentAccess`. This has the side effect of making `request.parameters` also wind up as a vanilla `Hash`, because Rails builds it by merging the query parameters into the result of this parsing.

lgebhardt, line 9: Thanks, that makes sense.

PR #532, opened 2015-11-18
fix bug with server_error_callbacks being called from subclasses

barelyknown:
Wow, this one took me a while to figure out.

There were a couple of recent changes made to `ActsAsResourceController`.
One of the changes switched away from the `ActiveSupport::Concern` `class_methods`
approach, and one of them refactored the way that the `server_error_callbacks`
were stored. The problem is that the class instance variable wasn't visible
to subclasses and the test suite missed it because it tested controller classes
that had the `on_server_error` block set directly on them. I fixed that problem
and completed the switch back to the non-`ActiveSupport::Concern` approach.

This is a bug that might impact a number of apps in a way that isn't clear if those apps
weren't testing that the `on_server_error` method/block was actually being called.

### DIFF
--- a/lib/jsonapi/acts_as_resource_controller.rb
+++ b/lib/jsonapi/acts_as_resource_controller.rb
@@ -2,10 +2,11 @@ require 'csv'
 
 module JSONAPI
   module ActsAsResourceController
-    extend ActiveSupport::Concern
 
-    included do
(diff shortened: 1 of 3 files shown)


lgebhardt:
@barelyknown Thanks!

PR #533, opened 2015-11-18
update dasherized key formatter to handle camel cased input

barelyknown:
The dasherized key formatter wasn't correctly converting from camel case
into the dasherized format. This was causing problems with polymorphic
has many relationships.


lgebhardt:
Thanks!

PR #534, opened 2015-11-18
Move fetchable_fields override to class level

reidab:
Just a small proposal to make the `*_fields` overrides more consistent.

Adds a class-level `self.fetchable_fields(context)` method to Resource and changes the existing `Resource#fetchable_fields` to call the class-level version with the current context. Updates docs to encourage overwriting `self.fetchable_fields` to be parallel with the workflows for `updatable_fields`, `creatable_fields`, and `sortable_fields`.

This change should be backwards-compatible with existing instance-level overrides of `fetchable_fields`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,7 +187,7 @@ end
 ##### Fetchable Attributes
 
 By default all attributes are assumed to be fetchable. The list of fetchable attributes can be filtered by overriding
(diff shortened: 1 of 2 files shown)


lgebhardt:
@reidab Thanks, that does make things more consistent.

PR #535, opened 2015-11-18
Readme had examples of attribute with multiple attributes

gr-eg:


lgebhardt:
Thanks! Guess we need some tests for the readme :smiley:

Issue #526, opened 2015-11-06
Custom filter with callable


iamvery:
# What?

Currently in order to define a custom filter, you must override `self.apply_filter` on the resource and switch over the `filter` being applied.

``` ruby
def self.apply_filter(records, filter, value, options)
  case filter
  when :custom
    records.custom(value)
  else
    super(records, filter, value, options)
  end
end
```

This style of overriding methods for custom functionality is admittedly proliferate in this lib, but I might find it more idiomatic (in a Ruby and Ruby on Rails sense) to specify a callable at the point the filter is defined to state the strategy for filtering.
# How?

The interface I imagine would be similar to ActiveRecord's `scope` class method. Here are a few options:

As a second argument...

``` ruby
filter :custom, ->(records, value) { records.custom(value) }
```

As an option...

``` ruby
filter :custom, with: -> (records, value) { records.custom(value) }
```

As a block...

``` ruby
filter :custom { |records, value| records.custom(value) }
```
# Why?

There are a number of reasons this strikes me as an improvement over the current style.
1. The filtering strategy is defined at the same time as the filter itself. They're "close" visually in context.
2. It takes considerable less lines to express the behavior.
3. It removes the need to switch over filter types.
4. It (possibly) removes the need entirely for overriding the `apply_filter` method.

Do you think this would be a useful addition the library?
If so, which style would be preferred?

lgebhardt:
@iamvery Yes, I think this would be a very nice addition to the library. I like the simplicity of the first, but the second option would be preferable as it leaves more room for expansion without changing the method signature. One expansion might be to have both an `apply` and a `verify` lambda for each filter. That could get rid of the `verify_custom_filter` and `verify_relationship_filter` methods (or at least allow us to deprecate them).

iamvery:
I've opened a PR for this #537. I haven't used the filter verification myself, so I may need help understanding how/why one might do that. I've found a couple cases in the test suite, so I'll try to take a closer look.

I'm closing this issue in favor of the PR. We can continue discussion there :+1:

PR #537, opened 2015-11-19
Custom filter lambda

iamvery:
Here's an implementation following #526.

In brief, this allows for the following syntax instead of overriding `apply_filter` on a resource.

``` ruby
filter :custom, with: -> (records, value) { records.custom(value) }
```

Based on the feedback from that issue, there are a couple things which may still need to be addressed.
- [x] Consider renaming the `:with` option to `:apply`
- [x] Consider adding an additional option for a `:verify` lambda

### DIFF
--- a/README.md
+++ b/README.md
@@ -616,7 +616,7 @@ This callable will be used to apply that filter.
 This example shows how you can implement different approaches for different filters.
 
 ```ruby
-filter :visibility, with: ->(records, value, _) {
+filter :visibility, apply: ->(records, value, _) {
   records.where('users.publicly_visible = ?', value == :public)
 }
 
--- a/README.md
+++ b/README.md
@@ -610,16 +610,18 @@ end
 
 ###### Applying Filters
 
-The `apply_filter` method is called to apply each filter to the `Arel` relation. You may override this method to gain
-control over how the filters are applied to the `Arel` relation.
+You may customize how a filter behaves by suppling a callable to the `:with` option.
+This callable will be used to apply that filter.
 
 This example shows how you can implement different approaches for different filters.
 
 ```ruby
+filter :visibility, with: ->(records, value, _) {
+  records.where('users.publicly_visible = ?', value == :public)
+}
+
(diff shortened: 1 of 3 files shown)


lgebhardt:
@iamvery I like your two suggestions. I think adding the `:verify` lambda makes sense, and if we have that then renaming the `:with` lambda to `:apply`seems most consistent.

DavidMikeSimon:
:+1: I have a very similar feature implemented in a Concern for my resources, it results in more readable code than the standard `apply_filter` way

iamvery:
Alright, I've updated the PR, addressing the two above tasks. I'm not sure how I feel about the `:verify` option. I think that's somewhat due to having not even known about it before. I noticed it's not covered in the readme. The interface for that method (and specific return value) makes it feel internal to me.

What do you think of the code here?

lgebhardt:
@iamvery Thanks for doing this. I'm not sure how I feel about the `:verify` option either. I don't like the format of the specific return, and it seems like we should allow for `verify` on relationship filters as well. I'll see if I can refactor the underlying code into something better and maybe the new filter option will then be cleaner.

lgebhardt:
I added the filter_verify branch (https://github.com/cerebris/jsonapi-resources/tree/filter_verify) with some changes that I think simplify the verify option a bit and will still be backwards compatible. Let me know what you think.

iamvery:
Looks good! I think I'll take a closer look and try rebasing those changes into the PR for continuity. Does that sound good? (might be a bit before I can roll that in)

lgebhardt:
@iamvery I could just create a PR which would pull both sets of changes in if you are good with that. That should be the easiest solution.

iamvery:
I added a few comments to commits on your branch. Hopefully I'm not overstepping by taking on this code review style. 😯

lgebhardt:
> Hopefully I'm not overstepping by taking on this code review style.

Not at all. In fact I appreciate the comments and extra review.

iamvery:
Should I go ahead and close this in favor of #555?

PR #541, opened 2015-11-19
Test case for JR::Resource.resource_for_model_path

fernandes:
The problem occurs when we have two Resources with the same
`model_name`, when we try to find the resource for a model, it returns
the first Resource that was loaded using `resource_for`, not the correct
one.

### DIFF
--- a/test/fixtures/active_record.rb
+++ b/test/fixtures/active_record.rb
@@ -15,6 +15,7 @@ ActiveRecord::Schema.define do
     t.belongs_to :preferences
     t.integer    :hair_cut_id, index: true
     t.boolean    :book_admin, default: false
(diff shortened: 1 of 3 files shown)


ggordon:
I think this is related to the STI patch. The `find` method now attempts to guess the resource based on the model class, previously it just used `self`. But as your example shows given a model it's impossible to _guess_ the correct resource.

fernandes:
@ggordon I totally agree, as it's working today its kind of "one way" algorithm heheh

about this problem, I discussed with @lgebhardt on gitter and sent this PR just to exemplify the problem with a test, it was introduced on v0.6.2, I don't know the internals to try a fix, but I'm available in what I can help

thank you

PR #544, opened 2015-11-21
Fix typo in prod env check

iamvery:
- The ActiveSupport::StringInquirer is just a fancy string comparison,
  so the strings have to match :laughing:. The correct environment name
  is "production".

h/t @xdmx https://github.com/cerebris/jsonapi-resources/pull/524#issuecomment-158616665


lgebhardt:
Thanks

PR #549, opened 2015-11-23
Include options in apply_sort

johantell:
Including options will allow to create sorting based on the actual context. It's especially useful when working with scopes that need request based parameters. In my case position or the current_user

### DIFF
--- a/lib/jsonapi/resource.rb
+++ b/lib/jsonapi/resource.rb
@@ -486,7 +486,7 @@ module JSONAPI
         records
       end
 
-      def apply_sort(records, order_options, _options = {})
+      def apply_sort(records, order_options, _context = {})
         if order_options.any?
           records.order(order_options)
         else
@@ -528,8 +528,8 @@ module JSONAPI
         apply_includes(records, options)
       end
 
-      def sort_records(records, order_options, options = {})
-        apply_sort(records, order_options, options)
+      def sort_records(records, order_options, context = {})
+        apply_sort(records, order_options, context)
       end
 
       def find_count(filters, options = {})
@@ -841,7 +841,7 @@ module JSONAPI
               sort_criteria =  options.fetch(:sort_criteria, {})
               unless sort_criteria.nil? || sort_criteria.empty?
                 order_options = relationship.resource_klass.construct_order_options(sort_criteria)
-                records = resource_klass.apply_sort(records, order_options, context: @context)
+                records = resource_klass.apply_sort(records, order_options, @context)
               end
 
               paginator = options[:paginator]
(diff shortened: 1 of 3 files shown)


xdmx:
:+1:

lib/jsonapi/resource.rb
```diff
@@ -486,7 +486,7 @@ def apply_pagination(records, paginator, order_options)
         records
       end
 
-      def apply_sort(records, order_options)
+      def apply_sort(records, order_options, options = {})
```
lgebhardt: Could you rename `options` to `_options` for style consistency as it's unused in this method?

johantell: certainly!

lgebhardt:
One more change discovered in #561. We should pass the context in as an option on line 864 (https://github.com/cerebris/jsonapi-resources/blob/master/lib/jsonapi/resource.rb#L864) so it's available when we are sorting on related records.

``` ruby
records = resource_klass.apply_sort(records, order_options, context: @context)
```

johantell:
Should we consider adding some information to the readme?

lgebhardt:
@mintcore I think adding some information to the readme would be good. I also wonder if we should pass the options provided to relationship_name, but with the context overridden. Or maybe just make `apply_sort` take context instead of options. I think I may like this better as the set of options won't be consistent between the two calls. Thoughts?

johantell:
I agree about the context part!

lib/jsonapi/resource.rb
```diff
@@ -544,7 +544,7 @@ def find(filters, options = {})
 
         sort_criteria = options.fetch(:sort_criteria) { [] }
         order_options = construct_order_options(sort_criteria)
-        records = sort_records(records, order_options)
+        records = sort_records(records, order_options, options)
```
lgebhardt: `options` => `context`

test/unit/resource/resource_test.rb
```diff
@@ -240,7 +240,7 @@ def test_to_many_relationship_sorts
 
     # define apply_filters method on post resource to not respect filters
     PostResource.instance_eval do
-      def apply_sort(records, criteria)
+      def apply_sort(records, criteria, options = {})
```
lgebhardt: `options` => `context`

test/unit/resource/resource_test.rb
```diff
@@ -253,7 +253,7 @@ def apply_sort(records, criteria)
   ensure
     # reset method to original implementation
     PostResource.instance_eval do
-      def apply_sort(records, criteria)
+      def apply_sort(records, criteria, options = {})
```
lgebhardt: `options` => `context`

lgebhardt:
@mintcore Thanks!

PR #550, opened 2015-11-25
remove unneeded check which introduces a bug

acid:
There is a verify_type call a few lines lower which does exactly what we
need without causing strange errors when you are using type members with
underscores in patch or pull requests.


lgebhardt:
@acid Thanks!

Issue #551, opened 2015-11-26
Rails autoloading (development code reload) doesn't work for resources


SimonKaluza:
Resources instantiated and cached for retrieval by `JSONAPI::Resource::resource_for` (such as any used by subclasses of `JSONAPI::ResourceController`) will not be reloaded by Rails autoloading when they are modified, as the `@@resource_types` class variable holds on to the stale Resource subclass object.

SimonKaluza:
https://github.com/cerebris/jsonapi-resources/pull/552 implements a naive fix to this, which basically re-instantiates the resource each time `Resource::resource_for` is called in the Rails development environment (only using the caching when `Rails::configuration.cache_classes` is true).

chamnap:
:+1:

PR #552, opened 2015-11-26
Fix Rails autoloading for Resources (code reload in development)

SimonKaluza:
This is a simplistic attempt at fixing https://github.com/cerebris/jsonapi-resources/issues/551 by disabling the utilization of resource_types cache when cache_classes is false.

Though this forces recreation of resources on each call to resource_for, this fixes Rails autoloading to actually reload Resource code while in development (previously `@@resource_types` held on to the stale version of the class from before the autoload).

A more efficient solution would only reinstantiate the resources after the Rails `set_clear_dependencies_hook` clears `ActiveSupport::Dependencies`, but would likely require more complex callback/finalizer code.

### DIFF
--- a/lib/jsonapi/resource.rb
+++ b/lib/jsonapi/resource.rb
@@ -293,7 +293,7 @@ module JSONAPI
       end
 
       def resource_for(resource_path)
-        unless @@resource_types.key? resource_path
+        unless @@resource_types.key?(resource_path) && Rails.configuration.cache_classes
           klass_name = "#{resource_path.to_s.underscore.singularize}_resource".camelize
           klass = (klass_name.safe_constantize or
             fail NameError,


meagar:
:+1: [Forked](https://github.com/meagar/jsonapi-resources/commits/master) and tested it out, appears to work both with a vanilla Rails project and with our project which is built ion [`rails-api`](https://github.com/rails-api/rails-api). Would love to see this merged, it's a relatively major blocker to us adopting jsonapi-resources. Without it, I have to kill and restart my Rails server after each change made to app/resource/*.rb

bobwhitelock:
+1 also, we've been using this for a few days and seems to fix the issue. Thanks

lgebhardt:
In #554 I ended up reworking the resource resolution logic and removed `@@resouce_types`. I'm holding off merging this PR as it should prove to be superfluous.

lgebhardt:
Closing because #554 should accomplish the same thing. Feel free to reopen if that isn't the case.

PR #554, opened 2015-11-26
Model hints for namespaced models

### LINKED ISSUE #513: Namespaced resources

### LINKED ISSUE #523: Namespaced model not working after STI fix added

### LINKED ISSUE #525: model_name and class_name options are not respected
As a result of #507, `Resource.model_name` and the `class_name` option of `Resource.has_many` are no longer respected.

For instance, requests for authors in this example will still be serialized using `UserResource` and so include `email` in their attributes hash.

``` ruby
  attributes :first_name, :email  
end

class AuthorResource < JSONAPI::Resource
  model_name 'User'
  attributes :first_name
end
```

Likewise, the authors in this example will be serialized using `UserResource` and so include `email` in their attributes hashes as well.

``` ruby
class Book < ActiveRecord::Base
  has_many :authors, class_name: 'User'
end

  has_many :authors, class_name: 'Author'
end
```

There are a few changes I think caused this, but there might be more:
- https://github.com/cerebris/jsonapi-resources/pull/507/files#diff-e6cf31c573bd5f2277b1f16812943df4L505
- https://github.com/cerebris/jsonapi-resources/pull/507/files#diff-e6cf31c573bd5f2277b1f16812943df4L517
- https://github.com/cerebris/jsonapi-resources/pull/507/files#diff-e6cf31c573bd5f2277b1f16812943df4L814

lgebhardt:
This PR may not be ready to merge and at this point is for discussion and testing only. I am open to other ideas on how to solve this issue, while still maintaining backwards compatibility.

This reverts the logic for namespaced models introduced in 0.6.2. This logic created a mapping between models and resources, but this mapping could not be relied on 100% of the time, as #541 identified. So in it's place a `model_hint` is used. This also maps models to resources, however these hints are added in the resource definition. In addition when the `model_name` is defined (say for a namespaced model) a hint is added (though this can be skipped with an option).

I am hoping this will fix #525 and #541while not breaking existing code, which fixed #523 (and possibly fixed #513).

Goals for this PR were:
- maintain backwards compatibility
  - particularly _not_ requiring namespaced models
- require a minimum of code to support namespaced models
  - work with existing `model_name` methodology
- However support custom mappings if needed

### DIFF
--- a/README.md
+++ b/README.md
@@ -365,7 +365,7 @@ end
 
 Resource instances are created from model records. The determination of the correct resource type is performed using a
 simple rule based on the model's name. The name is used to find a resource in the same module (as the originating 
-resource) that matches the name. This usually works quite well, however it can fail when model name's do not match
+resource) that matches the name. This usually works quite well, however it can fail when model names do not match
 resource names. It can also fail when using namespaced models. In this case a `model_hint` can be created to map model
 names to resources. For example:
 
--- a/README.md
+++ b/README.md
@@ -361,6 +361,39 @@ class AuthorResource < JSONAPI::Resource
 end
 ```
 
+When the `model_name` is called a corresponding `model_hint` is also added. This can be skipped by using the
+`model_hint` option set to false. For example:
+
+```ruby
+class AuthorResource < JSONAPI::Resource
+  model_name 'Legacy::Person', model_hint: false
+  model_hint model: 'Legacy::Person', resource: 'person'
+end
+```
+
+#### Model Hints
+
+Resource instances are created from model records. The determination of the correct resource type is performed using a
+simple rule based on the model's name. The name is used to find a resource in the same module (as the originating 
+resource) that matches the name. This usually works quite well, however it can fail when model name's do not match
+resource names. It can also fail when using namespaced models. In this case a `model_hint` can be created to map model
+names to resources. For example:
+
+```ruby
+class AuthorResource < JSONAPI::Resource
+  attribute :name
+  model_name 'Person'
+  model_hint model: Commenter, resource: :special_person
+
+  has_many :posts
+  has_many :commenters
+end
+```
+
+Model hints inherit from parent resources, but are not global in scope. The `model_hint` method requires a `model` and
+a `resource` named parameter. `model` takes an ActiveRecord class or class name, and `resource` takes a resource type or
+a resource class.
+
 #### Relationships
 
 Related resources need to be specified in the resource. These may be declared with the `relationship` or the `has_one`
--- a/README.md
+++ b/README.md
@@ -361,16 +361,6 @@ class AuthorResource < JSONAPI::Resource
 end
 ```
 
-When the `model_name` is called a corresponding `model_hint` is also added. This can be skipped by using the
-`model_hint` option set to false. For example:
-
-```ruby
-class AuthorResource < JSONAPI::Resource
-  model_name 'Legacy::Person', model_hint: false
-  model_hint model: 'Legacy::Person', resource: 'person'
-end
-```
-
 #### Model Hints
 
 Resource instances are created from model records. The determination of the correct resource type is performed using a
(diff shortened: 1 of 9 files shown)


lgebhardt:
Well looks like required named parameters are not ruby 2.0 compatible. Opinions on whether to require ruby 2.1 or to add a default to the `model_hint` method are also welcome.

fernandes:
I'm worried about resources giving the "hint" to the same models, I'm gonna make some tests here, but seems the only solution to this "lookup" problem

fernandes:
I added a second model, with a `lazy` attribute (as shown on this [gist](https://gist.github.com/fernandes/fb15dc527d8724c398b4)), and lookup worked fine... solved the problem

:+1: for this solution

lgebhardt:
In testing this @xtian found an issue (reported in #525) with the `type`. In fixing that I found a major error in how relationships were working in derived resources. I'd rather these were split into tow PRs, but fixing one broke tests for the other. And I'm not sure I want to take the time to separate them at this point.

fernandes:
great!

README.md
```diff
@@ -361,6 +361,38 @@ class AuthorResource < JSONAPI::Resource
 end
 ```
 
+#### Model Hints
+
+Resource instances are created from model records. The determination of the correct resource type is performed using a
+simple rule based on the model's name. The name is used to find a resource in the same module (as the originating 
+resource) that matches the name. This usually works quite well, however it can fail when model name's do not match
```
dgeb: name's -> names

fernandes:
:tada: :smile:

PR #555, opened 2015-11-30
Custom Filter Lambdas - rework verify

lgebhardt:
Reworks verify option in #537

### DIFF
--- a/README.md
+++ b/README.md
@@ -616,7 +616,7 @@ This callable will be used to apply that filter.
 This example shows how you can implement different approaches for different filters.
 
 ```ruby
-filter :visibility, with: ->(records, value, _) {
+filter :visibility, apply: ->(records, value, _) {
   records.where('users.publicly_visible = ?', value == :public)
 }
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -509,11 +510,65 @@ end
...
+
+This example shows how you can implement different approaches for different filters.
+
+```ruby
+filter :visibility, apply: ->(records, value, _options) {
+  records.where('users.publicly_visible = ?', value == :public)
+}
+```
+
+If you omit the `apply` callable the filter will applied as `records.where(filter => value)`.
```
dgeb: will _be_ applied

README.md
```diff
@@ -509,11 +510,65 @@ end
...
+      end
+    else
+      return super(records, filter, value)
+  end
+end
+```
+
+##### Verifying Filters
+
+Because filters are provided by the users of the API your code is given an opportunity to verify the value. You may
```
dgeb: The wording's a little awkward. How about:

> Because filters typically come straight from the request, it's prudent to verify their values. To do so, provide a callable to the `verify` option.

PR #559, opened 2015-12-02
Validation errors can have metadata added to them

benarmston:
The JSON API specification supports error objects having metadata added to them.  This PR adds support for metadata on validation errors.

A resource can now optionally define a `validation_error_metadata` method which can return any metadata to merge into the validation error.

Suppose `model_error_messages` returned the following error messages hash:

```
{password: ["too_short", "format"]}
```

Then to add data to the validation error `validation_error_metadata` could return:

```
{
  password: {
    "too_short": {"minimum_length" => 6},
    "format": {"requirement" => "must contain letters and numbers"}
  }
}
```

The specified metadata is then be merged into the validation error object.

An alternate API I considered was to have a `validation_error_metadata_for(attr_key, message)` method on `Resource` which would return the metadata for a single validation error object.

### DIFF
--- a/lib/jsonapi/exceptions.rb
+++ b/lib/jsonapi/exceptions.rb
@@ -301,10 +301,11 @@ module JSONAPI
     end
 
     class ValidationErrors < Error
-      attr_reader :error_messages, :resource_relationships
+      attr_reader :error_messages, :error_metadata, :resource_relationships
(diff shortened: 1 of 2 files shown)


lgebhardt:
@benarmston Thanks!

Issue #437, opened 2015-09-09
Expose JSONAPI::ResourceSerializer#attribute_hash ?


danielma:
Thanks for all the work on this gem. I've been using it for a new project and I really enjoy it.

For a project I'm currently working on, I have a `Song` model that `has_one :embed`. I don't want to expose the `Embed` model as part of the REST API. I always want to include the attributes of the embed in the response.

The [JSON API format document](http://jsonapi.org/format/#document-resource-object-attributes) explicitly states that says that attributes _"may contain any valid JSON value"_ and that _"complex data structures involving JSON objects and arrays are allowed as attribute values."_

I propose that models specified as attributes in a resource should be serialized simply, using `#attribute_hash`.

For example, this is what my JSON looks like if I simply add `:embed` as an attribute to my `SongResource`

``` json
"embed": {
  "id": 1,
  "identifier": "1101",
  "site": "youtube",
  "art_content_id": 1,
  "art_content_type": "Song",
  "created_at": "2015-09-04T19:55:41.596Z",
  "updated_at": "2015-09-04T19:55:41.596Z"
}
```

If it was passed through the serializer, this is what I would get. (this is what I want)

``` json
"embed": {
  "identifier": "1101",
  "site": "youtube"
}
```

Right now, I can do this by manually with code like this in my `SongResource`

``` ruby
def embed
  resource = JSONAPI::Resource.resource_for @model.embed.class.to_s.underscore
  serializer = JSONAPI::ResourceSerializer.new resource
  serializer.send :attribute_hash, resource.new(@model.embed)
end
```

If you would be willing to consider a feature like this, I would be happy to put together a pull request for it

lgebhardt:
> I propose that models specified as attributes in a resource should be serialized simply, using #attribute_hash.

Do you mean "resources specified as attributes" and not "models"? The ResourceSerializer will only work on resources.

Another way to solve this would be to put a flag on a relationship that indicates it should be treated as an attribute instead. That way you could take advantage of all the options on the relationship to resolve the underlying model relation.

How would to_many relationships be handled? Returning a full array of attribute_hashes could get very large since I assume they won't be paginated. Maybe that's not something we need to worry about.

All in all it seems like a useful feature.

PR #564, opened 2015-12-07
Add support for nesting attributes in jsonapi resources.

mfunaro:
This pull request addresses the issue of nesting additional data in attributes within a resource file, opened here #437. 

We came across a use case where we wanted to nest data within an attribute of the data object. I was able to implement the functionality following the approach in #437, however that solution is very manual and might have to be done in more than one resource depending on the complexity of your API. 

My changes now allow a user to specify data they wish to be nested in a clean way like the following 

``` Ruby
class ParentResource < JSONAPI::Resource
  attribute :name
  attribute :children, format: :nested_attribute, with: ChildResource
end
```

Which will produce the following

``` Ruby
{
  data: {
    type: 'parents',
    id: '1',
    links: {
      self: '/parents/1'
    },
    attributes: {
      name: 'John Smith',
      children: [
        {
          name: 'John Smith Jr.',
          age: 2
        },
        {
          name: 'Jane Smith',
          age: 5
        }
      ]
  }
}
```

Just a note, the test use model relationships to test the functionality for simplicity sake, direct relationships however should adhere to JSON api standard. The changes I am proposing would apply to situations where the nested data does not directly relate to the primary object or other 'special circumstance' situations where adhering strictly to the JSON api spec might not be possible.

### DIFF
--- a/lib/jsonapi/formatter.rb
+++ b/lib/jsonapi/formatter.rb
@@ -1,7 +1,7 @@
 module JSONAPI
   class Formatter
     class << self
-      def format(arg)
+      def format(arg, options = {})
         arg.to_s
       end
 
@@ -42,8 +42,8 @@ module JSONAPI
 
   class ValueFormatter < Formatter
     class << self
-      def format(raw_value)
-        super(raw_value)
+      def format(raw_value, options = {})
+        super(raw_value, options)
       end
 
       def unformat(value)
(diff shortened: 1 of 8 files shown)


mfunaro:
Muddied up the branch while working, I will reopen later with a clean branch if someone thinks this would be useful.

acid:
Yeah, please do!

jamesy829:
@mfunaro I think this addition is awesome, any progress made for the clean branch?

SvenSchoene:
FWIW: this would've been really cool!

PR #565, opened 2015-12-07
correctly set model_name of subclasses of non-abstract resources

barelyknown:
The `_model_name` of the superclass was being carried over its subclass (and overriding its `_model_name`) if the superclass wasn't abstract (as is the case with STI classes).


lgebhardt:
Thanks!

PR #566, opened 2015-12-09
Move Routing section and add to README ToC

reidab:
Promote the Routing section of the README to an H3, move it out of (what looks like) the middle of the Serializer section, and add it to the ToC.


lgebhardt:
Thanks @reidab!

PR #568, opened 2015-12-10
Revert "Make fetchable_fields more like other field overrides"

reidab:
This reverts my earlier accepted PR #534. :disappointed: 

I now realize that a class-level implementation of `fetchable_fields` loses access to the requested `@model` object, making checks like "is the requested model owned by the current user" impossible.

Example from my `UserResource`:

``` ruby
def fetchable_fields
  filtered_fields = [:password]

  current_user = context[:current_user]
  unless current_user && (current_user.admin? || current_user == @model)
    filtered_fields << :email
  end

  super - filtered_fields
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,7 +188,7 @@ end
 ##### Fetchable Attributes
 
 By default all attributes are assumed to be fetchable. The list of fetchable attributes can be filtered by overriding
-the `self.fetchable_fields` method.
+the `fetchable_fields` method.
 
 Here's an example that prevents guest users from seeing the `email` field:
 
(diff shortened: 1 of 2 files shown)


cdunn:
Just ran into this same issue :+1:

lgebhardt:
Thanks for fixing this @reidab . Wish I had caught it earlier.

Exelord:
@lgebhardt If this solution has been accepted, shouldn't we also change `updatable_fields` and `creatable_fields` to keep consistency and a possibility to use `@model` also in there?

Just asking...

lgebhardt:
@Exelord Yes, in the ideal world that would be the case. However where the methods are used we don't have a resource instance instantiated. It might be possible, but it would be a significant refactor. I'll look into it in the refactor I'm working on for multiple operation support since that's changing a lot of the request parsing logic.

PR #571, opened 2015-12-14
Rebased / updated @AKHarris' "custom links defined per resource"

reidab:
A quick attempt to resurrect @AKHarris' PR #486 :smile: 

This branch rebases the changes from #486 onto the current master, fixing merge conflicts, and updates the `custom_link` lambda to receive `ResourceSerializer#custom_generation_options`, per [this comment](https://github.com/cerebris/jsonapi-resources/pull/486#issuecomment-152297638).

### DIFF
--- a/test/unit/resource/resource_test.rb
+++ b/test/unit/resource/resource_test.rb
@@ -83,56 +83,35 @@ end
 class CustomLinkTestResource < JSONAPI::Resource
   model_name 'Post'
 
-  custom_link :raw, :self
+  custom_link :raw, ->(source, link_builder) { link_builder.self_link(source) + "/raw" }
 end
 
 class CustomLinkTestWithExtensionResource < JSONAPI::Resource
   model_name 'Post'
 
-  custom_link :raw, :self, ext: :xml
+  custom_link :raw, ->(source, link_builder) { link_builder.self_link(source) + "/raw.xml" }
 end
 
 class CustomLinkTestWithCustomPathResource < JSONAPI::Resource
   model_name 'Post'
 
-  custom_link :raw, :self, path: "super/duper/custom/path", ext: :xml
+  custom_link :raw, ->(source, link_builder) do
+    link_builder.self_link(source) + "/super/duper/path.xml"
+  end
 end
 
 class CustomLinkLambda< JSONAPI::Resource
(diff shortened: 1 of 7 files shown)


akharris:
:+1: 

@reidab thanks! Sorry I've been MIA.

reidab:
Another option here would be to make this more closely mirror the resource meta feature, by removing the `custom_link` DSL and just providing a `custom_links(options)` method that people can override to provide a hash that will be merged with the auto-generated links hash.

e.g.

``` ruby
  def custom_links(options)
    {
      raw: options[:serialzer].link_builder.self_link(self) + "/raw",
      related: {
        href: options[:serialzer].link_builder.self_link(self) + "/related",
        meta: {
          count: 10
        }
      }
    }
  end
end
```

akharris:
@reidab I started initially by implementing `custom_link` as a DSL but I've (fairly lately) become a big fan of using primitive data types by default. I really like Hashes, and the example above seems is clear and elegant to me. It also encourages me to look at the `link_builder` object to understand `#self_link` and any other methods that may be useful to me (and encouraging users of a library to look at source code is a plus imo).

And then maybe the DSL should emerge from use cases (perhaps the `raw` becomes super common and should thus have some shorthand for building it).

These are just a few of my immediate thoughts.

reidab:
Okay, I took a stab at simplifying things down to an override of `custom_links` that mirrors the `meta` implementation. If we like this approach, I'd also like to put together a PR implementing `custom_relationships` in the same way.

test/unit/resource/resource_test.rb
```diff
@@ -80,6 +80,40 @@ class RelatedResource < MyModule::RelatedResource
   end
 end
 
+class CustomLinkTestResource < JSONAPI::Resource
```
lgebhardt: These resources are not being used in the tests. Could you please remove?

You can run `COVERAGE=true bundle exec rake test` to make sure you're at 100% coverage.

akharris:
ping @reidab. take it away :clap:

reidab:
:+1: Removed unused resources, verified 100% coverage, and rebased onto current master.

lgebhardt:
Thanks @reidab and @AKHarris!

PR #577, opened 2015-12-30
Add i18n to easily customize errors messages

marceloboth:


lgebhardt:
@marceloboth Thanks! This is a nice addition.

iamvery:
We should have the default exception message generated or included to keep compatibility when upgrading.

iamvery:
The provided translations don't seem to load in an app.

(sorry for the previous, mistaken message about them not being attached)

iamvery:
I suspect we need the file to be copied as a part of a generator task or similar. Perhaps similar to how [devise does it](https://github.com/plataformatec/devise/blob/7df57d5081f9884849ca15e4fde179ef164a575f/lib/generators/devise/install_generator.rb#L17).

lgebhardt:
@iamvery A generator would be good, though I'd like to fix it with the default file loading too.

PR #584, opened 2016-01-07
Simplify `process_primary`.

beauby:
The following check:

``` ruby
if already_serialized?(resource.class._type, id)
  set_primary(@primary_class_name, id)
end
```

being already done inside `add_included_object`, the `process_primary` method can be simplified as suggested.
This is mainly a cosmetic change though.


lgebhardt:
@beauby Thanks!

PR #595, opened 2016-01-25
Inherit ResourceController from ActionController::Metal

thibaudgg:
Cherry pick `ActionController` modules that are only required makes `ResourceController` leaner and faster.

On Rails 5, we could also inherit directly from `ActionController::API`: https://github.com/rails/rails/blob/master/actionpack/lib/action_controller/api.rb

### DIFF
--- a/lib/jsonapi/resource_controller.rb
+++ b/lib/jsonapi/resource_controller.rb
@@ -1,5 +1,17 @@
 module JSONAPI
-  class ResourceController < ActionController::Base
-    include JSONAPI::ActsAsResourceController
+  class ResourceController < ActionController::Metal
+    MODULES = [
+      AbstractController::Rendering,
+      ActionController::Rendering,
+      ActionController::Renderers::All,
+      ActionController::StrongParameters,
+      ActionController::ForceSSL,
+      ActionController::Instrumentation,
+      JSONAPI::ActsAsResourceController
+    ].freeze
+
+    MODULES.each do |mod|
+      include mod
+    end
   end
 end


lgebhardt:
@thibaudgg Thanks!

thibaudgg:
@lgebhardt you're welcome, happy to seeing that one merged!

iamvery:
I ran into this excluding the `head` method on controllers. That affects the doorkeeper gem (and probably others). I'm not intimately familiar with the details, but I _think_ you would need to include [`ActionController::Head`](api.rubyonrails.org/classes/ActionController/Head.html) as a fix. You're probably more familiar with the implications of that and why you may or may not want to do so.

iamvery:
Just hit `ActionController::Rescue` as well.

thibaudgg:
@iamvery the idea of this PR was to only use the bare minimum to make JR specs pass and let every user add what the need. For example we added that in our app:

``` ruby
module Api
  module V1
    class ApiController < JSONAPI::ResourceController
      MODULES = [
        Doorkeeper::Helpers::Controller,
        Doorkeeper::Rails::Helpers,
        ErrorRenderer,
        Rails.application.routes.url_helpers
      ].freeze

      MODULES.each do |mod|
        include mod
      end
    end
  end
end
``
```

iamvery:
That makes sense. It's just that the change wasn't well documented and causes breakage in existing apps. How can we make the upgrade a little smoother for existing users?

lgebhardt:
I was afraid this might break existing apps. That was one of the drivers for moving to a beta release.

At the very least I'd like to get this better documented. But maybe it would be better to keep `ResourceController` as it was and implement the slimmed down controller as it's own class. That way it won't break existing apps, and people making the switch will expect a few potential bumps, which will be documented. Thoughts?

iamvery:
That came to mind as well. If you really want to encourage folks to build on the bare controller, you could introduce a deprecation warning when using the full `ResourceController`.

Also the beta release worked well for us. It's the exact reason I'm here thinking about this :smile:

john-griffin:
This change also broke our app, we are also using doorkeeper. This feels like a premature optimization.

PR #597, opened 2016-01-27
underscorize resource type name in links

galiminus:
Hello!

I came across a few issues when I tried to use polymorphic relationships. In my case the type field in the relationship object was something like "PurchaseRequisitions" instead of "purchase_requisitions".


lgebhardt:
@phorque Thanks!

PR #598, opened 2016-01-27
Remove duplicate check for params.nil?

scottgonzalez:
This function returns immediately if `params.nil?`, so the `unless params.nil?` check isn't necessary.


lgebhardt:
@scottgonzalez Good catch, thanks!

PR #599, opened 2016-01-28
Convert error code representation from integers to strings

### LINKED ISSUE #582: Error code should be a string
First at all, thanks for the nice gem! It does extremally huge work! And really easy to use! 

I've discovered that errors codes are returned as strings, but according to [JSONAPI specification](http://jsonapi.org/format/#errors), as I understood they should be strings.

> code: an application-specific error code, expressed as a string value.

 E.g. instead of `100` should be returned `"100"`. 

Do I understoond it right?

Thanks in advance!

beechnut:
Convert error code representation from integers to strings, to be compliant with [the standard](http://jsonapi.org/format/#error-objects) (see the bullet `status`). Also, set tests to expect strings instead of integers.

> `status`: the HTTP status code applicable to this problem, **expressed as a string** value.

Fixes #582.


beechnut:
I just noticed my last commit inserted a lot of changes re: newlines. I have my editor set up to strip trailing whitespace off the ends of lines. I can re-commit if necessary, just let me know if it's an issue.

lgebhardt:
@beechnut Thanks!

Issue #16, opened 2014-08-31
Authorization


Samsinite:
Hey guys,

This gem/library looks great! What about directly supporting an option to easily hook up authorization instead of requiring the resource controller index/show/create/update/destroy methods to be overridden?

Something like a simple interface that filters for index/read_list, returns true/false for show/create/update/destroy, and if in the future updating lists or destroying lists are supported, the interface could be updated to support filtering those as well. Than it would be straight forward for whoever to override it to support their preferred method of authorization, such as pundit or cancancan. I wouldn't mind implementing this if you guys are interested.

lgebhardt:
I have been thinking about different ways to handle authorization. I stripped out my first attempt (a set of callbacks of the resource) since I wasn't sure it was quite right.

One concern I have about overriding the controller to handle authorization is that once we get patch support implemented it will be possible to have a single request create multiple operations. Authorization will need to be checked for each of these operations. I don't know how well Pundit or cancancan can be adapted to this criteria.

One method I have been playing with (in my project using this gem) has been to create a custom OperationsProcessor, derived from the `ActiveRecordOperationsProcessor`. In `before_operation` I call a custom method on each resource which I have named `approve_operation`. This allows each resource to have it's own authorization scheme. I'm not totally happy with this approach since the `approve_operation` method isn't as clean as I think it should be (due to needing to pull the details out of the `Operation` object). 

I would certainly appreciate any work you wish to put into this problem. I just want to make sure any solution we adopt will support multiple operations, is flexible to support different authorization packages, and is resource centric (if possible).

Samsinite:
lgebhardt,

I'll put together an example showing the gist of what I was thinking, and see what you think.

I'm thinking an interface that can be specified in the controller -- similar to how context is specified -- so nothing on the controller has to be overridden. I don't see this getting in the way for a patch single request that handles multiple operations, as whatever handles the type of patch operation that occurs can know the appropriate type of authorization that it needs to do. I currently think authorization should occur in the controller and not in the resource -- thus why I think a callback on a resource is a bad idea.

lgebhardt:
@Samsinite, thanks, I look forward to seeing your proposal.

lgebhardt:
From the http://jsonapi.org/format/#patch-urls

[...]

PR #602, opened 2016-01-29
Return a relation instead of a class from Resource.records

lime:
When filter, sorting etc, the result is usually an `ActiveRecord::Relation`. However, the default `Resource.records` returns `_model_class` which is a `Class`. This caused some confusion in at least one comment: https://github.com/cerebris/jsonapi-resources/issues/16#issuecomment-137519516

Calling `.all` on the class returns a relation, making for more consistent behaviour.


lgebhardt:
@lime Thanks!

PR #603, opened 2016-02-02
Add default values into the error messages if the locale doesn't exist in the project.

lgebhardt:
Rename the keys for the exception translations to include jsonapi-resources.

Fixes the issue @iamvery raised in #577. It would still be nice to get a generator for the locale, but that can be in a separate PR.

### DIFF
--- a/lib/jsonapi/exceptions.rb
+++ b/lib/jsonapi/exceptions.rb
@@ -18,8 +18,10 @@ module JSONAPI
 
         [JSONAPI::Error.new(code: JSONAPI::INTERNAL_SERVER_ERROR,
                             status: :internal_server_error,
-                            title: I18n.translate('exceptions.internal_server_error.title'),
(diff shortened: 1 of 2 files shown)


iamvery:
Looks good! It would be nice if there was a way to reduce the duplication of the `:default` values and the translation file included in the app. Any idea how that might be pulled off?

lgebhardt:
Thanks for testing.

I didn't see a simple way to reduce the duplication, but I also didn't spend a lot of time looking. We're planning some major changes to the `Request` object and the `OperationsProcessor` code which will require generators. I'm thinking it will make sense to add the locales file into new projects as part of the generator for the app. But if you have any ideas on how to cleanly handle it I would certainly be open to implementing them.

iamvery:
The generator seems like a decent solution. I probably wouldn't expend energy on this if you know that's coming.

I'm quite interested in seeing the changes you have in plan for the lib.

PR #606, opened 2016-02-05
Add gem version badge to README

reidab:


lgebhardt:
@reidab Good idea, thanks!

PR #609, opened 2016-02-11
Rails 5 compat: use version-relevant param parser

olleolleolle:
Many users have suggested this fix, but this PR is against the `beta` branch.

### DIFF
--- a/lib/jsonapi/mime_types.rb
+++ b/lib/jsonapi/mime_types.rb
@@ -4,7 +4,9 @@ end
 
 Mime::Type.register JSONAPI::MEDIA_TYPE, :api_json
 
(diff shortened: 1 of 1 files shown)


olleolleolle:
Ah! Am I supposed to work against the `rails5` branch, instead? (I found patches related to Rails-compatibility in this branch, so I assumed things.)

lgebhardt:
@olleolleolle, Yes, the `rails5` branch should have fixed this. If that's not the case please reopen this and I'll take a look at this PR.

olleolleolle:
@lgebhardt There was a global namespace leakage in that file, when last I looked, and this is just a rephrasing of that, which does not leak.

PR #615, opened 2016-02-15
Make offset and paged paginators links consistent

coryodaniel:
Per the pagination spec (http://jsonapi.org/format/#fetching-pagination)
previous links should be 'prev'.  It was done this way in the paged
paginator but not the offset.

I left the original key name in there for backwards compat and updated the tests to expect an additional key. LMK if you want me to add a deprecation message and notes around the tests where the link count has changed.

### DIFF
--- a/lib/jsonapi/paginator.rb
+++ b/lib/jsonapi/paginator.rb
@@ -58,10 +58,12 @@ class OffsetPaginator < JSONAPI::Paginator
 
       previous_offset = 0 if previous_offset < 0
 
-      links_page_params['previous'] = {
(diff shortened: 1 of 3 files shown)


lgebhardt:
@coryodaniel Thanks. I think we should just remove support for `previous` and stick with `prev` to stay compliant with the spec. No sense preserving an incompatible (with the spec) way of working. This will go into a beta release and I'll call out the change in the release notes.

coryodaniel:
Updated the PR tests and class to not include 'previous'

lgebhardt:
@coryodaniel Thanks!

PR #616, opened 2016-02-15
Add create_operations_processor override example

cball:
I struggled a bit finding where to override this (thanks @barelyknown for pointing me in the right direction). I'm not sure if this section belongs here or below after ResourceControllers are introduced, but thought it would be a good addition to the README.


lgebhardt:
@cball Thanks!

PR #622, opened 2016-02-17
Meta: page_count

### LINKED ISSUE #126: Pagination: total record/page count in meta

utilityboy:
Added option for top level `page_count` meta for paged paginators in addition to the record_count.

A number of client-side paging libraries look for APIs to return a total page count.

Potential closes #126?


utilityboy:
@lgebhardt or @barelyknown is there anything else I can do here?

lgebhardt:
@utilityboy Sorry for the very late response. I swear I commented about this back in March, but I probably forgot to hit the comment button. Anyway, it's missing tests that cover the page count in the find operation. You can see coverage with:

``` bash
COVERAGE=true bundle exec rake test
```

In addition the README needs to be updated for the configuration options.

utilityboy:
No problem @lgebhardt.  I'll get on it this weekend.  Thanks for the update.

utilityboy:
All set @lgebhardt.  Let me know if there's anything else.

lgebhardt:
@utilityboy Thanks! Should be quite useful.

PR #624, opened 2016-02-19
Allow Hash values for filter params

reidab:
e.g. `?filter[search][name]=fred&filter[search][facet]=orange`

Rails handles this nicely ,but JR was blowing up up when trying to CSV-parse the Hash prior to filter validation.


lgebhardt:
@reidab Thanks!

PR #632, opened 2016-02-24
Flatten and check exception class whitelisting with a method

vijoc:
I just spent an embarrassingly long time chasing down the reason why my exceptions were not passed through, after adding them to the configuration as `config.exception_class_whitelist << [MyError]`. This, of course, caused a nested Array to be stored and the checks failed to account for that.

This PR extracts the exception class whitelist checking to a new method and adds a `flatten` to the chain to avoid just these kinds of situations.


lgebhardt:
@vijoc Thanks!

PR #637, opened 2016-02-26
fix resource identifier objects

acid:
According to
http://jsonapi.org/format/#document-resource-identifier-objects
rescource identifier objects can contain an additional meta attribute.
This fix adds supports for the full specification.

### DIFF
--- a/lib/jsonapi/request_parser.rb
+++ b/lib/jsonapi/request_parser.rb
@@ -381,22 +381,20 @@ module JSONAPI
 
     def parse_to_one_links_object(raw)
       if raw.nil?
-        return {
-          type: nil,
-          id: nil
-        }
+        return link_object_hash(nil, nil)
       end
 
-      if !raw.is_a?(Hash) || raw.is_a?(ActionController::Parameters)) ||
-        (raw.length != 2 && !raw.key?('meta')) || !(raw.key?('type') && raw.key?('id'))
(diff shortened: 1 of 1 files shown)


lib/jsonapi/request.rb
```diff
@@ -382,13 +382,14 @@ def parse_to_one_links_object(raw)
...
-      if !raw.is_a?(Hash) || raw.length != 2 || !(raw.key?('type') && raw.key?('id'))
+      if !raw.is_a?(Hash) || (raw.length != 2 && !raw.key?('meta')) || !(raw.key?('type') && raw.key?('id'))
         fail JSONAPI::Exceptions::InvalidLinksObject.new
       end
 
       {
         type: unformat_key(raw['type']).to_s,
-        id: raw['id']
+        id: raw['id'],
+        meta: raw['meta']
```
beauby: This does add a (possibly `nil`) meta attribute to each resource identifier object, right?
Maybe the following would make more sense:

``` ruby
hash = {
  type: unformat_key(raw['type']).to_s,
  id: raw['id']
}

hash[:meta] = raw[:meta] if raw.key?(:meta)

hash
```

acid: I like that, thanks! :)

lgebhardt:
@acid Looks like the meta will get dropped in `parse_to_many_links_object`, though I haven't tested it.

acid:
@lgebhardt you're right. Fixed it, thanks!

PR #639, opened 2016-02-28
Improve filters doc

miguelsan:
Make the wording in the corresponding section of the README a bit more precise.

### DIFF
--- a/README.md
+++ b/README.md
@@ -525,13 +525,12 @@ class ContactResource < JSONAPI::Resource
 end
 ```
 
-Then a request could pass in a filter for example `http://example.com/contacts?filter[name_last]=Smith` and the system
-will find all people where the last name exactly matches Smith.
+Then a request could pass in a filter like for example `http://example.com/contacts?filter[name_last]=Smith` and the system
+would find all people whose last name exactly matches Smith.
 
 ##### Default Filters
 
-A default filter may be defined for a resource using the `default` option on the `filter` method. This default is used
-unless the request overrides this value.
+A default filter with a default value may be defined for a resource using the `default` option on the `filter` method.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -525,13 +525,12 @@ class ContactResource < JSONAPI::Resource
 end
 ```
 
-Then a request could pass in a filter for example `http://example.com/contacts?filter[name_last]=Smith` and the system
-will find all people where the last name exactly matches Smith.
+Then a request could pass in a filter like for example `http://example.com/contacts?filter[name_last]=Smith` and the system
```
lgebhardt: How about "Then a request could pass in a filter, for example" ?

miguelsan: How about the new wording?

utilityboy:
Hi @miguelsan and @lgebhardt .  I've mentioned this elsewhere, and @lgebhardt I know we discussed briefly about eventual deprecation of the wiki, but I did a pretty extensive writeup here on the topic of querystring parameters:

https://github.com/cerebris/jsonapi-resources/wiki/JSONAPI::Resources-Querystring-Examples

I'm happy to bring this into the README, or at least make links out to the wiki.  Let me know if I can help at all or if you think this is useful.  I'd be very happy to work up a section on filtering.

lgebhardt:
@utilityboy Thanks for bringing that back up. The fact that I forgot about your page in the Wiki says to me it's not the best place to keep documentation, at least not without callouts from the Readme. If you're up for it I'd say we should move it into the Readme, with a possibility that the readme will eventually get split into a set of doc pages in the project.

utilityboy:
I'll do this this weekend and also add examples for filtering.

PR #640, opened 2016-02-29
[Fixes #289] Add 'Location' header if single resource created

### LINKED ISSUE #289: Add location to response?
Right now this unit test doesn't pass:

```
      it "returns 201 with corresponding location" do
        request.env['CONTENT_TYPE'] = 'application/vnd.api+json'
        post :create, valid_attributes, valid_session
        expect(response.status).to eq(201)
        expect(response_body['data']['links']['self']).to eq(v1_concept_url(Concept.last)) # this passes
        expect(response.location).to eq(v1_concept_url(Concept.last)) # not this
      end
```

It would appear the response location header is not set?

```
     Failure/Error: expect(response.location).to eq(v1_concept_url(Concept.last))

       expected: "http://test.host/v1/concepts/730"
            got: nil

       (compared using ==)
```

The JSONAPI spec [does seem to require it](http://jsonapi.org/format/#crud-creating-responses-201), though.

vrybas:
Adds `Location` response header, equal to `self` link, if user
successfully creates single resource. If multiple resources are created,
the `Location` header stays `nil`.

Fixes issue https://github.com/cerebris/jsonapi-resources/issues/289

From specification:

> **201 Created**
> ...
> The response SHOULD include a Location header identifying the location
> of the newly created resource.
> ...
> If the resource object returned by the response contains a self key in
> its links member and a Location header is provided, the value of the
> self member MUST match the value of the Location header.

http://jsonapi.org/format/#crud-creating-responses-201

### DIFF
--- a/lib/jsonapi/acts_as_resource_controller.rb
+++ b/lib/jsonapi/acts_as_resource_controller.rb
@@ -152,7 +152,17 @@ module JSONAPI
 
     def render_results(operation_results)
       response_doc = create_response_document(operation_results)
-      render status: response_doc.status, json: response_doc.contents
+
+      render_options = {
+        status: response_doc.status,
+        json:   response_doc.contents
+      }
+
+      render_options[:location] = response_doc.contents[:data]["links"][:self] if (
+        response_doc.status == :created && response_doc.contents[:data].class != Array
+      )
+
+      render(render_options)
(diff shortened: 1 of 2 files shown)


lgebhardt:
@vrybas Thanks!

PR #642, opened 2016-02-29
Rails 5 beta 3-compatible parameter parsing

### LINKED ISSUE #641: Rails5.0beta3 breakage: ActionDispatch::Request.parameter_parser does not get our parameter parser
[mime_types.rb](https://github.com/cerebris/jsonapi-resources/blob/rails5/lib/jsonapi/mime_types.rb) works with 5.0beta2, but breaks in beta3. 

We used to change DEFAULT_PARSER in-place. 

With beta3, the ActionDispatch::Request.parameter_parser does not get our newly-freedom-patched-in symbol in that hash.

jesjos:
Fixes #641

### DIFF
--- a/lib/jsonapi/mime_types.rb
+++ b/lib/jsonapi/mime_types.rb
@@ -1,15 +1,28 @@
 module JSONAPI
   MEDIA_TYPE = 'application/vnd.api+json'
-end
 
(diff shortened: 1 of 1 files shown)


olleolleolle:
(This also fixes the "leaking a variable name to the global namespace".)

lgebhardt:
@jesjos Thanks!

PR #649, opened 2016-03-08
Sentence that seemd to be about file name convention said 'class name'

tasandberg:
Seems you're talking about the file name here, though it said "class name".


lgebhardt:
@Timmehs Thanks!

PR #651, opened 2016-03-09
Allow values of attributes to be delegated. This allows the json api …

jedld:
…resource attribute to

easily have a different name without having to define a method for it.


lgebhardt:
@jedld, Thanks!

Issue #656, opened 2016-03-17
Clear relationship by sending null data with PUT doesn't work


Zooip:
In JSON-API specs, it is written that the following request must clear my relationship

```
PATCH /articles/1/relationships/author HTTP/1.1
Content-Type: application/vnd.api+json
Accept: application/vnd.api+json

{
  "data": null
}
```

When I try, i get a 500 error :

```
"exception": "undefined method `[]' for nil:NilClass",
"backtrace": [
          "**/jsonapi-resources-0.7.0/lib/jsonapi/request.rb:589:in `parse_update_relationship_operation'",
          ....
  ]
```

On the other way, when I send an update request with a non-existing id, server responds with 204 No Content and the relationship is set to nil. I would have expected a 422 error but the JSON-API doesn't specifies anything on this.

Did I get something wrong ?

Zooip:
After some tests, error seams to appear only with my polymorphic associations.
https://github.com/cerebris/jsonapi-resources/blob/master/lib/jsonapi/request.rb#L587

It doesn't verify presence of `verified_params[:to_one].values[0]`

PR #657, opened 2016-03-17
Accept nil data when updating polymorphic to_one relationship 

Zooip:
Related to #656
Ensure that `verified_params[:to_one].values[0]` is present before calling `[:id]` and `[:type]`

This way the following request clear the polymorphic relationship in the same way as a non-polymorphic relationship.

```
PATCH /articles/1/relationships/author HTTP/1.1
Accept: application/vnd.api+json

{
  "data": null
}
```

Wasn't able to create a test case using defined test model since there is no to_one retaltionship.


lgebhardt:
@Zooip Thanks!

PR #659, opened 2016-03-17
Documentation of model using current_user from context hash

danielpclark:
It is likely that people won't understand how to use the context hash as I didn't understand it at first.  It would be good to have an example so I've written one for the README.

``` ruby
class PeopleResource < JSONAPI::Resource
  before_save do
    @model.user_id = context[:current_user].id if @model.new_record?
  end
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,6 +179,14 @@ end
 class PeopleController < ApplicationController
 
 end
+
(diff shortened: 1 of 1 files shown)


lgebhardt:
@danielpclark Thanks!

PR #665, opened 2016-04-05
Add ability to specify validation context

lessless:
In some cases, users want to run API-specific validations. This commit allows
to pass a validation context to the JSONAPI::Resource#_super method.

### DIFF
--- a/lib/jsonapi/resource.rb
+++ b/lib/jsonapi/resource.rb
@@ -182,8 +182,8 @@ module JSONAPI
     #   return :accepted
     # end
     # ```
-    def _save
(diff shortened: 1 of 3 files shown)


lgebhardt:
@lessless Thanks! I'm a little concerned that this is going to break existing code that uses POROs, but it does make it more consistent with ActiveRecord so that's probably worth it. We just need to call it out in the release notes.

lessless:
Thank you for the help, Larry! Unfortunately, my english not good enough to write documentation and I'm afraid developers will get wrong impression about the project if they see my wording in the README.

PR #679, opened 2016-04-16
Extract IncludeDirectives.

beauby:
Extract `IncludeDirectives` into the [jsonapi_parser](http://github.com/beauby/jsonapi_parser) gem, in an attempt to share work between `ActiveModelSerializers` and `jsonapi-resources` (c.f. https://github.com/rails-api/active_model_serializers/pull/1685).

c.f. https://github.com/beauby/jsonapi_parser/blob/master/lib/json/api/include_directive.rb


beauby:
Note that Travis is failing because of dependency resolution failing for rails 4.1 (`jsonapi_parser` currently requires `activesupport` ~> 4.2, but this is not a strong requirement at all, I'll change it if there is interest in this PR).

beauby:
A bit of context here: 
- The proposed `IncludeDirective` implementation has an underlying structure of a Hash. example:

``` ruby
{
  posts: {
    author: {},
    comments: {
      author: {}
    }
  }
}
```

whereas the current implementation has a somewhat similar structure, with the addition of an `included` flag for each relationship (which in practice is always true), and the sub-relationships are stored under an `:include_related` key.
- The proposed implementation has the additional benefit of handling wildcards (`*` for "all relationships", `**` for "all relationships and their relationships, and so on recursively"). Those are not part of the spec, and are therefore optional (using the `:allow_wildcard` option to the `IncludeDirective` constructor – not enabled in this PR). There currently are discussions whether the spec should allow _some kind_ of wildcard, and whichever choice is made would be easily added in the proposed implementation.

lgebhardt:
@beauby Good idea. I'm looking at the parsing logic for requests again (in anticipation of support for  Operations in the spec), so this is good timing. We should coordinate on the features.

The `included` flag is a remnant from before the spec was finalized, and can probably be removed (I doubt anyone is using it).

beauby:
Rebased.

PR #684, opened 2016-04-21
bug fix removing the assumption that primary key and sti column are defaults

will3216:
I couldn't get the tests to run, but I don't have much time at the moment. This should solve any issues with people using non-standard inheritance_columns or non-standard primary keys, though I didn't have time to verify the latter.

If you pass me some info on test setup, I'll try and revisit this later today when I have some time to add relevant tests and whatnot. Also, let me know if you have any issues with how I bypassed the issue, I honestly did not put much thought into it, just needed to get myself unblocked >.<


lgebhardt:
@will3216 Thanks for the fix. However as #685 points out there's a security hole in accessing the model. So this section of code will need a more drastic change. So I'm holding off on the merge (and the rebase it now needs).

PR #694, opened 2016-05-03
Operation processing

### LINKED ISSUE #675: Error Objects are invalid according to the JSON API Schema
I'm not sure if this is normal or not but, currently some (all?) errors are invalid according to the [JSON API Schema](http://jsonapi.org/schema).

An example error generated in a test:

``` json
{
  "errors": [
    {
      "title": "Missing Parameter",
      "detail": "The required parameter, data, is missing.",
      "id": null,
      "href": null,
      "code": "106",
      "source": null,
      "links": null,
      "status": "400",
      "meta": null
    }
  ]
}
```

Some issues according to the [spec](http://jsonapi.org/format):
- `id` must be a string
- `href` is not a valid member of an error object
- `source`, `links` and `meta` must be an object 

I'm using `json-schema` gem in the tests but the same results can be observed using another tool; e.g.  http://jsonschemalint.com/draft4/# by pasting the contents of http://jsonapi.org/schema into the left side ("JSON Schema") and the error JSON into the right side.

lgebhardt:
This PR is making some changes to get ready for an extension to JSON-API this is more operation centric. While this is a breaking change to how the system processes operations, it should be transparent for projects which have not created custom operations processors.
- Replaces the `OperationProcessor` with `Processor`s, which can be specific for each resource type. A default processor can also be set in the configuration if a resource specific one is not found.
- should fix #675, with null values in the error objects
- Removes Ruby 2.0 support
- Refactors `Request` into `RequestParser`
- Operation classes are simplified into a single `Operation` class, which contains a type and an options hash


README.md
```diff
@@ -1416,6 +1381,56 @@ class UsersController < JSONAPI::ResourceController
...
+
+Operation Processors are called to perform the operation(s) that make up a request. The controller (through the `OperationDispatcher`), creates an `OperatorProcessor` to handle each operation. The processor is created based on the resource name, including the namespace. If a processor does not exist for a resource (namespace matters) the default operation processor is used instead. The default processor can be changed by a configuration setting.
+
+Defining a custom `Processor` allows for custom callback handling of each operation type for each resource type. For example:
+
+```ruby
+class Api::V4::BookProcessor < JSONAPI::Processor
+  after_find do
+    unless @results.is_a?(JSONAPI::ErrorsOperationResult)
+      @result.meta[:total_records_found] = @result.record_count
```
dgeb, line 1395: Should `@result` be `@results` here? (actually `@results` above should be `@result`)

On a related note, we should document what instance variables and callbacks are available within a Processor.

lgebhardt, line 1395: Fixed.

Agreed, we should get that documented.

lib/jsonapi/processor.rb
```diff
@@ -0,0 +1,264 @@
...
+      def _processor_from_resource_type(resource_klass)
+        processor = resource_klass.name.gsub(/Resource$/,'Processor').safe_constantize
+        if processor.nil?
+          processor = JSONAPI.configuration.default_processor_klass
+        end
+
+        processor
+      end
+
+      def transactional(operation_type)
```
dgeb: perhaps `transactional?` or even `transactional_operation_type?` would be a more clear name?

lib/jsonapi/processor.rb
```diff
@@ -0,0 +1,264 @@
...
+        page_options[:record_count] = resource_klass.find_count(verified_filters,
+                                                                context: context,
+                                                                include_directives: include_directives)
+      end
+
+      if JSONAPI.configuration.top_level_links_include_pagination && paginator
+        page_options[:pagination_params] = paginator.links_page_params(page_options)
+      end
+
+      JSONAPI::ResourcesOperationResult.new(:ok,
```
dgeb: for complex methods which return a value, I prefer to explicitly use `return` to clarify that the returned value is significant

lib/jsonapi/processor.rb
lgebhardt: I'm kind of in favor of `transactional_operation_type?`, though it's a bit long.

dgeb:
@lgebhardt Just a few nits, but this is looking really good.

We should make sure we capture the need to document instance variables available in a `Processor` in an issue, if we don't do so in this PR.

@barelyknown et al - please review this PR as well - it's an important change to the way we prepare and process operations.

lib/jsonapi/processor.rb
dgeb: I think that's ok, since it's at the class level. `transactional?` would probably work better on an instance.

lib/jsonapi/processor.rb
lgebhardt: I agree, though I get the feeling the greater ruby community doesn't. We should probably standardize on returns everywhere, or only for early returns. Just using them some places would indicate to me that only those places are significant.

dgeb: Here's some pseudo-significant poll data supporting explicit returns https://twitter.com/_jayphelps/status/727046905264381952

We don't have to make the change now (or ever), but just throwing that out there.

lgebhardt:
@dgeb Thanks. Fixed the nits. I'll do the documentation of the instance variables in a bit.

lgebhardt:
Rebased to bring in the `page_count` feature (#622).

PR #695, opened 2016-05-03
Load type based on resource_klass's type and not the class_name.

### LINKED ISSUE #689: "type" reference in Relationship
Hi there,

I'm probably missing something, so I would like to know if there is any particular reason to grab the relation's type by:

`@type = class_name.underscore.pluralize.to_sym`

Instead of something like this:

`@type = resource_klass._type.to_sym`

IMHO it would make more sense to grab it from `resource_klass` than `class_name` in order to avoid incoherences such as:

Having this value for `type` at `relationship` member:

screen shot 2016-04-26 at 3 42 29 pm

And this one at `included` member:

screen shot 2016-04-26 at 3 45 29 pm

Thoughts?

lgebhardt:
fixes #689

### DIFF
--- a/lib/jsonapi/relationship.rb
+++ b/lib/jsonapi/relationship.rb
@@ -1,6 +1,6 @@
 module JSONAPI
   class Relationship
-    attr_reader :acts_as_set, :foreign_key, :type, :options, :name,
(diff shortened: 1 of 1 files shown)


tiagopog:
👍

PR #700, opened 2016-05-05
Ensure media type of Accept header is valid according to the JSON API spec

### LINKED ISSUE #699: ensure_correct_media_type does not validate against an invalid Accept header
I noticed that the `ensure_correct_media_type` `before_action` does not validate against an invalid `Accept` header. It appears to only validate against the `Content-Type` header.

jerelmiller:
Closes #699 

Ensure media type of Accept header is valid according to the JSON API spec.
- Adds before_action to validate accept header
- Add tests to cover all use cases
- Respond with 406 status if invalid
- Include appropriate error classes and translations


lgebhardt:
@jerelmiller Thanks! Do we need tests for "*/*" and others that are not acceptable, e.g. "text/plain"?

``` ruby
  def test_accept_header_all
    @request.headers['Accept'] = "*/*"

    get :index
    assert_response :success
  end

  def test_accept_header_not_jsonapi
    @request.headers['Accept'] = "text/plain"

    get :index
    assert_response 406
    assert_equal 'Not acceptable', json_response['errors'][0]['title']
    assert_equal "All requests must use the '#{JSONAPI::MEDIA_TYPE}' Accept without media type parameters. This request specified '#{@request.headers['Accept']}'.", json_response['errors'][0]['detail']
  end

```

jerelmiller:
Great question. From my understanding of the spec, it should only `406` if the JSON API media type is present and all instances of it has been modified with media type parameters. Here is what it has to say.

> Servers MUST respond with a `406 Not Acceptable` status code if a request’s Accept header contains the JSON API media type and all instances of that media type are modified with media type parameters.

From the way I read that, it seems like it should accept non-JSON API media types in the `Accept` header, as long as the JSON API media type is not there at all.

Is that the way you interpret that as well? I've already interpreted it wrong a couple times so I want to make sure that your understanding if it is the same as mine.

That being said, it wouldn't hurt to get those tests in place for those `Accept` headers. I think they should both be acceptable in the case.

Your thoughts?

**Edit**: I made a typo and said it "would" hurt to get those tests in place, when I really meant it "wouldn't" hurt. I've corrected above. I'll get those test cases in place and ready for you to look at.

lgebhardt:
As I see it the purpose of that requirement is that a client must accept the plain `application/vnd.api+json` response. So the acceptable options would be: 
- none specifice, assume any is OK
- "*/*" - all are OK
- or `application/vnd.api+json` without media type parameters

I imagine a mixed set would also be Ok, for example `text/plain,application/vnd.api+json`. But it's not very clear to me. 

Also I think my example above should have been a `415` and not `406`, based on a reread. 

@dgeb Can you shed some light on this?

jerelmiller:
One of my colleagues did happen to find this not long ago. Maybe this helps?

https://github.com/elliotttf/jsonapi-headers/commit/db85f3215238014ff980cdad07fec7f6071042f2

jerelmiller:
This has definitely been one of the more confusing aspects for me of that spec since its not super clear what exactly it looks like. Too bad the spec doesn't include any examples.

I'll be glad to update test cases and requirements after we have a better understanding of what the spec is trying to say. I wouldn't be surprised if I misinterpreted it for this PR.

jerelmiller:
@lgebhardt I went ahead and added those tests you mentioned above. I will update as we get clarification around this spec.

dgeb:
The purpose of this requirement in the spec is to ensure that clients do not _require_ a modified version of the media type. This gives servers the choice to serve the base version of the media type if that's all they are capable of serving.

@lgebhardt has interpreted it correctly here:

> So the acceptable options would be:
> - none specified, assume any is OK
> - "*/*" - all are OK
> - or application/vnd.api+json without media type parameters
> 
> I imagine a mixed set would also be Ok, for example text/plain,application/vnd.api+json. But it's not very clear to me.

To be clear, yes, a mixed set like `text/plain,application/vnd.api+json` would be fine.

jerelmiller:
Awesome. I will update the test cases and get this implemented shortly. Thanks for the clarification @dgeb. That helps tremendously.

lgebhardt:
Note, I updated my comment and @dgeb's quote of it. The `*` got lost in the markdown, "*/*" was displayed as "_/_". I should have typed it as `"*/*"` instead of using just double quotes.

jerelmiller:
Thank you. I'll be sure I use the right one

jerelmiller:
@lgebhardt I've updated the tests and implementation. Second pass is ready for your review.

lgebhardt:
@jerelmiller Thanks!

timrwood:
With this patch, all of my `get` request specs are returning `406` instead of `200`, as well as any requests from a browser.

> To be clear, yes, a mixed set like `text/plain,application/vnd.api+json` would be fine.

@dgeb, should a mixed set like `accept:text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8` work, as it includes `*/*`? That is the default accept header for chrome on a get request.

sebyx07:
this sucks a lot, would be nice if there would be away to disable it from the initializer

dgeb:
> @dgeb, should a mixed set like accept:text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,_/_;q=0.8 work, as it includes _/_? 

@timrwood sorry to miss your question. Yes an accept header that includes `*/*` should be fine. See https://github.com/cerebris/jsonapi-resources/pull/700#issuecomment-217984449

> this sucks a lot, would be nice if there would be away to disable it from the initializer

@sebyx07 the only thing we're preventing here (to align with the JSON-API spec) is adding media type parameters to the JSON-API media type. Is that what you're objecting to?

lgebhardt:
@sebyx07 While it may suck, claiming so isn't necessary and is probably counter productive to your goals.

The issue is the browser is sending `*/*;q=0.8` and the code is looking to accept plain `*/*`. See https://github.com/cerebris/jsonapi-resources/blob/master/lib/jsonapi/acts_as_resource_controller.rb#L158

PRs are most welcome.

sebyx07:
@dgeb it's pretty hard to debug from the browser. For now I have this fix, which you can put in the base controller

``` ruby
unless Rails.env.production?
  skip_before_filter :ensure_correct_media_type
  skip_before_filter :ensure_valid_accept_media_type
end
```

dgeb:
@sebyx07 see @lgebhardt's comment [above](https://github.com/cerebris/jsonapi-resources/pull/700#issuecomment-239260151) which explains _why_ it's hard to debug from the browser. Also see the proposed fix in #799 which will provide a more compliant solution than simply disabling the filters.

dajacques:
The way I'm reading the spec and this thread, any Accept is ok except something like..

`application/vnd.api+json;foo=bar`

But..

`application/vnd.api+json;foo=bar,application/vnd.api+json`

would be ok because the bare media type is included.

Let me know if I'm missing something. I'm overriding this method in my base controller to only fail in this case. I can submit a PR if I'm not off the mark.

PR #703, opened 2016-05-06
Fix dasherized paths for scoped routes

corthmann:
I've had some issues with scoped (namespaced) resources not getting the proper route formatting. The code in this pull request fixes that issue and adds tests to cover the scenario.

When I wrote the tests I discovered that the test failed "randomly" depending on the order in which they were run. I added a fix for that too, but the larger problem seem to be that there are a bunch of statements  like `JSONAPI.configuration.route_format = :underscored_route` inside `test_helper.rb`. If you agree then I'll gladly make a separete pull request attempting to solve that problem.


lgebhardt:
@corthmann Thanks! If you wish to fix the tests with configuration changes in them that would be great. Most, but not all, of the configuration changes used in the tests have been reset using an `ensure` section. One idea I've had is to create a method to reset the configuration to a default and call that from `ensure`, or maybe even better before every test. As it is now it certainly could be improved.

PR #707, opened 2016-05-10
Make id respect creatable_fields

### LINKED ISSUE #673: creatable_fields & updatable_fields for primary Id
Hi,

I am new to ruby and specially to this gem. After reading on the code and examples I figure out that if I want certain fields (ID in my case) to be excluded from JSON request I can use the following code:

```
class ApiResource < JSONAPI::Resource

  def self.updatable_fields(context)
    super - [:id]
  end

  def self.creatable_fields(context)
    super - [:id]
  end
end
```

`

I inherit all my other resources from this class. So I dont accept the ID attribute for any of the JSON call for POST method.

But its not working and ID is always passed to the ActiveRecord to be save in DB.
How can I achieve this? Or is my knowledge wrong on the api.

My POST Request from POSTMAN:

```
`{"data":
    {
    "id":"5",
    "type":"applications",
    "attributes": {
        "name":"Ehsan",
        "clientAppId":"Client"
        }
    }
}
```

lgebhardt:
Fixes #673


rainerborene:
+1

PR #708, opened 2016-05-10
Don't introduce git dependency in gemspec

mecampbellsoup:
## Description

Introducing `git` as a dependency in the `<gem-name>.gemspec` file introduces a problem for those of us isolating our applications within Docker containers that don't otherwise need `git`.


lgebhardt:
@mecampbellsoup Thanks!

PR #711, opened 2016-05-13
provide fields parameter from request to operation

senid231:
it allows for example to do something like:

``` ruby
class CatResource < JSONAPI::Resource
  attributes :name, :breed

  filters :name

    fields = options.fetch(:fields, {})[:cats]
    coll = super
    fields.nil? ? coll : coll.select(fields.join(','))
  end

end
```

``` ruby
class ExtendedMetaOperationsProcessor < ActiveRecordOperationsProcessor

  after_find_operation do
    @operation_meta[:fields] = @operation.fields
  end

end
```


senid231:
any updates about this?

lgebhardt:
@senid231 Sorry, merging #694 created a few conflicts with this. Could you rework this to work with the new changes?

senid231:
done

lgebhardt:
@senid231 Thanks!

PR #712, opened 2016-05-13
Resource serialization fragment caching

DavidMikeSimon:
This PR is a **request for review**, it works well but I don't think it's clean or well-tested enough to be merged just yet.

I've written a fragment caching feature for jsonapi-resources, which saves individual serialized resources to a cache and, on future requests for the same resource with the same serialization settings, can pull the serialized data back out from the cache without instantiating ActiveRecord or Resource, and (except for the 'relationships' sub-hash) without even re-generating the JSON strings. This provides substantial speed benefits:

```
bench_large_index_request_uncached .  4.180000   0.020000   4.200000 (  4.200144)
bench_large_index_request_caching .  1.280000   0.020000   1.300000 (  1.302931)
```

Caching is enabled on a Resource-by-Resource basis, because some Resources may (for whatever reason) be difficult to keep properly synced with the cache via changes to `updated_at`.

Various implications:
- I switched from using ActiveSupport's JSON to using the JSON gem directly. This adds a 20% or so speed increase to normal operations without caching, and also provides the benefit of allowing JR to cache and use pre-generated JSON substrings (class `JSONAPI::CompiledJson`) within larger JSON objects. There were a number of places where JR was implicitly relying on ActiveSupport's `as_json` behavior, which now must be handled explicitly in ValueFormatter.
- OperationsProcessor now accepts an `options` hash, which is used for now only to tell it about caching settings.
- ResourceSerializer now has a `config_key` method, which must return a different value if serialization settings are different. This is included in cache keys, which means that different representations of a resource are cached separately. If any additional settings are added to ResourceSerializer in the future, they must also be included in the `config_key` method!
- Options for filtering, sorting, and pagination can now be passed to the associated records method generated for relationships (e.g. `records_for_comments`).
- When caching is enabled, it is not safe to override instance-level association methods on Resource classes, because these methods are not called when recovering associations to or through cached resources. However, using a lambda for the association name is still OK.

I'm currently testing the caching behavior with a test helper method called `assert_jsonapi_cacheable_get`, which is used in all request tests that previously just called `get` and expected 200s back. This method issues the get request normally once, then issues it again with caching enabled to make sure that the caching system didn't affect the response, and then issues the request a third time and asserts that the response is still the same and that the cache was hit.

The request tests are a pretty good measure of the caching system, but I also want to do a similar thing for controller tests before I'm confident I've covered the edge cases. Additionally, I need to add tests on when caching is enabled on only a subset of Resource classes.

### DIFF
--- a/lib/jsonapi/request_parser.rb
+++ b/lib/jsonapi/request_parser.rb
@@ -339,6 +339,7 @@ module JSONAPI
         source_klass: @source_klass,
         source_id: @source_id,
         filters: @source_klass.verify_filters(@filters, @context),
+        include_directives: @include_directives,
         sort_criteria: @sort_criteria,
         paginator: @paginator,
         fields: @fields,
--- a/README.md
+++ b/README.md
@@ -51,6 +51,8 @@ backed by ActiveRecord models or by custom objects.
   * [Routing] (#routing)
     * [Nested Routes] (#nested-routes)
   * [Authorization](#authorization)
+  * [Resource Caching] (#resource-caching)
+    * [Caching Caveats] (#caching-caveats)
 * [Configuration] (#configuration)
 * [Contributing] (#contributing)
 * [License] (#license)
@@ -1644,10 +1646,11 @@ class DefaultValueFormatter < JSONAPI::ValueFormatter
   class << self
     def format(raw_value)
       case raw_value
-        when String, Integer
-          return raw_value
+        when Date, Time, DateTime, ActiveSupport::TimeWithZone, BigDecimal
+          # Use the as_json methods added to various base classes by ActiveSupport
+          return raw_value.as_json
         else
-          return raw_value.to_s
+          return raw_value
       end
     end
   end
@@ -1680,16 +1683,14 @@ resource for a system wide change).
 and
 
 ```ruby
-class MyDefaultValueFormatter < JSONAPI::ValueFormatter
+class MyDefaultValueFormatter < DefaultValueFormatter
   class << self
     def format(raw_value)
       case raw_value
-        when String, Integer
-          return raw_value
         when DateTime
-          return raw_value.in_time_zone('UTC').to_s
+          return super(raw_value.in_time_zone('UTC'))
         else
-          return raw_value.to_s
+          return super
       end
     end
   end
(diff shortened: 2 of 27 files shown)


DavidMikeSimon:
Hm, a test is failing in Rails 5. I'll check that out Monday, but at a glace I think it's just a symbol keys vs string keys thing.

DavidMikeSimon:
Note to self: commit 55a329e7 should be removed

DavidMikeSimon:
Ok, got the Rails 5 issue fixed. It had to do with the test helper monkey patches related to time precision; I was able to remove entirely those due the new serialization method, and tests pass assuming millisecond precision available in all environments (even Rails 4.0).

lgebhardt:
Thanks, I like the direction. A few recent PRs have created some conflicts which we'll need to get resolved. The biggest is probably the changes to the operations with the addition of processors.

DavidMikeSimon:
@lgebhardt Rebased on latest master, working on some additional tests now.

DavidMikeSimon:
Argh, hold on, my rebase accidentally cut off the functionality. :-\ Fixing...

DavidMikeSimon:
Ok, caching is working for reals again, and I added some controller tests, but the test suite isn't complete just yet.

DavidMikeSimon:
Ok, added a few tests which reveal some issues in how my caching code deals with self-referential includes, will look into it on Monday.

DavidMikeSimon:
I'm seeing failures on Travis for tests that run fine on my local machine. :-\ Investigating...

lgebhardt:
@DavidMikeSimon It's almost certainly order dependent (I thought I got all of them, but probably not). Grab the seed from the travis test that failed and run the test suite with that seed locally to reproduce the errors. I can also take a look shortly.

DavidMikeSimon:
@lgebhardt Nah, it's the same tests failing in all Travis environments, and the tests pass on my system when I run the full suite with the same seed. :-(

lgebhardt:
Looks to all be related to `warmup action made too many queries`. I haven't seen that before. I wonder if the travis environment is an issue.

DavidMikeSimon:
Yeah, that's part of an additional test helper I added and am using throughout controller_test: `assert_cacheable_get`. It runs the query normally, then runs it again with various caching configurations and makes sure that (a) the cached responses match the normal response and (b) the cached responses don't use absurd numbers of queries. I'm not sure why (b) would differ between environments, but it must be! Maybe a subtle version difference in a dependency.

DavidMikeSimon:
I ran the local Travis docker image and still couldn't reproduce the failures. :-(

DavidMikeSimon:
Found the problem, my master was slightly out of date. Rebased and solved the issue, tests should pass now...

DavidMikeSimon:
@lgebhardt Okay, I think this is ready to merge. A few additional points of discussion beyond what I mentioned in the first comment:
- I refactored RequestParser. I ran into a chicken-and-egg problem when constructing ResourceSerializer early, because the serializer needs the inclusions and fields from the request, but the request needs the serializer if caching is enabled. Anyways, RequestParser is now about 80 lines shorter, and generates the operations on demand when you call the `operations` method, rather than when RequestParser is instantiated. Also, it now raises exceptions on parsing errors rather than adding to an `errors` array. To handle situations where multiple errors need to be returned, I created a new error class  `AggregatedError`, which has convenience methods to collect and flatten errors from loops.
- My testing strategy is to ensure that caching causes no externally visible changes in output. So, I created two new test methods, `ActionDispatch::IntegrationTest#assert_cacheable_jsonapi_get` and `ActionController::TestCase#assert_cacheable_get`, both of which run their normal respective `get` operations, then repeat the request with various resource caching configurations enabled and assert that the output remains the same. Then I changed nearly all calls to `get` in the test suite to use the helper functions instead. Therefore I can confidently say that the caching system doesn't break any existing tested functionality. :-)
- I added some documentation about caching to the README, but I'm not sure if it's clear. Let me know if any topic needs more detail or a better explanation.

My apologies that this is such a large, complicated PR. Please let me know if you want me to clarify or explain anything I did, general or specific.

lib/jsonapi/request_parser.rb
```diff
@@ -196,154 +304,103 @@ def parse_fields(fields)
...
 
-      unless filters.class.method_defined?(:each)
-        @errors.concat(JSONAPI::Exceptions::InvalidFiltersSyntax.new(filters).errors)
-        return
+      unless filter_params.respond_to?(:each)
+        fail JSONAPI::Exceptions::InvalidFiltersSyntax.new(filter_params)
       end
 
-      filters.each do |key, value|
+      @filters = {}
```
lgebhardt: This is overwriting the default filters set above. I suspect, but haven't checked, that the default filter tests do not include additional filters since this isn't showing up as an error. My project using this did fail a test luckily.

DavidMikeSimon:
@lgebhardt Okay, it seems I was not as careful refactoring RequestParser as I'd thought. :-(

So, I've reverted request_parser.rb and jsonapi_request_test.rb to their original states and am instead passing serializer to Operation after RequestParser has finished. I think the flow of things makes more sense this way anyways, and the PR changelist is much shorter now.

I have the refactored RequestParser code sitting in a different branch on my fork to revisit later if it seems like it would be useful.

DavidMikeSimon:
@lgebhardt Does the test on your project still fail with this revision of the PR?

lgebhardt:
@DavidMikeSimon The tests all pass with the current revision when caching is off.

I am having (and it was on the older revision as well) an issue when I enable caching system wide (I haven't turned it on for any resource). Line 1210 in resource is getting a string with the value of `id` as the second `attrs`value, which causes a problem. I'm trying to track down what's different about my system vs the tests. Any ideas?

Edit: The exception would help: `undefined method 'relation' for "id":String`

lgebhardt:
Ok, line 1145 is getting a SqlLiteral in my case and not an Arel Table. Gotta run now, but I'll see if I can find a solution. My code is littered with custom SQL.

DavidMikeSimon:
@lgebhardt Yeah, without being able to look at your app's models I'm not sure I can be much help. :-( Can you reproduce the problem with a small test case?

lgebhardt:
@DavidMikeSimon It should be easy to reproduce, but I suspect it's just one of many potential issues with accessing the underlying Arel (which may be hand coded SQL) as part of the caching. We will probably want a way to override this so caching can be made to work for custom joins as well. Hopefully I'll have some time to delve into this in the next few days.

DavidMikeSimon:
It's difficult to work around entirely, because we cannot use AR eager loading directly. Possibly I can get preload_included_fragments to deal with custom joins a little more intelligently, or to accept configuration.

DavidMikeSimon:
@lgebhardt Can you paste example code to reproduce the issue? I am not sure where precisely the trouble spots are for custom SQL. The obvious place, specifying a lambda to `has_many` in ActiveRecord, seems to work for me, so either your app is doing something more complex there than I'm used to, or the custom SQL is coming from somewhere else.

DavidMikeSimon:
@lgebhardt I really want to move forward on this, but I'm not sure what to do. If you have a moment, even a rough description of the model configuration that caused the error would let me start working on a test that reproduces it.

lgebhardt:
@DavidMikeSimon sorry I dropped the ball on this. I'll see if I can get something worked up in the next few days to show this issue.

DavidMikeSimon:
@lgebhardt :-(

lgebhardt:
@DavidMikeSimon So sorry for the delay. I had two issues interacting on this, but the core is related to having the `records` using a SQL fragment along with includes. On [resource.rb line 1161](https://github.com/cerebris/jsonapi-resources/blob/712_sqlliteral_issue/lib/jsonapi/resource.rb#L1161) `table` is a `SqlLiteral` in my code. I added a [branch](https://github.com/cerebris/jsonapi-resources/tree/712_sqlliteral_issue) with a test to show the issue. I haven't started looking for a solution. Maybe there isn't a good one other than raising a warning that caching should be turned off for this resource.

DavidMikeSimon:
@lgebhardt Ok, I've handled the test case you posted and a slight variant I came up with. Can you enable the resource_cache in configuration on your app with the code as it is now?

BTW, I wanted to mention: You don't need to write separate controller tests that use `get` and `assert_cacheable_get` to do the same thing, because `assert_cacheable_get` does everything `get` does (e.g. performing the request with caching disabled and setting the response) in addition to checking that caching works.

lgebhardt:
@DavidMikeSimon Thanks! The test suite now passes, for the most part. I have a few query count failures, which I expected. And a few which are almost certainly related to the context dependent results (and I'm sure some more which should fail, but are not showing the context issue). But nothing that I attribute to the new caching code.

I had figured out how `assert_cacheable_get` works, I just wanted to show that the test worked with caching off.

DavidMikeSimon:
Context dependent results?

lgebhardt:
I'm assuming that's what it is. I haven't investigated yet. I've got a hierarchical tree of objects with inherited and calculated permissions. This is all built up in SQL based on the context. At this point I haven't implemented the `attribute_caching_context` method. It's also possible it's unrelated to this.

lgebhardt:
@DavidMikeSimon Thanks for all your work on this.

DavidMikeSimon:
@lgebhardt No worries, glad to help. :-) And, if you find any caching-related bugs in the future, let me know and I'll help resolve them.

lgebhardt:
@DavidMikeSimon Have you had a chance to look at #802 and #803?

DavidMikeSimon:
@lgebhardt Not yet, I'll take a look now.

lgebhardt:
@DavidMikeSimon I added #832 which has two failing tests.

PR #718, opened 2016-05-20
Extract Resource#find's resource transformation

rfb:
Most of the behaviour of JSONAPI::Resource#find can be controlled by
reimplementing #records, #filter_record, #sort_record, etc. By moving the
resource transformation into #resources_for, we've remove one of the final bits
of logic out of the #find. All #find defines is the order of operations and in
many cases can be left as is.

My specific use case is to delegate a single class of models to different
resources. The caching behaviour in #find's transformation prevents me from
simply reimplementing Resource#resource_for.

### DIFF
--- a/lib/jsonapi/resource.rb
+++ b/lib/jsonapi/resource.rb
@@ -636,6 +636,10 @@ module JSONAPI
 
         records = apply_pagination(records, options[:paginator], order_options)
 
+        resources_for(records, context)
+      end
+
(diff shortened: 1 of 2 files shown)


lgebhardt:
@rfb Thanks!

rfb:
Great, thanks @lgebhardt - great work on JSONAPI::Resources! it's been a pleasure to implement into our stack.

Issue #714, opened 2016-05-15
custom record_count like methods


Fedcomp:
How do i make custom methods like record_count? not possible without monkey patching? record_count is hardcoded in FindOperation, any way i can replace with my custom operation?
Say, i want to group results by category and output in total meta information.

gottfrois:
I'm also interested for a solution here. I'm using mongoid where the `count` method do not take any parameter unlike the active-record version.

lgebhardt:
Yesterday I merged a branch that reworks the operation processing aspect of the gem. The old global OperationsProcessor is out and individual resource processors are in (a default one can be defined as well). See https://github.com/cerebris/jsonapi-resources#operation-processors

I think this should give you the flexibility you need. But please see the note at the end of the section. If you override the find method it may not pickup future changes if we alter how find works.

As of now this is only on the master branch, though I hope to get a beta release out soon.

Fedcomp:
i'm afraid i phrased it wrong. I was asking for a way to **_add_** methods like record_count. So i could have my own methods like `aggregate_by(:field)` in `FindOperation` and then use them in my own operation_processor to say, add faceted search metadata in response. Currently lib is highly coupled, and not allowing to substitute own class implementations.

lgebhardt:
@Fedcomp I think I understand, but maybe not. The processor replaces the code that was buried on the `FindOperation`. So it gives you a place to implement your own code to override the default find implementation. You could add your custom methods on a resource specific processor or on a customized default processor. Look at the `Api::V4::BookProcessor` example in the link above.

gottfrois:
@lgebhardt thanks for the clarification. I'm still afraid of this kind of code in master https://github.com/cerebris/jsonapi-resources/blob/master/lib/jsonapi/processor.rb#L163

I think you should have an interface defined on the resource that basically proxy the `count` call to the records. This way, one, can simply override the method definition in the resource if he needs to:

``` ruby
class FooResource < JSONAPI::Resource
  def self.record_count(*args)
    records.count(*args)
  end
end
```

Where, in my case would be as simple as `records.count`

lgebhardt:
@gottfrois Does #720 work for you? It's a little different than you propose. I might be making a wrong assumption that it's the `:all` that AR now requires that's messing you up. With the proposed PR you could override this on a resource level, but there isn't the option to pass in custom arguments as you proposed.

gottfrois:
[...]

PR #720, opened 2016-05-24
Add count_records method to allow easier override for non activerecor…

lgebhardt:
…d counting.

Addresses issue in #714


gottfrois:
Looks good

PR #722, opened 2016-05-25
Improve rescue_from documentation

alyssais:
`rescue_from` wasn't working for me (using JSONAPI::Resources master, Rails 5.0.0.rc1), and I figured out that it was because JSONAPI::ResourceController does not include ActionController::Rescue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -770,6 +770,9 @@ class ApiController < ApplicationController
 end
 ```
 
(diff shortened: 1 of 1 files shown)


lgebhardt:
@penman I think this is another side effect from #595. I'm going to hold off merging until I decide how to handle changes to #595.

PR #741, opened 2016-06-18
Refactor Resource._add_relationship into builder class

mmartinson:
I have found in the past that trying to work with relationship method overrides to deal with edge cases has led me and a trusty debugger on a challenging trip into the heart of the `Resource_add_relationship method`. I had been meaning to have a go at a refactor for a while, partly out of sport, and because I suspect that the complexity there may be discouraging to many trying to work with or understand the source.

Here is a go at it. @lgebhardt I believe this is mostly your code from the early days. I'd appreciate some feedback if you ever have time. I am happy to make further modifications if I missed anything. 

I made changes incrementally with tests passing. I didn't add new unit tests because the current test coverage seemed to very adequately demonstrate when I was breaking things. Larry, if you think there may be edge cases that deserve some close testing let me know. There might be a small performance overhear incurred for the changes at start time(I didn't benchmark) but I suspect it is minimal and worth it for maintainability.

Last thing. Someone else may want to QA this against a real project. I'm not able to run any of my apps against master until I clean up some hacks that don't work beyond 0.7.0 (thanks, I don't need them anymore it seems)

### DIFF
--- a/lib/jsonapi/relationship_builder.rb
+++ b/lib/jsonapi/relationship_builder.rb
@@ -1,152 +1,157 @@
-require 'pry'
-
 module JSONAPI
   class RelationshipBuilder
-    attr_reader :_model_class, :options
+    attr_reader :model_class, :options, :relationship_class
+    delegate :register_relationship, to: :@resource_class
 
-    class DefinitionProxy
-      def initialize(target)
-        @target = target
-      end
-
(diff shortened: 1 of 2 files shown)


lgebhardt:
@mmartinson Thanks!

 Merged by way of #770

PR #743, opened 2016-06-20
Use Resource#_primary_key for default sort_criteria

knapo:
Currently the default `sort_criteria` has hardcoded `id`, what's wrong because the primary key might be different and hardcoding magic string is always bad. Better to refer to `Resource#_primary_key`.

### DIFF
--- a/lib/jsonapi/request_parser.rb
+++ b/lib/jsonapi/request_parser.rb
@@ -16,7 +16,7 @@ module JSONAPI
       @operations = []
       @fields = {}
       @filters = {}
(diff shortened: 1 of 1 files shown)


lgebhardt:
Since `id` has a fixed meaning in a resource, as per the JSON API spec, I'm seeing the need to keep it as a sortable field. The [`construct_order_options` method ](https://github.com/cerebris/jsonapi-resources/blob/master/lib/jsonapi/resource.rb#L864) translates `id` to the `primary_key` that model uses.

knapo:
Ok, makes sense. I didn't override primary key the proper way.

PR #746, opened 2016-06-29
use resource value instead of model attribute to get relationship model id for one to many relationships

seako:
Instead of fetching records using the foreign_key use the relationship name instead. The value stored in the map under the foreign key contains the records for the first relationship defined using that foreign
key. By contrast, using the relationship name returns the records fetched for the relationship defined with that name. I have included a test case that fails without this patch that illustrates how I discovered this issue.

This is analogous to the recently merged pull (https://github.com/cerebris/jsonapi-resources/pull/736) that does the same thing for to_one relationships.

### DIFF
--- a/test/unit/serializer/serializer_test.rb
+++ b/test/unit/serializer/serializer_test.rb
@@ -2175,5 +2175,169 @@ class SerializerTest < ActionDispatch::IntegrationTest
     assert_hash_equals(custom_link_spec, serialized_custom_link_resource)
   end
 
+  def test_includes_two_relationships_with_same_foreign_key
+    serialized_resource = JSONAPI::ResourceSerializer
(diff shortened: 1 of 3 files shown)


test/unit/serializer/serializer_test.rb
```diff
@@ -2175,5 +2175,169 @@ def test_custom_links_with_lambda
...
+            },
+            oddPosts: {
+              links: {
+                self: "/personWithEvenAndOddPosts/1/relationships/oddPosts",
+                related: "/personWithEvenAndOddPosts/1/oddPosts"
+              },
+              data:[
+                {
+                  type: "posts",
+                  id: "1"
```
seako, line 2212: prior to this patch this data contained the even posts and not the odd posts

lgebhardt:
@seako Thanks!

PR #752, opened 2016-07-05
Make default sort criteria blank.

pcreux:
Enable `records(options)` to know if a sort criteria was passed
by the client.

Before this change, we didn't know if the sort criteria "id ASC" passed in
the `options` hash was provided by the client or was the default one.
This would prevent us from implementing a default sort order.

### DIFF
--- a/lib/jsonapi/request_parser.rb
+++ b/lib/jsonapi/request_parser.rb
@@ -16,7 +16,7 @@ module JSONAPI
       @operations = []
       @fields = {}
       @filters = {}
-      @sort_criteria = [{ field: 'id', direction: :asc }]
+      @sort_criteria = []
       @source_klass = nil
       @source_id = nil
       @include_directives = nil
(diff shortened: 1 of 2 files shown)


jalberto:
This seems quite neccesary, Thanks @pcreux, any idea when it will be merge?

lgebhardt:
This is potentially risky with pagination. Without a default sort there's no assurance that the database will return the records in the same order on the next request. I see your point about knowing if the client provided the sort or if it was default. However we can't solve that by removing the default sort entirely.

jalberto:
Fair point.
What about trusting AR default sorting?

lgebhardt:
I'd like to have a default sort for resources instead of relying on AR defaults.

How does #756 look to you? It lets you fully remove the default if you wish, and gives the resource control over the default instead of the request parser.

jalberto:
Thanks! #756 seems a good solution

pcreux:
👍

Closing this PR in favor of #756.

Issue #648, opened 2016-03-06
Controller generator not in current release but is referenced in README


rustyameadows:
I just noticed that the `jsonapi:resource` generator is not in the 0.7.0 release that comes by default from the gem. But you are referencing the generator in the [Controller section](https://github.com/cerebris/jsonapi-resources#controllers) of the README. 

Could be useful to either remove it from the README until it's out of beta or have a versioned README.

PR #754, opened 2016-07-07
specify gem version on Readme

edshadi:
https://github.com/cerebris/jsonapi-resources/issues/648


lgebhardt:
I've added a note to the README about the current version. At this point I don't want a manual update process keeping the README text in sync with the current version.

PR #756, opened 2016-07-09
Default Sort

lgebhardt:
Moves the setting of the default sort from the request_parser to a method on the resource, named default_sort.

This is an alternative to #752

### DIFF
--- a/README.md
+++ b/README.md
@@ -333,6 +333,21 @@ The request will look something like:
 GET /books?include=author&sort=author.name
 ```
 
+###### Default sorting
+
+By default JR sorts ascending on the `id` of the primary resource, unless the request specifies an alternate sort order.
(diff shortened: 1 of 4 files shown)


jagthedrummer:
Was just looking for a way to do this. PR looks great!

krainboltgreene:
A default sort of `id` makes no sense on UUIDs. They cannot be meaningfully sorted.

lgebhardt:
> A default sort of `id` makes no sense on UUIDs. They cannot be meaningfully sorted.

@krainboltgreene Agreed. Here was my thought process in choosing `id`. We need a default sort to, at a minimum, make pagination consistent. `id` is the only field JR requires, so it became the default, by default.

This will work well enough for the simple systems that use counter based ids. For systems using GUIDs you will want to set the default sort to something like `created_at`, though this will vary from project to project

I'm open to other ideas.

PR #760, opened 2016-07-15
Performance fix to skip data loading of belongs_to relations when not necessary

nagirrab:
Hi,

I've been investigating performance issues when loading large amounts of data, and noticed a whole load of data loading for `belongs_to` and `has_one` relationships that shouldn't be happening.

I finally nailed it down to this: it will load the relationships even if it doesn't have to, because it doesn't check for the id directly but instead loads the resource then picks the id.

This seems to work, but let me know if there's another way you'd prefer to fix it and I'd be happy to fix it up that way.

Best,
Hugh

### DIFF
--- a/lib/jsonapi/resource_serializer.rb
+++ b/lib/jsonapi/resource_serializer.rb
@@ -298,9 +298,18 @@ module JSONAPI
 
     # Extracts the foreign key value for a to_one relationship.
     def foreign_key_value(source, relationship)
-      related_resource = source.public_send(relationship.name)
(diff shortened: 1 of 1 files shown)


DavidMikeSimon:
@lgebhardt I'm concerned that this seems to undo the fix created by #736

lgebhardt:
@DavidMikeSimon I think you are right. I wonder if the right option is to keep this as the default and make the #736 logic optional for the rare cases where you need the model lookup.

PR #762, opened 2016-07-17
Allow you to disable eager loading on relationships on include

nagirrab:
Hi,

Consider this more of a discussion-starter than a full PR - I'm happy to refactor it/keep it on my own fork as necessary.

The automatic eager loading of relationships works great in the normal case but I have 2 cases where it doesn't fit:
- Very large queries where the complexity of the join grows to large, but the actual data being joined is the same repeatedly - without eager loading, this will naturally get picked up by the query cache.
- 'Relationships' that aren't actual active record relationships. E.g. when using the ancestry gem, you cannot expose `children` or `parent`. This used to work in old versions prior to eager loading, and I've found it useful.

In order to fix this, I've added a new `eager_load_on_include` option to relationship, which defaults to true. When set to false, it will not include the item unless it forms part of the filter.

Happy to make any changes required, etc.

Best,
Hugh

### DIFF
--- a/README.md
+++ b/README.md
@@ -541,6 +541,7 @@ The relationship methods (`relationship`, `has_one`, and `has_many`) support the
  * `polymorphic` - set to true to identify relationships that are polymorphic.
  * `relation_name` - the name of the relation to use on the model. A lambda may be provided which allows conditional selection of the relation based on the context.
  * `always_include_linkage_data` - if set to true, the relationship includes linkage data. Defaults to false if not set.
+ * `eager_load_on_include` - if set to false, will not include this relationship in join SQL when requested via an include. You usually want to leave this on, but it will break 'relationships' which are not active record, for example if you want to expose a tree using the `ancestry` gem or similar, or the SQL query becomes too large to handle. Defaults to true if not set.
 
 `to_one` relationships support the additional option:
  * `foreign_key_on` - defaults to `:self`. To indicate that the foreign key is on the related resource specify `:related`.
--- a/lib/jsonapi/include_directives.rb
+++ b/lib/jsonapi/include_directives.rb
@@ -19,7 +19,9 @@ module JSONAPI
     #   }
     # }
 
-    def initialize(includes_array)
+    def initialize(resource_klass, includes_array, force_eager_load: false)
+      @resource_klass = resource_klass
+      @force_eager_load = force_eager_load
       @include_directives_hash = { include_related: {} }
       includes_array.each do |include|
         parse_include(include)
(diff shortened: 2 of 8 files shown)


lgebhardt:
Looks good to me. Is there any thing else you want to add to this PR?

nagirrab:
I'm still trying to get query complexity down, but I think this has gone far enough that I'll probably have to write something custom, so this one's good to go.

lgebhardt:
@nagirrab Would you mind rebasing this against master?

nagirrab:
@lgebhardt done

lgebhardt:
@nagirrab Thanks!

thibaudgg:
Hi @nagirrab and @lgebhardt,

We're getting the `[RELATIONSHIP NOT FOUND]` warning message when including polymorphic association but the association is still properly included. Seems weird, any reasons for this?

\cc @senny

hidde-jan:
@thibaudgg did you define the controller and the resource for the polymorphic model?

From the readme:

> The polymorphic relationship will require the resource and controller to exist, although routing to them will cause an error.
> 
> ``` ruby
> class TaggableResource < JSONAPI::Resource; end
> class TaggablesController < JSONAPI::ResourceController; end
> ```

thibaudgg:
@hidde-jan only the resource, from what I see from the code base in `IncludeDirectives` it doesn't look for a controller no?

hidde-jan:
Hmm, I'm checking and indeed the controller doesn't seem necessary. Did you disable eager loading for the polymorphic relationship?

``` ruby
class RelatedModelResource
  has_one :polymorphic_model, polymorphic: true, eager_load_on_include: false
```

Are you using the include like `/main_model?include=related_model.polymorphic_model`?

thibaudgg:
I tried with and without `eager_load_on_include: false` but I get the warning message in both cases.
I'm using the include like that `/main_model?include=polymorphic_model.another_model.another_model`

hidde-jan:
> `/main_model?include=polymorphic_model.another_model.another_model`

In that case, are you sure the `polymorphic_model` instances _always_ have the `another_model` relationship? Did you also define the relationship on the `PolymorphicModelResource`?

thibaudgg:
The polymorphic model resource is an "empty" class:
 `class PolymorphicModelResource < JSONAPI::Resource; end` 
and the `another_model` relationship only exists on some of them, but I get the warning message even when the `another_model` relationship exists on that polymorphic model.

hidde-jan:
I think you need to define the relationship on the polymorphic model resource as well.

thibaudgg:
It seems like a duplication of logic to me, I don't see the point of having associations that are not shared by all polymorphic model duplicated in the abstract polymorphic resource.

@lgebhardt is it really how it should be done?

PR #769, opened 2016-07-20
[README] Add notes on Authorization

shellandbull:
It doesn't come built-in with the framework, but it's a good idea to point people in the right direction 👍

### DIFF
--- a/README.md
+++ b/README.md
@@ -1890,7 +1890,7 @@ phone_number_contact GET    /phone-numbers/:phone_number_id/contact(.:format) co
 
 ### Authorization
 
-To this day `json-api-resources` doesn't come with built-in primitives for authorization. However multiple users of the framework have come up with different approaches, check out:
+Currently `json-api-resources` doesn't come with built-in primitives for authorization. However multiple users of the framework have come up with different approaches, check out:
 
 - [jsonapi-authorization](https://github.com/venuu/jsonapi-authorization)
 - [pundit-resources](https://github.com/togglepro/pundit-resources)
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ backed by ActiveRecord models or by custom objects.
     * [Key Format] (#key-format)
   * [Routing] (#routing)
     * [Nested Routes] (#nested-routes)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1199,7 +1200,7 @@ rails generate jsonapi:controller contact
 ###### ResourceControllerMetal
 
 `JSONAPI::Resources` also provides an alternative class to `ResourceController` called `ResourceControllerMetal`.
-In order to provide a lighter weight controller option this strips the controller down to just the classes needed 
+In order to provide a lighter weight controller option this strips the controller down to just the classes needed
```
shellandbull, line 1203: trailing whitespace, caught by my text editor

README.md
```diff
@@ -1211,7 +1212,7 @@ end
 ```
 
 Note: This may not provide all of the expected controller capabilities if you are using additional gems such as DoorKeeper.
- 
+
```
shellandbull, line 1215: same as above

README.md
```diff
@@ -1887,6 +1888,15 @@ phone_number_contact GET    /phone-numbers/:phone_number_id/contact(.:format) co
 
 ```
 
+### Authorization
+
+To this day `json-api-resources` doesn't come with built-in primitives for authorization. However multiple users of the framework have come up with different approaches, check out:
```
lgebhardt: How about replacing "To this day" with "Currently"

shellandbull: will do

shellandbull:
@lgebhardt Small question though - Shouldn't we add a small entry mentioning how a user could use a processor to validate a request? I think there's enough stuff in `json-api-resources` to come up with something clean, but I'd have to dedicate it some time

lgebhardt:
@mariogintili Thanks! Yes, I think getting some additional documentation on using processors for authorization would be nice. I merged this PR, but a second PR when you have some time to dedicate to this would certainly be welcome.

shellandbull:
@lgebhardt I gave it some thought and added an extra comment on #16 let's keep the discussion there or maybe open a new PR 😄

PR #773, opened 2016-07-21
Don't cache resource classes based on model class

alyssais:
`resource_for_model` passes the model instance, so resources can vary based on model instance.
Caching them based on model class, therefore, can wrongly use a resource that for a record that should use a different resource.

I just ran into this when choosing resource based on the type of a polymorphic relationship.


lgebhardt:
@penman Thanks!

Issue #766, opened 2016-07-19
Authorization?


shellandbull:
Hello guys,

I've been looking at the wiki/project and can't find any documentation in regards to restricting some actions in a controller based on some characteristics in the request, i.e:

I'm using `json-api-resources` for a resource called `tickets` and I want to restrict write actions to only those requests coming from users that are administrators. How could I go on and accomplish this?

bmichelsen:
Hi there!

From the authorization discussion at #16 there are (at least) two viable options:
- [jsonapi-authorization](https://github.com/venuu/jsonapi-authorization)
- [pundit-resources](https://github.com/togglepro/pundit-resources)

Also, have a look at the excellent comment by @valscion that [explains the differences](https://github.com/cerebris/jsonapi-resources/issues/16#issuecomment-222438975) between the two approaches.

Hope that gets you started!

shellandbull:
@bmichelsen @lgebhardt

I was looking at [this](https://github.com/cerebris/jsonapi-resources/blob/master/lib/jsonapi/processor.rb#L55) And I think we can get away with building validation primitives into the framework in a very simple _and_ lightweight manner:

How about we just add another callback called `validate` that yields the request object - coming from the controller - and lets the end user validate agains't whatever he wants i,e: ip address, params, session, etc?

My only concern would be how to let the users actually inject errors to the errors payload in case the request is valid - any takers? I'm happy to implement this one since I need it for a project 🌞 If you guys are happy with this we can open a proper issue with a `TODO` list, etc.

So at a resource level you could define something like

``` ruby
validate do |request, errors|
  errors.add(:access, "you don't have access to this application") if request. remote_ip.match(/a-blacklisted-ip-address/)
end
```

The user doesn't need to raise exceptions or anything else, just touch an `errors` object, you can instantiate a pundit policy if you want there...pretty much anything 🌞

lgebhardt:
@mariogintili I would put remote_ip checks more as an Authentication concern and as such could be handled earlier in the stack as a controller callback. Though I'm sure a case could be made for controlling updating a particular resource based on ip address.

Are you proposing `validate` on the processor or the resource?

shellandbull:
@lgebhardt the remote_ip example was banal and just to showcase an example use case for some developers -

> Are you proposing validate on the processor or the resource?

Proposing a `validate` on the processor that yields 2 things:
[...]

PR #778, opened 2016-07-22
[FEATURE] Processor Validations

shellandbull:
# TODO
- [x] Ensure a user can validate by implementing `validate`
- [ ] Ensure an error response is compliant with [JSON-API](http://jsonapi.org/format/#errors)
- [ ] Better test coverage

ref #766


shellandbull:
@lgebhardt Would like to get some feedback on this before I get going :)

lib/jsonapi/acts_as_resource_controller.rb
```diff
@@ -219,6 +219,18 @@ def render_results(operation_results)
...
     def create_response_document(operation_results)
+      if operation_results.custom_errors?
+        build_error_response_document(operation_results)
+      else
+        build_response_document(operation_results)
+      end
+    end
+
+    def build_error_response_document(operation_results)
+      fail NotImplementedError.new("TODO")
```
shellandbull, line 230: Was planning to create a `JSONAPI::ErrorsResponseDocument` that inherits from `JSONAPI::ResponseDocument`

lib/jsonapi/operation_results.rb
```diff
@@ -20,16 +20,44 @@ def has_errors?
...
     end
 
+    def custom_errors?
+      @has_errors &&
+      @results.all? do |result|
+        result.is_a?(JSONAPI::ErrorsOperationResult) && result.errors.is_a?(ActiveModel::Errors)
+      end
+    end
+
     def all_errors
```
shellandbull, line 30: Unsure if `all_errors` should be an array - Maybe an Enumerable that represents a collection of failed operation results

lib/jsonapi/operation_results.rb
```diff
@@ -20,16 +20,44 @@ def has_errors?
...
-      if @has_errors
-        @results.each do |result|
-          if result.is_a?(JSONAPI::ErrorsOperationResult)
-            errors.concat(result.errors)
-          end
-        end
+
+      if concateable_errors?
+        concatenate_errors(errors)
+      elsif custom_errors?
```
shellandbull, line 35: Follow-up from comment above: since we now have multiple types we have to adapt the interfaces somehow - In my opinion we should have some sort of proxy that knows how to normalise what an error IS regarding of what it is or where it comes from

test/integration/requests/request_test.rb
```diff
@@ -1107,4 +1107,16 @@ def test_getting_resource_with_correct_type_when_sti
...
+
+  # TODO Needs to be discussed an implemented
+  def test_errors_payload_from_write_processor_validation
+  end
+
+  # TODO Needs to be discussed an implemented
+  def test_errors_payload_from_write_resource_validation
+  end
+
+  # TODO Needs to be discussed an implemented
```
shellandbull, line 1119: Happy to listen to suggestions as of what I should be covering here

lib/jsonapi/acts_as_resource_controller.rb
lgebhardt, line 230: `operation_results` should not need to contain custom_errors. You could define a new `JSONAPI::Error` derived error for each of the custom error types. That will buy you all the built in handling for rendering them compliant with JSON API.

lgebhardt:
Reviewing this I'm concerned you are missing some key details of how the system was designed to work. Not that that should surprise me as it's got a lot of layers, and isn't as clear as I'd like.

So could we step back and go over the goals you have for the `validate` method?

shellandbull:
@lgebhardt Yep, we all due honesty I found the project a bit hefty to work with because:
- TheDSL goes too far away from conventions, for example: the error objects don't inherit from `StandardError` or something similar, they are just POROs
- Test scenarios manipulate globals - `$test_user` - which can get out of hand really quick
- Some declared primitives here are already a part of Rails and are just duplicated, for example the `lib/jsonapi/error_codes.rb` vs http://apidock.com/rails/ActionController/StatusCodes

I've decided not to continue developing this PR but feel free to take over 😄

Things like managing errors will be something we'll manage through filters as previously suggested - thanks for the help anyways

lgebhardt:
Understood. I'm still interested in what your goals are for the `validate` method on the processor. So far I've been able to use the callbacks to handle any pre or post processing of operation data, including returning errors. The one limitation I've run into has been that it's hard to return a full set of errors if the request has multiple, with the exception of model validations where we are able to covert each validation error into a JSON API error.

To address your points:

The Errors are derived from RuntimeError which is derived from StandardError.

I agree the $test_user global feels hackish. I'm certainly open to better options.

`error_codes.rb` does share some overlap with the standard StatusCodes, but that's mostly incidental. JSON API provides for having a `code` in the error objects. The `error_codes` are used to provide a standard set of codes for errors which can either be presented as numeric codes or text. JSON API also has a `status` in the error objects. This is where the http://apidock.com/rails/ActionController/StatusCodes are used.

shellandbull:
@lgebhardt the goal of `validate` is to give users an interface to populate a custom `errors` document, this could be handled at a model level or in a callback that takes over if need be 😄

lgebhardt:
Model validation errors are already supported. If you raise single errors in the existing callbacks they will result in an `ErrorsOperationResult` which I think is analogous to an errors document.

I don't see why you couldn't create new Error similar to `ValidationErrors` which would wrap multiple custom error types. You could do all your checks in the `before_operation` callback, collect all the custom errors and raise them at the end in your new Error. It would be very little work and should give you all of what I think you are trying to accomplish.

PR #784, opened 2016-07-28
Extend filter functionality

dsojevic:
This PR adds the following:
- Ability to use custom filter classes and/or instances using `with:` on the `filter` and `filters` methods
- Ability to set default custom filters for a resource
- Ability to set default custom filters for all resources

The README shows examples of how to use the custom filters and how to set global and resource default filters.

The current test suite doesn't appear to test the existing functionality, so I'm not sure where to put tests for this feature.

This functionality is currently being used in-house and would be good to see it in core.


lib/jsonapi/resource.rb
```diff
@@ -819,13 +830,17 @@ def verify_filter(filter, raw, context = nil)
...
         end
 
-        strategy = _allowed_filters.fetch(filter, Hash.new)[:verify]
+        strategy = fetch_filter_strategy(filter.to_sym, :verify)
 
         if strategy
           if strategy.is_a?(Symbol) || strategy.is_a?(String)
             values = send(strategy, filter_values, context)
-          else
+          elsif strategy.is_a?(JSONAPI::Filter)
```
NuckChorris: Would it be better to duck-type this with `elsif strategy.respond_to?(:verify)` or is it truly necessary that it inherit from `JSONAPI::Filter`?

dsojevic: I initially had it checking if it would respond to the type (ie. `:apply` or `:verify`), but thought that perhaps restricting it to descendants of `JSONAPI::Filter` might be best so that the filtered attribute was present/available in each of the methods.

I'm not attached to what I settled on though, so I'd be more than happy to change it if people would prefer it to be duck-typed!

dsojevic:
@lgebhardt, if you've had a chance to look over this PR, are there any suggested/required changes to this to take it to the next stage?

Thanks!

lgebhardt:
I would like to get some tests for this. I'll see if I can add a few.

HoyaBoya:
> I would like to get some tests for this. I'll see if I can add a few.

Not sure if these tests are applicable, but feel free to grab.

https://github.com/cerebris/jsonapi-resources/pull/939/files#diff-529919e48099bb2326882042e858d0efR1064

https://github.com/cerebris/jsonapi-resources/pull/939/files#diff-248ce206528041a667a94b064423dd5bR228

It was the best I could do based on my understanding of the code and coverage. Hope this helps this PR. Great work!

lgebhardt:
@dsojevic Sorry for letting this sit for so long. #939, #942 and a discussion on [gitter](https://gitter.im/cerebris/jsonapi-resources) have thankfully brought this back into my focus again.

lgebhardt:
A few comments for discussion:

The `resource_filter` name seems confusing to me. Its purpose looks to be as the default strategy. Maybe rename to `filter_strategy` or `default_filter_strategy`? Same with `config.default_filter`.

I'm unsure if we need the complexity of being able to specify separate `apply` and `verify` `resource_filter`s. I'd rather see the Filter object handle both, but maybe I'm missing a good use-case for it.

Could we model the `Filter` class on [Validator](https://github.com/rails/rails/blob/master/activemodel/lib/active_model/validator.rb#L94) and store the options as part of the `filter` call and use them in the filter initializer?

PR #787, opened 2016-08-01
Refactor controller so caching is easier 

drakedowsett:
Because the `process_request` call is so tightly coupled with the final `render` intercepting results for caching is non-trivial. This PR refactors the code so that an intermediate method is called which produces the final result and then renders it. To add caching, one need only override this function:

``` ruby
    def process_request(action)
      return super(action) unless [:index].include?(action)
      result = Rails.cache.fetch(request.fullpath, expires_in: 1.hour) do
        _process_request
      end
      render(result)
    rescue => e
      handle_exceptions(e)
    end
```

And this can easily be customized on a per-project basis.

### DIFF
--- a/lib/jsonapi/acts_as_resource_controller.rb
+++ b/lib/jsonapi/acts_as_resource_controller.rb
@@ -13,72 +13,75 @@ module JSONAPI
     end
 
     def index
-      process_request
+      process_request(__method__)
     end
 
     def show
-      process_request
+      process_request(__method__)
     end
 
     def show_relationship
-      process_request
+      process_request(__method__)
     end
 
     def create
       return unless verify_content_type_header
-      process_request
+      process_request(__method__)
     end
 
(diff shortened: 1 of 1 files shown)


drakedowsett:
Submitting this based on `master` instead of #786. @lgebhardt please let me know if anything else is needed.

drakedowsett:
@lgebhardt @DavidMikeSimon Any thoughts on this/timeframe for consideration?

lgebhardt:
@drakedowsett I'm in the middle of a major refactor to support the new JSON API Operations spec. This will require quite a few more changes to the existing codebase, and so far it looks like it will step on your changes in this PR. I'd like to hold off on this for now. Sorry for the lack of feedback on this.

drakedowsett:
@lgebhardt Thanks, will look forward to the refactor.

PR #788, opened 2016-08-02
Add table of contents to README.

eoinkelly:
The README is huge which is awesome :sparkles: but it does make it a bit hard to navigate as I refer to it a lot. This PR adds a Table of Contents.

Unfortunately this ToC will require manual updating whenever the _structure_ of the README is changed. I think this wouldn't be too burdensome but I can understand the team not wanting to take it on so no worries if you need to decline this.

The table of contents was generated with the `markdown-toc` npm package via

```
$ cd path/to/this/repo
$ npm install -g markdown-toc
$ markdown-toc -i README.md
```

I used the `markdown-toc` tool because it is works, is freely available from a package manager that we all probably use already and has some editor plugins ([atom has a package](https://atom.io/packages/markdown-toc) that uses it). 

Thanks for your work on this awesome project btw :100:

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,57 @@ branch. This may contain information that is not relevant to the release you are
 
  ---
 
+<!-- toc -->
+
+- [Table of Contents](#table-of-contents)
+- [Demo App](#demo-app)
+- [Client Libraries](#client-libraries)
+- [Installation](#installation)
+- [Usage](#usage)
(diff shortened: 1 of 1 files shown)


eoinkelly:
Urgh there is already a ToC. I may need a holiday ... Sorry to waste time.

PR #789, opened 2016-08-02
avoid situation where the linkage data could be partially calculated

barelyknown:
The existing tests pass, though I did not take the time to create a failing test case for the situation (mostly because the easiest way would have been to test the `to_one_linkage` and `to_many_linkage` methods directly, and I didn't have time to figure out how to isolate them effectively).

This fixes the situation where the current user does not have permission to access a relationship, and the library was returning a partially complete `data` object for that relationship when it was polymorphic (the `id` would be provided, but the `type` would be an empty string because of the way that `Relationship.type_for_source` works). To keep it all simple and avoid any side effects, I'm just _not_ returning a partial data object (where either the `id` or the `type` are `nil`) now.

### DIFF
--- a/lib/jsonapi/resource_serializer.rb
+++ b/lib/jsonapi/resource_serializer.rb
@@ -353,8 +353,10 @@ module JSONAPI
     end
 
     def to_one_linkage(source, relationship)
-      return unless linkage_id = foreign_key_value(source, relationship)
-      return unless linkage_type = format_key(relationship.type_for_source(source))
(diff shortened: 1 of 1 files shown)


barelyknown:
Hold on with this. Need to manually test a bit more.

barelyknown:
OK, I needed to guard against empty strings too. It should be good now.

lgebhardt:
@barelyknown Thanks!

PR #794, opened 2016-08-05
Adds multi_json

warmwaffles:
Allows others to choose their flavor of json parsing.


NuckChorris:
:-1: `multi_json` is slowly being removed from other projects in favor of just using the built-in `JSON` gem, especially now that both OJ and Yajl have a compatible mode via `Oj.mimic_json` and `require 'yajl/json_gem'`.  There's really no point to MultiJson any more

warmwaffles:
https://github.com/ohler55/oj/blob/v2.17.1/lib/oj/mimic.rb

warmwaffles:
I get what you are saying though, https://github.com/jmespath/jmespath.rb/issues/6#issuecomment-95721763

PR #795, opened 2016-08-07
Remove duplicate definition of Resource#cached_resources_for method

valscion:
The `cached_resources_for` method is defined twice with the same method body. See two methods up, and there is already a definition for this method.


dgeb:
Good catch - thanks!

PR #797, opened 2016-08-07
Use strings in whitelist to avoid mismatch of object_id

ghost:
I had a problem while trying to rescue_from an custom error defined inside application (in app/controllers/concerns). During debug I discovered that `CustomError.object_id` did not match the one from `exception_class_whitelist`.

To avoid this I propose to compare class names as strings rather than objects.


lgebhardt:
@nooulaif Thanks!

PR #801, opened 2016-08-17
Refactor controller filters

lgebhardt:
Refactor `ensure_correct_media_type` and `ensure_valid_accept_media_type` filters to the methods that need them.

Prevents order conflicts with application defined filters.


dgeb:
Will be good to remove those filters.

Wondering if we should refactor `ensure_correct_media_type` -> `verify_content_type_header` and `ensure_valid_accept_media_type` -> `verify_accept_header` to be a bit more clear about what the methods are checking.

lgebhardt:
Good point, and done.

dgeb:
Thanks! :shipit:

PR #804, opened 2016-08-19
Fix media type matching

iosadchiy:
Bug: When client sends Accept header with allowed media type not in the first position, the request gets rejected. For example:

```
GET /posts
Accept: application/json, application/vnd.api+json, */*
```

Expected: HTTP 200 OK
Got: HTTP 406 Not acceptable

This happens due to incorrect header parsing in JSONAPI::ActsAsResourceController#media_types_for. The PR addresses the issue


lgebhardt:
@iosadchii Thanks!

PR #806, opened 2016-08-20
Update method name in README

iosadchiy:
This method has been renamed in #801. Let's update README accordingly


lgebhardt:
@iosadchii Thanks!

PR #808, opened 2016-08-22
Fix request tests

shunwen:
Hi there, here are two small fixes
1. Not setting content type in `test_put_single_without_content_type` 
2. `test_get_accepting_multiple_content_types` is testing multiple accept types, suggest to rename it.


lgebhardt:
@shunwen Thanks!

PR #809, opened 2016-08-22
Get relationship id from the resource

gottfrois:
We should avoid getting relationship id from the `model` directly to allow clients to customize how relationship ids are fetched. The resource exposes the relationship id through a method which delegates to the `model`. This way we allow clients to override the relationship id in the resource to fit their needs.

Today, by default, the resource object already delegate the relationship id method to the model:

``` ruby
class AuthorResource
end

class MessageResource
  has_one :author
end

message.author_id
=> 1
message._model.author_id
=> 1
```

This way, a client can override the relationship id method:

``` ruby
class MessageResource
  has_one :author

  def author_id
    @model.author.some_custom_id
  end
end
```


knapo:
+1

lgebhardt:
@gottfrois Thanks!

PR #814, opened 2016-08-28
Making sure we return the right path for a rails engine that doesn't have additional namespaces

nathanpalmer:
The current rails engine code makes an assumption that you will always have additional namespaces under your main Engine. _This isn't the default behavior in rails_ and I found when experimenting with putting our v2 api into an engine that it wouldn't generate the paths properly.

The main fix is in link_builder to only generate the path with `end_path_name` instead of concating `base_path_name` with it (which would of course generate `_person_path`.)

Added to the engine tests to show that it was generating properly.


lgebhardt:
@nathanpalmer Thanks!

PR #818, opened 2016-08-31
Fix nested resources creating N+1 queries on includes

mgoodings:
Added include_directives support to resource relationships to avoid N+1 queries from includes on a nested route.

I had a route of /resource1/:id/resource2?include=resource3 and noticed it was creating N+1 queries for each resource3 access.

I've patched JSONAPI::RelationshipBuilder.build_to_many to follow the same logic JSONAPI::Resource.find


lgebhardt:
@mgoodings Thanks!

PR #819, opened 2016-09-01
Removes ActiveRecord dependency for #model_hint

gottfrois:
Class.name is the same as Class.to_s

I don't see a particular reason to have the AR dependency here. Missing something?


lgebhardt:
@gottfrois Thanks!

PR #822, opened 2016-09-07
Fix code example for relationship specific method definitions

valscion:
There is no such method or variable as `relationship` defined in the
special record_for_#{relationship} method. For the commented-out code
to actually work, we first need to fetch the relationship object before
we can ask for the relation name.

We could also discuss whether these special methods could somehow also have the `relationship` already available to them, so that changes like these wouldn't be necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -735,11 +735,13 @@ class PostResource < JSONAPI::Resource
   has_many :comments
 
   # def record_for_author
+  #   relationship = self.class._relationship(:author)
   #   relation_name = relationship.relation_name(context: @context)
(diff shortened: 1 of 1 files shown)


lgebhardt:
@valscion Thanks!

I'm open to ideas for making the `relationship` automatically available. If you think it's worth pursuing it would be best to open an issue to track it.

valscion:
> I'm open to ideas for making the relationship automatically available. If you think it's worth pursuing it would be best to open an issue to track it.

Yeah, I'll think about it and open an issue when/if I have something to share ☺️

PR #827, opened 2016-09-09
call server_error_callbacks on handle_exceptions in controller

senid231:
if exception is raised not within `OperationDispatcher`  server_error_callbacks does not called


lgebhardt:
@senid231 Thanks!

PR #828, opened 2016-09-11
Let `jsonapi_resource` recognize immutable

yawboakye:
Currently, `create`, `update`, `destroy` routes are generated for
resources that are marked as immutable. This change fixes that bug.

### DIFF
--- a/lib/jsonapi/routing_ext.rb
+++ b/lib/jsonapi/routing_ext.rb
@@ -32,6 +32,12 @@ module ActionDispatch
             options[:except] = [:new, :edit]
           end
 
+          if res._immutable
+            options[:except] << :create  unless options[:except].include?(:create)  || options[:except].include?('create')
+            options[:except] << :update  unless options[:except].include?(:update)  || options[:except].include?('update')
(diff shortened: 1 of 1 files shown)


lib/jsonapi/routing_ext.rb
```diff
@@ -32,6 +32,12 @@ def jsonapi_resource(*resources, &_block)
             options[:except] = [:new, :edit]
           end
 
+          if res._immutable
+            options[:except] << :create
```
lgebhardt: We should protect against having duplicates in the except list. This is done above and could look like:

``` ruby
  if res._immutable
    options[:except] << :create unless options[:except].include?(:create) || options[:except].include?('create')
    options[:except] << :update unless options[:except].include?(:update) || options[:except].include?('update')
    options[:except] << :destroy unless options[:except].include?(:destroy) || options[:except].include?('destroy')
  end
```

yawboakye: I copied how `jsonapi_resources` updates the `options` hash. I think I have to change it in both places then.

lgebhardt:
@yawboakye Thanks!

PR #833, opened 2016-09-14
Improve Filter docs on apply/verify + correct simple example + add a practical complex example

kevinzwhuang:
**Description**
The _filter_ section in the documentation is lacking in describing what `value` is for newcomers who want to `apply` a custom filter. It's easy for newcomers to assume that `value` is the string value of the query param of the filter, when it is actually the array of string params by default (if not provided a `verify` callable). 

The example also confuses this distinction too, by comparing the value to the symbol `:public` - where `value` would actually be an array of `['public']`, since there is no `verify` callable to modify `value`.

**Changes Proposed:**
- Add clarifying note describing what `value` is to educate newcomers.
- Simplify the example by providing the query param in comments and explicitly using the `apply` callable in the default case with a `value` array without a `verify` callable.
- Add a more complex example that combines both `verify` and `apply` callables to show how to use them in actual practice.

### DIFF
--- a/README.md
+++ b/README.md
@@ -654,10 +654,10 @@ You may customize how a filter behaves by supplying a callable to the `:apply` o
 apply that filter. The callable is passed the `records`, which is an `ActiveRecord::Relation`, the `value`, and an
 `_options` hash. It is expected to return an `ActiveRecord::Relation`.
 
-This example shows how you can implement different approaches for different filters.
-
 Note: When a filter is not supplied a `verify` callable to modify the `value` that the `apply` callable receives,
 `value` defaults to an array of the string values provided to the filter parameter.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -656,9 +656,14 @@ apply that filter. The callable is passed the `records`, which is an `ActiveReco
 
 This example shows how you can implement different approaches for different filters.
 
+Note: When a filter is not supplied a `verify` callable to modify the `value` that the `apply` callable receives,
```
lgebhardt, line 657: Could we put the Note either above line 657 or below the example.

kevinzwhuang:
@lgebhardt 👍  moved the note up

lgebhardt:
@kevinzwh Thanks!

PR #845, opened 2016-09-21
Handle all base spec requests as a single operation

lgebhardt:
Getting ready for multiple operation support in requests

Also removes the feature of creating and modifying multiple resources as this is not compatible with final version of the JSON API spec.


lib/jsonapi/processor.rb
```diff
@@ -13,7 +13,7 @@ class Processor
                                        :replace_polymorphic_to_one_relationship,
                                        :create_to_many_relationship,
                                        :replace_to_many_relationship,
```
dgeb: Seems like `replace_to_many_relationship` should now be `replace_to_many_relationships` and `create_to_many_relationship` should now be `create_to_many_relationships` for consistency.

lgebhardt: Good point. Fixed.

valscion:
Ooh, interesting! I guess this'll break `jsonapi-authorization`, but patching for this change shouldn't be a big deal.

I didn't quite get what you meant with this:

> Also removes the feature of creating and modifying multiple resources as this is not compatible with final version of the JSON API spec.

Was this something like sending an array of objects in POST payload (i.e. `data` in root was array) that was dropped? If so, I didn't even know JR had handled them 😅

lgebhardt:
@valscion Yes, it used to accept an array of objects in the payload. That was a feature from before the JSON API spec was finalized and I missed removing it.

Sorry about the churn breaking `jsonapi-authorization`. I suspect more breaking changes will be coming with the new operations support.

valscion:
No worries, we've got our specs on the nice reques layer so upgrading is quite easy. I'm really looking forward to new operations support, our app would benefit from them :)

Issue #485, opened 2015-10-12
Reducing N+1 queries by eager loading


nathanpalmer:
I read through the documentation and couldn't find a built-in solution to this problem. When certain models are serialized we need to make sure that the query to retrieve that entity always adds an includes to avoid all the lazy loaded N+1 queries. We can do this via `records` as long as the request is on the resources controller. If it's not then I couldn't see a place to extend it.

Let me give an example

``` ruby
class ProjectResource
  has_one :company
end
```

``` ruby
class CompanyResource
  attribute :plan_start_at

  def plan_start_at
    @model.subscription.created_at
  end
end
```

As you'll notice each time a company is rendered it is going to retrieve a value out of it's subscription. However, the problem comes in when we request a URL like this

`/api/projects?include=company`

It's using the `records` method off of the projects resource and even when it goes to process the include on company it's not calling records on the company resource so we end up with a separate query for every companies subscription.

Now, from what I can tell maybe this is supposed to be taken care of by also requesting the subscription like this

`/api/projects?include=company,company.subscription`

However this doesn't completely work since I don't need the subscription object and requesting the subscription isn't a concern of the client (since we simply need the subscription in order to calculate a property on the company.)

lgebhardt:
If requesting `company.subscription` does invoke the eager loading you could ensure it's in the `includes` call by overriding the `apply_includes` method on your Projects resource. You could selectively add it only when the company is requested. It's obviously a hack, but it should work.

A better solution might be to allow additional includes for eager loading on each relationship. I'm not sure how common that would be and whether it's worth the time to add it to the project.

nathanpalmer:
Thanks, that's where I've looked. So far we've put together a quick extension that allows us to include anything additional by adding a function called `always_include` to the resource

``` ruby
    def _append_always_include(resource_klass, model_include, options)
      if resource_klass.present? && resource_klass.respond_to?(:always_include)
        resource_klass.always_include(options) + model_include
      else
        model_include
      end
    end

    def resolve_always_includes(resource_klass, model_includes, options = {})
      case model_includes
        when Array
          return model_includes.map do |value|
            resolve_always_includes(resource_klass, value, options)
          end
        when Hash
          return Hash[model_includes.map do |key, value|
            relationship = resource_klass._relationships[key]
            [key, resolve_always_includes(relationship.resource_klass, _append_always_include(relationship.resource_klass, value, options), options)]
          end]
        when Symbol
[...]

PR #847, opened 2016-09-22
Add ::always_load to resources for eager loading.

mrloop:
Enables resources to be defined with `always_includes` for eager loading. e.g.

```
class TagResource < JSONAPI::Resource  
  always_include posts: [:section]
  has_many :posts
end

  has_one :section
  has_many :tags, acts_as_set: true, inverse_relationship: :posts, eager_load_on_include: false
end

class SectionResource < JSONAPI::Resource
  has_many :posts
end
```

Normalize and merge the model includes and always includes, iterate over the model includes and merge their always includes. The model and always includes can both be nested.

See #485

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@ backed by ActiveRecord models or by custom objects.
     * [Filters] (#filters)
     * [Pagination] (#pagination)
     * [Included relationships (side-loading resources)] (#included-relationships-side-loading-resources)
+    * [Loaded relationships (eager-loading resources)] (#loaded-relationships-eager-loading-resources)
     * [Resource meta] (#resource-meta)
     * [Custom Links] (#custom-links)
     * [Callbacks] (#callbacks)
@@ -1034,6 +1035,40 @@ Will get you the following payload by default:
     }
   }]
 }
+```
+
+#### Loaded relationships (eager-loading resources)
+
+To avoid N+1 queries when a resource attribute relies on attributes of related model records to calculate it `always_load` can be used. [Included relationships](#included-relationships-side-loading-resources) have their `always_load` relationships loaded.
+
+```ruby
+class CompanyResource
+  attribute :plan_start_at
+
+  always_load :subscription
+
+  def plan_start_at
+    @model.subscription.created_at
+  end
+end
+```
+
+Nested records to eager load can be declared.
+
+```ruby
+class CompanyResource
+  attribute :plan_start_at
+
+  always_load subscription: [:settings]
+
+  def plan_start_at
+    @model.subscription.created_at
+  end
+
+  def needs_tokens
+    @model.subscription.settings.needs_tokens
+  end
+end
 ```
 
 #### Resource Meta
--- a/lib/jsonapi/processor.rb
+++ b/lib/jsonapi/processor.rb
@@ -148,7 +148,7 @@ module JSONAPI
       fields = params[:fields]
 
       # TODO Should fetch related_resource from cache if caching enabled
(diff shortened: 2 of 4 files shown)


mrloop:
Test suite passing for 5.0 and 4.2 but is failing on rails 4.1. 4.1 is [no longer officially supported](http://rubyonrails.org/security/) and [isn't supported by railslts](https://railslts.com/). No unpatched [secuirty vunrebilities reported](https://groups.google.com/forum/#!forum/rubyonrails-security) yet but it's just a matter on time. Will 4.1 support be dropped or do I need to make this PR 4.1 compatible to be considered for merging?

lgebhardt:
@mrloop This seems to be in violation of the spec. From http://jsonapi.org/format/#fetching-includes:

> An endpoint MAY return resources related to the primary data by default.

Implies it's Ok, but it breaks a subsequent section:

> If an endpoint supports the include parameter and a client supplies it, the server MUST NOT include unrequested resource objects in the included section of the compound document.

I think we need to handle this as a `default_includes`. This would only be used if the request didn't specify any includes. I think this would satisfy compliance with the spec, and hopefully your needs.

mrloop:
Maybe the naming is a bit off and should be `always_load` as in `preload` or `eager_load` to better express the intent, that is to `MyActiveRecord.includes` related records so not to n+1 load db records when calculating attributes which rely on loading the related records and not includes in the jsonapi-resource sense, that is add the records to the payload.

lgebhardt:
@mrloop I see. I was getting hung up on the name an assuming the feature would extend to including the resources, not just the eager loading. I should have looked at the code and I would have seen how it was used - sorry. A name change would definitely be appreciated, especially since I hope to add the `default_includes` feature in the future. I'd also like to see some documentation in the readme so people know how to use this.

mrloop:
Have renamed `always_include` to `always_load` and similar methods. Documentation to follow.

mrloop:
@lgebhardt that's the renaming done and the documentation added.

NuckChorris:
Wouldn't this make more sense as `default_scope { includes(:thing) }` in the model?  I know "default scopes are bad", but it seems strange to put this in the Resource layer tbh and if you use it to compute virtual attributes it seems to be a reasonable use of a default scope

nathanpalmer:
@NuckChorris Something like this would be useful at the resource layer when you are exposing a property that would otherwise lazy-load a related model. I wouldn't want it to always load this data directly on the model since it's use-case is only for serialization at the resource.

mrloop:
~@lgebhardt I'm thinking that Resource is the incorrect location to define eager loading. Instead eager loading should be defined on Serializers. You could have different Serializers for the same Resource, each Serializer having different eager loading requirements. Doing eager loading from the serializer also allows you to take into account `fields` and reduce what is eager loaded based on fields. I'm going to spike out this approach.~
~1. Defer loading records until Serializer requires them.~
~2. Simple eager loading implement in Serializer similar to current PR~
~3. Have a look at fields and how these can be used.~
~My main goals would be to get 1 and 2 working. Would a PR along these lines be considered for merging?~

Edit, I don't think the above is correct, going to spike something to get a better idea of what's needed.

Genkilabs:
@mrloop what is the status of this PR? Did it work for you? We are being bitten by N+c queries loading records that need relations in their computed attributes.

mrloop:
@Genkilabs it's stalled, using similar version on code in small production app at the moment, working on another larger app at the moment using it as well, hope to be able to iron out any issues when that's deployed and in use, looks ok in testing.

Issue #774, opened 2016-07-21
Gem Load Error is: private method `include' called for JSONAPI::ResourceController:Class


sjauld:
Currently using 0.7.1.beta2
ruby 2.0.0-p645

```
Gem Load Error is: private method `include' called for JSONAPI::ResourceController:Class
Backtrace for gem load error is:
.../gems/jsonapi-resources-0.7.1.beta2/lib/jsonapi/acts_as_resource_controller.rb:10:in `included'
.../gems/jsonapi-resources-0.7.1.beta2/lib/jsonapi/resource_controller.rb:3:in `include'
.../gems/jsonapi-resources-0.7.1.beta2/lib/jsonapi/resource_controller.rb:3:in `<class:ResourceController>'
.../gems/jsonapi-resources-0.7.1.beta2/lib/jsonapi/resource_controller.rb:2:in `<module:JSONAPI>'
.../gems/jsonapi-resources-0.7.1.beta2/lib/jsonapi/resource_controller.rb:1:in `<top (required)>'
.../gems/activesupport-4.2.0/lib/active_support/dependencies.rb:274:in `require'
.../gems/activesupport-4.2.0/lib/active_support/dependencies.rb:274:in `block in require'
.../gems/activesupport-4.2.0/lib/active_support/dependencies.rb:240:in `load_dependency'
.../gems/activesupport-4.2.0/lib/active_support/dependencies.rb:274:in `require'
.../gems/jsonapi-resources-0.7.1.beta2/lib/jsonapi-resources.rb:5:in `<top (required)>'
.../gems/bundler-1.12.5/lib/bundler/runtime.rb:86:in `require'
.../gems/bundler-1.12.5/lib/bundler/runtime.rb:86:in `block (2 levels) in require'
.../gems/bundler-1.12.5/lib/bundler/runtime.rb:81:in `each'
.../gems/bundler-1.12.5/lib/bundler/runtime.rb:81:in `block in require'
.../gems/bundler-1.12.5/lib/bundler/runtime.rb:70:in `each'
.../gems/bundler-1.12.5/lib/bundler/runtime.rb:70:in `require'
.../gems/bundler-1.12.5/lib/bundler.rb:102:in `require'
```

Problem triggering it is:
https://github.com/cerebris/jsonapi-resources/blob/master/lib/jsonapi/acts_as_resource_controller.rb#L10

lgebhardt:
[...]

PR #853, opened 2016-09-27
Update minimum required Ruby version.

campreb:
According to issue #774, Ruby 2.0 appears to be no longer supported by this project. This is change is just to update the gemspec, so nobody accidentally installs it into a legacy project.


lgebhardt:
@campreb Thanks!

PR #856, opened 2016-09-29
Drop support for rails 4.1

lgebhardt:
With the release of Rails 5 version 4.1 has reached it's end-of-life and is no longer supported. http://weblog.rubyonrails.org/2016/6/30/Rails-5-0-final/

I'd like to drop support for Rails 4.1 as it adds testing complexity, time, and minor issues crop up that we need to work around.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: ruby
 sudo: false
 env:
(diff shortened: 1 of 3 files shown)


dgeb:
:+1:

PR #860, opened 2016-10-04
Remove rails as dependency

quinn:
(this is a work in process probably...)
- remove 'rails/all' from test helper
- vender minitest-spec-rails
- add activerecord and railties to gemspec

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -14,10 +14,10 @@ version = ENV['RAILS_VERSION'] || 'default'
 
 case version
 when 'master'
-  gem 'rails', { git: 'https://github.com/rails/rails.git' }
+  gem 'railties', { git: 'https://github.com/rails/rails.git' }
   gem 'arel', { git: 'https://github.com/rails/arel.git' }
 when 'default'
-  gem 'rails', '>= 5.0'
+  gem 'railties', '>= 5.0'
 else
-  gem 'rails', "~> #{version}"
+  gem 'railties', "~> #{version}"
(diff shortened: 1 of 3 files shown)


lgebhardt:
Thanks. I'd like to avoid needing to vendor minitest-spec-rails. Without it the tests seem to pass for rails 5, but not 4.2.

quinn:
@lgebhardt yeah that's reasonable. I wanted to confirm that nothing would break in the test suite if the other parts of rails were included (even after removing rails, it was all still being pulled in by minitest-spec-rails). 

Not sure what you mean by "without it", but I'll push again with the vendored stuff removed and add back into the gemspec

quinn:
(i tested again on 4.2 as well)

quinn:
@lgebhardt pushed. I'll squash these when it's ready.

lgebhardt:
Thanks. Looks ready to squash to me.

quinn:
@lgebhardt ok cool, done.

lgebhardt:
@quinn Thanks!

evolve2k:
Nice work folks :)

konk303:
why updated dependency of activerecord/railties to 4.1 instead of 4.0 ? 😢 
any particular reason?

I am using rails4.0 and noticed I couldn't update to 0.9.0.

using unsupported old rails is absolutely my bad I understand,
want to know if there is stg actually doesn't work with rails4.0

PR #863, opened 2016-10-05
Add "absolute" option for relationships

Rabsztok:
This option may be handy for anyone that tries to have namespaced resources, along with relations between them.

Example:
I have `MovieResource`, `Movie::ReviewResource` and `UserResource`.
Right now, when i try to add user relationship to `Movie::ReviewResource`, it fails, because jsonapi-resources searches for `Movie::UserResource` which do not exists.

With this PR, I am able to define my relation as:
`relationship :user, to: :one, class_name: 'User', absolute: true`
which results in using correct resource, in this case - `UserResource`

If you find this concept useful and able to be merged into your official gem, let me know. I will write tests / docs for it and I will prepare everything for merge.


thewalkingtoast:
Could you achieve the same effect with `class_name: '::User'`?

Rabsztok:
Unfortunately not.

The result is:
`JSONAPI: Could not find resource 'movie//users'. (Class Movie::::UserResource not found)`

lgebhardt:
Thanks, this seems useful. We need to ensure all the relationship features are fully working across namespaces.  Go ahead and finish up the tests and doc changes. Hopefully you won't run into any show stoppers. Let me know if you need any help.

The `:absolute` flag could be calculated from the presence of the `::` in the `class_name`.

Fivell:
more complete PR to solve this #893

PR #868, opened 2016-10-07
Allow for explicit resource_class on relationship definition

zentourist:
Allow for an explicitly defined `resource_class` on a relationship.

The use case is with namespace models and flat resources:

``` ruby
class Namespace::Form < ActiveRecord::Base
end

class TestFormResource < ::JSONAPI::Resource
  has_one :parent
end

class ParentResource < ::JSONAPI::Resource
  has_many :test_forms, resource_class: 'TestFormResource'
end
```

I was not able to find a way to have a flattened Resource for a namespaced model. If I namespaced the resource and that had a relationship back to the parent it tried to put the parent in the namespace of the form.

### DIFF
--- a/lib/jsonapi/relationship.rb
+++ b/lib/jsonapi/relationship.rb
@@ -23,7 +23,13 @@ module JSONAPI
     end
 
     def resource_klass
-      @resource_klass ||= @parent_resource.resource_for(@class_name)
+      @resource_klass ||=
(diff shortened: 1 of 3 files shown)


zentourist:
I just realized that `class_name` will work. I thought for some reason that this was for the AR model name. Sorry for my confusion.

PR #873, opened 2016-10-12
DOCS: Fix ResourceSerializer#initialize option

olleolleolle:
Spelling of an option to match what's in the code.


barelyknown:
Thanks for that!

PR #878, opened 2016-10-16
Allow users to use global, controller, or per action media type configuration

krainboltgreene:
Sometimes actions have specific media_types (Example: csv exports).

Sometimes entire resources have specific media_types (Example: Proxies).

Entire applications can have a specific media type that isn't the jsonapi mediatype (Example: https://github.com/json-api/json-api/issues/614#issuecomment-101460686 and https://github.com/json-api/json-api/issues/614#issuecomment-101741478, author Roy Fielding)

### DIFF
--- a/lib/jsonapi/acts_as_resource_controller.rb
+++ b/lib/jsonapi/acts_as_resource_controller.rb
@@ -139,7 +139,7 @@ module JSONAPI
     end
 
     def verify_content_type_header
-      unless request.content_type == JSONAPI::MEDIA_TYPE
+      unless request.content_type == conten_type_mediatype
         fail JSONAPI::Exceptions::UnsupportedMediaTypeError.new(request.content_type)
       end
       true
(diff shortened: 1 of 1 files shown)


lib/jsonapi/acts_as_resource_controller.rb
```diff
@@ -139,7 +139,7 @@ def resource_klass_name
     end
 
     def verify_content_type_header
-      unless request.content_type == JSONAPI::MEDIA_TYPE
+      unless request.content_type == conten_type_mediatype
```
tiagopog, line 142: Hey, perhaps I'm missing something but I couldn't figure out where `conten_type_mediatype` is/should be defined. 

There's also a typo (a missing "t") in `conten_type_mediatype`; maybe you'd better to name it to something like: `content_type_media_type`.

tiagopog: It's also important to note that JSON API spec describes that [only](http://jsonapi.org/format/#content-negotiation-servers) `"application/vnd.api+json"` is acceptable as media type for `Content-Type` (which is the value of `JSONAPI::MEDIA_TYPE`). So I think it'd be better to keep the old check.

scottgonzalez:
This is in violation of the JSON API specification. Please see https://jsonapi.org/format/#content-negotiation.

PR #880, opened 2016-10-18
Make exception backtraces configurable

jimsynz:
We discovered that our staging environments were leaking internals onto the web because JSONAPI::Resources assumed that as long as the environment wasn't `production` then it could include the backtrace in the serialized error output.  This PR adds a configuration option - with the same default behaviour as previous - however it allows the behaviour to be configured if needed.


vinithanatarajan (approved):

barelyknown:
Nice catch.

PR #882, opened 2016-10-19
Fix polymorphic type setter

Exelord:
Hey @lgebhardt!
We encounter a problem associated with the polymorphic relation. Here is an example in code:
## Setup

Models:

``` ruby
class Conversation < ActiveRecord::Base
  has_many :topics
end

class Account::Template < ActiveRecord::Base
  has_many :topics
end

class Topic < ActiveRecord::Base
  belongs_to :document, polymorphic: true
end
```

Resources:

``` ruby
class Template < JSONAPI::Resource
  model_name 'Account::Template'
  has_many :topics
end

class Topic < JSONAPI::Resource
  has_one :document, polymorphic: true
end
```
## Usage

Request: [ POST ] `/topics`

``` json
{
    "data": {
        "attributes": {
            "name": "Topic name",
        },
        "relationships": {
            "document": {
                "data": {
                    "type": "templates",
                    "id": "1"
                }
            },
        },
        "type": "topics"
    }
}
```
## Problem

Newly created topic has the wrong name of `document_type` filed.
After typing `topic.document_type` it was `Templates` but in fact, the polymorphic type should be taken from a model name and it should look like `Account::Template` because it's the real name of the resource's model's class name.
## Solution

I overwrite a method which assigns polymorphic type, and I created a method which looking for a real model class name of given relation's `key_type`.

I hope it will help :)

### DIFF
--- a/lib/jsonapi/resource.rb
+++ b/lib/jsonapi/resource.rb
@@ -324,7 +324,7 @@ module JSONAPI
       relationship = self.class._relationships[relationship_type.to_sym]
 
       _model.public_send("#{relationship.foreign_key}=", key_value)
-      _model.public_send("#{relationship.polymorphic_type}=", key_type.to_s.classify)
+      _model.public_send("#{relationship.polymorphic_type}=", _model_class_name(key_type))
 
       @save_needed = true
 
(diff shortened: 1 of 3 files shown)


barelyknown:
Nice catch. Can you add a test that fails without this fix and passes with it?

Exelord:
@barelyknown I added a test. I'm not familiar with the rails test framework (I used to use rspec) so feedback is welcome.

barelyknown:
Nice and simple - thanks!

Issue #879, opened 2016-10-17
relationships disappears when using fields param


Paprikas:
`post?include=comments` will return

```
data: [{
id
attributes
relationships
}]
included
```

`posts?include=comments&fields[posts]=name`
will return everything I'm asking for, except `relationships`

Help! =)

Paprikas:
Finally got it ;D
https://github.com/cerebris/jsonapi-resources/pull/883

PR #883, opened 2016-10-21
Note on relationships object

Paprikas:
See https://github.com/cerebris/jsonapi-resources/issues/879

### DIFF
--- a/README.md
+++ b/README.md
@@ -1036,6 +1036,9 @@ Will get you the following payload by default:
 }
 ```
 
+Note: When passing `include` with `fields` params, `relationships` object does not return.
+To include it, add `fields[posts]=comments` to request.
+
 #### Resource Meta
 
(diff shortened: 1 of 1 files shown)


lgebhardt:
We should probably add some code to, at a minimum, return an error stating that combination of `fields` and `includes` makes this request invalid. Better might be to inject the `include` into the `fields` array since that's probably what the requester intended.

In the mean time I agree we should have a note. What you have is technically incorrect: `relationships` will be returned if any relationship is specified in fields. How about a change like this:

```
Note: When passing `include` and `fields` params together, relationships not included in the
`fields` parameter will not be serialized. This will have the side effect of not serializing
the included resources. To ensure the related resources are properly side loaded specify them
in the `fields`, like `fields[posts]=comments,title,&include=comments`.
```

Paprikas:
Done @lgebhardt

PR #886, opened 2016-10-23
Improved error messages for invalid field and invalid type errors

jaggederest:
I ran into these errors and found them inscrutable, so I thought I would update the response strings with some suggestions about how to fix them.

We could extend this to other errors, as well, but I wanted to get feedback before I forged off into the wilderness too far.

Thanks, all feedback appreciated!

Justin

### DIFF
--- a/lib/jsonapi/exceptions.rb
+++ b/lib/jsonapi/exceptions.rb
@@ -244,7 +244,7 @@ module JSONAPI
                             title: I18n.translate('jsonapi-resources.exceptions.type_mismatch.title',
                                                   default: 'Type Mismatch'),
                             detail: I18n.translate('jsonapi-resources.exceptions.type_mismatch.detail',
-                                                   default: "#{type} is not a valid type for this operation.", type: type))]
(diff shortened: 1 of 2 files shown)


lgebhardt:
@jaggederest That sort of expanded error message seems good to me. Don't forget we need to change it in `locales/en.yml` as well.

jaggederest:
Will do, thanks for the feedback

lgebhardt:
@jaggederest I'm not sure if you've seen it, but one of the existing tests is failing due to the expanded error message:

```
Failure:
JSONAPIRequestTest#test_parse_dasherized_with_underscored_fields [/home/travis/build/cerebris/jsonapi-resources/test/unit/jsonapi_request/jsonapi_request_test.rb:141]:
--- expected
+++ actual
@@ -1 +1 @@
-"iso_currency is not a valid field for expense-entries."
+"iso_currency is not a valid field for expense-entries. Consider checking the spelling of the field name."
bin/rails test unit/jsonapi_request/jsonapi_request_test.rb:121
```

PR #895, opened 2016-10-30
Rescue JSON parsing error and concat JSONAPI exception.

Petesta:


Petesta:
@lgebhardt What do you think of this approach for handling incoming malformed JSON?

chadmetcalf:
@Petesta Instead of rescuing in `mime_types.rb`, appending to the params, and then checking the params hash, is it possible to rescue in the `request_parser.rb` similar to how the `ActionController::ParameterMissing` is rescued?

In addition, @lgebhardt Would it be possible to cherry pick this onto a 1.6 branch as well?

Thank you both for doing OSS. Big pats on the back all around.

Petesta:
@chadmetcalf You can't `rescue` inside of `request_parser.rb`. Once the program has started, `MimeTypes.install` is initialized because of this [call](https://github.com/cerebris/jsonapi-resources/blob/master/lib/jsonapi-resources.rb#L14) but the `lambda` has yet to be evaluated. So you can put additional statements above the `lambda` statement and on start you'll see those statements be evaluated but not the `lambda`. Once you've made a request with application type as JSON, the incoming request then forces, not sure if that's the best way of phrasing it, the lambda to execute thus blowing up on [`JSON.parse`](https://github.com/cerebris/jsonapi-resources/blob/master/lib/jsonapi/mime_types.rb#L22). With the parsing failure you never hit [initialize](https://github.com/cerebris/jsonapi-resources/blob/master/lib/jsonapi/request_parser.rb#L10) in `request_parser.rb` which in turn calls [setup_action](https://github.com/cerebris/jsonapi-resources/blob/master/lib/jsonapi/request_parser.rb#L27).

lgebhardt:
@Petesta Thanks for doing this, and sorry for the late reply. I swear I responded to this, but probably never finished the comment.

In general the approach looks good to me, though I'd like to see a few changes:
- In the case of a ParserError we know the problem is invalid JSON format. So we could return a more helpful error message. I'd like to see a new Error created with helpful messages. Ideally we shouldn't be leaking 500 errors if we can help it.
- We could catch all exceptions and return them in `{ _exception: e }`. Might be overkill if no other ones can arise - I haven't checked.
- We need some simple tests. For the tests these could be added to `test/integration/request_test`:

``` ruby
  def test_put_valid_json
    put '/posts/3', params: '{"data": { "type": "posts", "id": "3", "attributes": { "title": "A great new Post" } } }',
        headers: {
            'CONTENT_TYPE' => JSONAPI::MEDIA_TYPE,
            'Accept' => JSONAPI::MEDIA_TYPE
        }

    assert_equal 200, status
  end

  def test_put_invalid_json
    put '/posts/3', params: '{"data": { "type": "posts", "id": "3" "attributes": { "title": "A great new Post" } } }',
        headers: {
            'CONTENT_TYPE' => JSONAPI::MEDIA_TYPE,
            'Accept' => JSONAPI::MEDIA_TYPE
        }

    assert_equal 400, status
    assert_equal 'Something about invalid JSON', json_response['errors'][0]['title']
    assert_equal 'Maybe the parser error here', json_response['errors'][0]['detail']
  end
```
- I noticed we aren't handling arrays in top level JSON. This could be done as:

``` ruby
    def self.parser
      lambda do |body|
        begin
          data = JSON.parse(body)
          if data.is_a? Hash
            data.with_indifferent_access
          else
            fail JSONAPI::Exceptions::InvalidRequestFormat.new
          end
        rescue => e
          { _exception: e }
        end
      end
    end
```

And add a test:

``` ruby
  def test_put_valid_json_but_array
    put '/posts/3', params: '[{"data": { "type": "posts", "id": "3", "attributes": { "title": "A great new Post" } } }]',
        headers: {
            'CONTENT_TYPE' => JSONAPI::MEDIA_TYPE,
            'Accept' => JSONAPI::MEDIA_TYPE
        }

    assert_equal 400, status
    assert_equal 'Request must be a hash or some such thing', json_response['errors'][0]['title']
  end
```

Petesta:
@lgebhardt No worries.

Thanks for the feedback. I've addressed the bullet points you made and added tests. One thing to note is in `self.parser` you must return a hash in the `lambda` otherwise there will be an error. The value produced by the `lambda` has to always evaluate to a hash type. That's the reason for creating a hash instead of calling `fail`.

lgebhardt:
@Petesta Thanks! I'm fine with the way you are handling the two error cases in your update. I'm not sure if you're referring to the example code I provided. If so, I think it would catch the `fail` and pass it in the returned hash, as well as any other potentially raised exceptions (which may not exist). Or am I missing something else?

Petesta:
@lgebhardt Yea, in the example code you provided. The `fail` is not caught and returned into a hash. If you wrap the exception in a hash then `lambda` returns the correct type for whatever is evaluating the `lambda`.

``` ruby
if data.is_a?(Hash)
  hash.with_indifferent_access
else
  fail JSONAPI::Exceptions::InvalidRequestFormat.new
end
```

The `else` case was causing an error.

screen shot 2016-11-14 at 11 11 59 am

That's the reason for this.

``` ruby
{ _invalid_request_format: JSONAPI::Exceptions::InvalidRequestFormat.new }
```

lgebhardt:
@Petesta That's odd. Solving it is totally academic, but did you change the rescue as well?

Petesta:
@lgebhardt Is it odd? I don't think so. The return type is expected to be a hash. So you're not able to pass any other type in including any `JSONAPI::Exception` and array types. Even if you had valid JSON and tried to pass an array. The call would still fail. Technically array is a valid JSON type .

lgebhardt:
@Petesta What I'm saying is odd is the exception raised by `fail` should have been caught by the `rescue` at the end of the block and then returned back in `{ _exception: e }`. In either case a hash would be returned. I just tested it on my system and it worked as I expected.

Petesta:
@lgebhardt Ahh, I see what you mean. I misunderstood.

We call `fail` now and are explicit in our rescue calls in `self.parser`. If you rescued `JSON::ParserError`, you don't have method `errors` like all exceptions defined in `exceptions.rb`. I thought it would be better to continue with the pattern of calling `errors` on exceptions in `setup_action`.

Petesta:
@lgebhardt I'm not sure why the new changes cause old Ruby and Rails versions to fail here. Is there any additional logging on Travis? There's no stack trace. Just files executed. I wouldn't expect that change to cause failures now.

lib/jsonapi/exceptions.rb
```diff
@@ -119,6 +119,32 @@ def errors
...
+        @exception = exception
+      end
+
+      def errors
+        [JSONAPI::Error.new(code: JSONAPI::BAD_REQUEST,
+                            status: :bad_request,
+                            title: I18n.translate('jsonapi-resources.exceptions.bad_request.title',
+                                                  default: 'Bad Request'),
+                            detail: I18n.translate('jsonapi-resources.exceptions.bad_request.detail',
+                                                   default: @exception))]
```
lgebhardt, line 133: `default: @exception.to_s`

This should fix the type conversion error in rails 4

lgebhardt:
@Petesta Thanks! I made a small change to the way errors were returned which also fixed the issue with Rails 4. In addition I squashed your commits for ease of cherry picking into the beta branch.

Petesta:
@lgebhardt No problem and thanks for taking a look at it and merging it in! 😃

PR #904, opened 2016-11-14
Allow includes to follow namespacing

nagirrab:
Minor bug fix: there was a bug where if you do an `include` across a namespace boundary, e.g. in the case of:

    class BaseResource
      relationship :intermediate, to: :one, class_name: "Namespaced::Intermediate",
    end

    class Namespaced::IntermediateResource
       relationship :child, to: :one
    end
   
    class Namspaced::ChildResource
      relationship :intermediate, to: :one
    end

An includes request of `?includes=intermediate.child` would bomb out because it was using `BaseResource` rather than `Namespaced::IntermediateResource` for looking up `ChildResource`. This fixes that specific issue.

Not sure what your intentions with regards to namespaced resources are in general - I've had to put a few hacks in place to make things work in general, but this particular bug fix seems like it should go in right away.

### DIFF
--- a/lib/jsonapi/request_parser.rb
+++ b/lib/jsonapi/request_parser.rb
@@ -201,7 +201,7 @@ module JSONAPI
       relationship = resource_klass._relationship(relationship_name)
       if relationship && format_key(relationship_name) == include_parts.first
         unless include_parts.last.empty?
-          check_include(Resource.resource_for(@resource_klass.module_path + relationship.class_name.to_s.underscore), include_parts.last.partition('.'))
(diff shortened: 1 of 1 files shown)


lgebhardt:
@nagirrab Thanks!

The original intent was to have namespaces used for versioning, so it's not a surprise that you've had some issues mixing resources across namespaces. In the cases fully supported by JR the namespace of the @resource_klass and resource_klass should be the same. It's not going to be simple to get mixed namespaces fully supported, but it seems there is demand for it.

PR #905, opened 2016-11-15
Fix issue where resources received context nested in hash

austenito:
* If caching has been enabled, however disabled for a resource, the context values are nested in a a 'context' key. For example:

```
{ context: { foo: :bar } } 
vs. 
{ foo: :bar }
```

I wasn't sure how to test, so I ended up adding context to a resource. I'm open to suggestions on how you wanted this case covered.


lgebhardt:
@austenito Thanks!

Issue #889, opened 2016-10-26
Polymorphic relations do not work with fragment caching


DuaneCompton:
A simple polymorphic relationship like: 
 `class Api::V1::TagResource < BaseResource
     belongs_to taggable, polymorphic: true
 end`
and a  call like /api/v1/tags/1?include=taggable

will fail on any of the .joins(:ar_hash) i.e.
https://github.com/cerebris/jsonapi-resources/blob/master/lib/jsonapi/resource.rb#L1143
with the message:
`Can not eagerly load the polymorphic association :taggable`

Any ideas on how to avoid this. It seems like trying to not cache the polymorphic relations for now might be the best approach. However, this code appears to get hit regardless of if a resource is marked as cached or not. Which means that for simpler relations caching can not be enabled if any requests involve a poly. Any help would be appreciated. Thank you.

DavidMikeSimon:
@DauneCompton I'll look into this, thank you for the bug report

DavidMikeSimon:
@DuaneCompton Okay, I think my PR should fix this. Sorry about the delay.

DuaneCompton:
Fantastic! I'll give it a run tomorrow. Thank you again.

DavidMikeSimon:
@lgebhardt Everything here look okay to you?

PR #911, opened 2016-11-28
Bugfix: Partial preloading of cached fragments

DavidMikeSimon:
Fixes bug #889


lgebhardt:
@DavidMikeSimon Thanks!

PR #924, opened 2016-12-08
Public model_name_for_type method

Exelord:
Hi,
I just think this method could be used in many cases associated with polymorphic relations like filtering by polymorphic model type. In that case having this public could make it easier to implement.

### DIFF
--- a/lib/jsonapi/resource.rb
+++ b/lib/jsonapi/resource.rb
@@ -324,7 +324,7 @@ module JSONAPI
       relationship = self.class._relationships[relationship_type.to_sym]
 
       _model.public_send("#{relationship.foreign_key}=", key_value)
-      _model.public_send("#{relationship.polymorphic_type}=", self.class.real_model_name(key_type))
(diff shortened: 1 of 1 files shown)


lgebhardt:
@Exelord Looking at it a little closer the method should probably be a class method and moved down to the `self` section. I should have caught that in #882. How about renaming it to something closer to `resource_type_for`, which is it's reciprocal, such as `model_name_for_type` or just `model_name_for`.

```ruby
  class << self
  #...
      def model_name_for_type(type)
        type_class_name = type.to_s.classify
        resource = resource_for(type_class_name)
        resource ? resource._model_name.to_s : type_class_name
      end
  end
```

Exelord:
@lgebhardt You are right! I just forgot about moving it to `self`. I applied your feedback with method name and fixed tests :)

If u have more ideas or comment, go on!

Thanks :)

lgebhardt:
@Exelord Thanks!

toobulkeh:
How can this new feature be used?

Trying to see if it's something that can temporarily be used for #927

Issue #450, opened 2015-09-20
Unable to autoload constant Engine in LinkBuilder#build_engine_name


nsarno:
Hi guys,

The line **58** from `link_builder.rb` gives the following error (`Unable to autoload constant Engine`) when used with another gem containing an `engine.rb` file.

``` ruby
 "#{ scopes.first.to_s.camelize }::Engine".safe_constantize
```

To reproduce this issue:
- clone [duboff/peeps](https://github.com/duboff/peeps)
- run the server
- visit [/api/contacts](http://localhost:3000/api/contacts)

[This issue was first opened on this repo](https://github.com/nsarno/knock/issues/5), but I think it should actually be fixed here as this could happen with any rails engine having a  file named `engine.rb`.

In this case, the autoloader expect `lib/knock/engine.rb` to define an `Engine` constant, even though by convention it is namespaced as `Knock::Engine`.

Somehow I feel like this is actually a weird behaviour from the autoloader, but maybe it would still be good to add a rescue here anyway or something like this.

If you think there is actually something wrong in my code, don't hesitate to reopen the issue on my side and letting me know what you think.

Cheers!

christophermlne:
why is this issue closed? what was the resolution?

lgebhardt:
@christophermlne This issue is still open. It was the issue in the nsarno/knock project that is closed.

christophermlne:
my bad.

aldesantis:
Has this been fixed? Is there a workaround of any kind?

nsarno:
I'm not aware of anything yet. I'll try and see if I can submit a PR here in the coming weeks.

aldesantis:
A workaround for the time being is to create an initializer containing:

``` ruby
module JSONAPI
  class LinkBuilder
    private

    def build_engine_name
      scopes = module_scopes_from_class(primary_resource_klass)

      unless scopes.empty?
        begin
          "#{ scopes.first.to_s.camelize }::Engine".safe_constantize
        rescue LoadError => e
        end
      end
    end
  end
end
```

johnmosesman:
I ran into the same issue with:

```
rails 5.0.0.1
jsonapi-resources 0.8.1
knock 2.0
```

The fix by @alessandro1997 above worked.

gnfisher:
[...]

PR #926, opened 2016-12-13
Catch LoadError when checkin for engine

gnfisher:
resolves issue #450, solution from @alessandro1997 comment in discussion.

### DIFF
--- a/lib/jsonapi/link_builder.rb
+++ b/lib/jsonapi/link_builder.rb
@@ -57,8 +57,12 @@ module JSONAPI
     def build_engine_name
       scopes = module_scopes_from_class(primary_resource_klass)
 
-      unless scopes.empty?
-        "#{ scopes.first.to_s.camelize }::Engine".safe_constantize
+      begin
+        unless scopes.empty?
+          "#{ scopes.first.to_s.camelize }::Engine".safe_constantize
+        end
+      rescue LoadError => e
+        nil
       end
     end
 


gnfisher:
This is an autoloader issue. The Rails guide covers it [here](http://guides.rubyonrails.org/autoloading_and_reloading_constants.html#autoloading-algorithms). There may be another way to check if the controller is from an Engine or not without relying on LoadError to be thrown (going to look into this) but as the code stands, just wrapping this in a begin/rescue to catch it should be fine. If the controller is part of an engine, the proper engine paths and urls are formatted, and if not, the expected LoadError from the Rails autoloader is caught and non engine paths and urls are formatted.

Should I look for an alternative way to handle the Engine check, or is this above solution acceptable?

lgebhardt:
@gnfisher In general I'd like to avoid exceptions for flow control, however if raising `LoadError` is the way rails handles this I don't see a problem with your solution.

This passes the existing tests. I haven't looked into writing a test that confirms that this fixes the problem in #450. Is that something you could add? If it's not easy I can try to confirm using the steps in #450.

gnfisher:
When following the steps ins #450 no LoadError was raised, and the response processed correctly. I can look into writing a test for it - the current ones didn't catch the problem so perhaps there's a little retooling I can do to the existing suite to fix that. I will also do a bit of poking around for a solution that doesn't depend on catching Rails autloader's LoadError.

gnfisher:
I've been unable to craft a test for this. Somehow we'd have to use Rails autoloader to load the MyEngine and ApiV2Engine modules and their Engine classes... Admittedly I've not had a ton of time to put into this. If you could point me in a good direction I would be happy to look for a way to get this covered by the unit test.

lgebhardt:
@gnfisher I'm not sure how to test for it either. Since it's not easy to test I've merged it into master. If we see no errors soon I'll cherry pick it into beta and maybe 0.8 as well if I can get a test figured out. 

Thanks for the fix and looking at the tests.

Issue #68, opened 2015-01-27
Support for non-CRUD actions


hderms:
Hey, I had a use-case where I was trying to create a controller action which didn't fit neatly into CRUD actions. I'm wondering if there's a way to do this in jsonapi-resources without too much trouble. I'd like to take advantage of the standardized success/error response formatting, but don't want to play too much around with the internals to achieve this goal. Anyone have any ideas?

lgebhardt:
Have you tried adding the method to your controller?

hderms:
@lgebhardt yeah, I have. I'm thinking about stuff like what if my controller action raises an exception, it would be nice to have a way to make it render errors in the same format as jsonapi-resources, etc...

lgebhardt:
You could reuse the `handle_exceptions` method. For example:

``` ruby
class FoosController < JSONAPI::ResourceController

  def bar
    #handle bar stuff

    rescue => e
      handle_exceptions(e)
  end
end
```

You would need to make sure all your exceptions are derived from `JSONAPI::Exceptions::Error`, or you could override the `handle_exceptions` class to account for your custom exceptions.

7h3kk1d:
Similarly what's the standard for overriding a crud action and allowing other parameters. I have an object which takes file upload which obviously is using multipart form-data so I don't want to use the default controller action.

lgebhardt:
I think one approach that has potential would be to check the Content-Type header. If the Content-Type is `application/vnd.api+json` you could treat the request as a JSONAPI request, and otherwise you could process it as a file upload.

The content type is not currently being enforced through the controller, but #69 is almost ready.

Once #69 is merged you could try some code like (I have not tested this, so no promises):

``` ruby
class PostsController < JSONAPI::ResourceController
   skip_before_filter :setup_request, only: [:create]
   skip_before_filter :ensure_correct_media_type, only: [:create]

    def create
      if request.content_type == JSONAPI::MEDIA_TYPE
        setup_request
        process_request_operations
      elsif request.content_type == your type here
         #Handle file upload
      else
         raise JSONAPI::Exceptions::UnsupportedMediaTypeError.new(request.content_type)
      end
    end
end
```

xdmx:
Would it be possible to use all the goodness (attributes validation, errors, etc) of jsonapi-resources for different methods?

I have a `publish` method which behaves very similar to the `update` method, but it validates a different set of attributes and sets some other extra info in the model.

[...]

PR #928, opened 2016-12-14
Feature Custom Actions for resources

Exelord:
Hi,
I just moved our extension for custom actions to my fork. I know this behavior is not documented in JSONAPI standard but this will probably come:
 - https://github.com/json-api/json-api/issues/745
 - https://github.com/json-api/json-api/issues/684

So, I built a solution for that which can be later extended by the up-coming feature.
# Issues
#68

# Documentation
### custom_action(:name, options = {})
To define custom action you can use `custom_action` method in the resource.

As options you can define a hash with the given parameters:
- `type` - a type of the request eg, `:get`, `:post`, `:put` | default: `:get`
- `method` - a custom name of the action method
- `level` - `:instance` or `:collection` define on which level of routes the action should exist | default: `:instance`
- `apply` - a proc with the custom action logic (receive `model, context, data`)
```ruby
custom_action :publish, type: :post, apply: (lambda do |model, context, data|
    # your logic
end)
```  
- `includes` - you can specify default includes eg. `'comments,author.user'`

##### Instance actions
```ruby
class Api::V1::PostResource < JSONAPI::Resource
  custom_action :publish, type: :post, includes: 'publisher'

  def publish(data)
    # in `data` you can find passed parameters
    publisher = data.require(:publisher)
    @model.publish(publisher) # publsh a post
    # return a post [ActiveRecord model/collection] (or other class to return empty object)
    @model
  end
end
```

Make a request:
```
// POST { "data": { "publisher": "4" } }
https://api.app.com/posts/:id/publish
```
It will return published post with a included publisher.

##### Collection action
```ruby
class Api::V1::ProjectResource < JSONAPI::Resource
  custom_action :remove_all, level: :collection apply: (lambda do |model, context, data|
    #model is empty cuz it is a collection action
    nil if Post.destroy_all # return just an empty {} with 202 response
  end)
end
```

Make a request:
```
// GET { }
https://api.app.com/posts/remove-all
```
It will return empty `{}` with `202 Accepted` resposne.


Exelord:
@lgebhardt @barelyknown could you guys, give me some feedback and remarks?
Should I still continue work on it or just drop the idea?

lgebhardt:
@Exelord Sorry for the delay. I'll look it over soon.

nagirrab:
In case it's useful to the discussion, I built something similar, fully tested and working, see: 

https://github.com/nagirrab/jsonapi-resources/commit/72bd6359f3628c4d71660221dec73c083e9519b6

It has many of the same features but includes collection level actions as well as instance level. Instead of specifying a single method, you create another nested resource class to validate the parameters sent to the custom action the same way a PUT/POST call would be validated.

In the end, I refactored the offending methods into separate resources because I wasn't fully satisfied with the mechanic, but happy to resurrect it or help on someone else's version if the wind's blowing that way.

Looking at the mentioned spec discussion, it seems that it's more likely @Exelord's solution will fit the bill.

jufemaiz:
Would love to know if this is something that will progress?

Exelord:
It's still waiting for @lgebhardt comments.

jufemaiz:
Did @lgebhardt's comments disappear then?

Exelord:
Yep, @jufemaiz :/ Here is what he wrote:
> @Exelord I have some concerns about putting the new actions on the resources. I feel it's more consistent to place the implementation on the processors. However I recognize that the parsing of the request for a new action will be awkward since we still haven't given a decent way to extend the RequestParser.
I hope to have some time soon to look into this.

lgebhardt:
Sorry, I'm not sure why the comment disappeared, probably when I cancelled the review (just a few line comments), but that seems wrong. Glad you kept the email.

I've put #1010 together as a place to discuss changes so we can get your custom actions finished and not lock us into an architecture I'm not happy with.

Thanks for your patience with this.

Exelord:
Great, so as I understand, we can continue work on it and then change the architecture, is that correct?

Together with my team, we test it out on production since September or maybe longer. According to that, I have a few proposal that we can discuss.

## Custom actions in controllers
I suggest splitting `custom_actions` method in controllers to separate methods, eg: `invite`, `login`, `search` etc... or add a prefix or suffix to them to pretend accidental overwriting.

We have to think about `resource actions` and `models actions`. They can have the same name, so in that case we should really consider adding a prefix or suffix.

Spliting the action can help with testing, performance measurement tools, overwriting (if needed) and will increase readability of controller API (the actions logic will be explicit)

## Routes
I was thinking about covering a case for `resource actions`. In my opinion to declare them in `routes` will be sth like it was for my proposal:
```rb
jsonapi_resources :projects do
  jsonapi_custom_actions
end
``` 
In my opionin `jsonapi_custom_actions` method should generate a routes for `models actions` and also for `resource actions`. What do you think? Or we should force `jsonapi_resources` method to handle that by default?

_______
I will add more ideas and concerns after the architecure of this API's.

Exelord:
Hi @lgebhardt,
Together with @poslinskiNet we:
- upgraded the PR to the latest version
- updated the code
- added the code documentation
- updated PR documentation
- added a support for custom actions on collection level

We are also in progress with tests (almost at the end - they will show up in a few days)
Feedback is welcome... We spend a lot of time on this so we hope it will follow us to some point.

PoslinskiNet:
@Exelord @lgebhardt I've pushed the specs and some improvements. I hope it will speed up the integration of the code. If you have any remarks, just let me know. Enjoy your weekend 👍

chrism:
This is really brilliant and something that would be incredibly useful on one of my projects.

Thank you for all the hard work on it.

Is there any way to help get this merged and released?

Thanks!

Exelord:
Maybe @barelyknown or @senid231 could help :)

lgebhardt:
This isn't following the flow of the existing actions. For example `setup_custom_actions_action` is calling the `apply` method for the action. `apply` shouldn't be called until the operation is being processed. The setup should create an `Operation` object which will then be processed by the `Processor`. Using a custom `Processor` and moving the logic from the `apply` callable would fit better with the existing architecture.

I propose that we allow the creation of custom `RequestParser`s for a controller (or set by Resource). That will create a custom `Operation` which is processed by a custom `Processor`.  The response would be handled with a custom `OperationResult`, if needed, which handles the conversion to JSON.

The new piece needed would be the new `RequestParser` support. I think the existing rails route methods would work for setting up the routes. This is a much smaller set of changes and would leverage the existing framework.

Custom actions need to be able to set whether they are transactional. Now that's done by method name, but it might be better to handle it on the `Operation`.

Support for a custom media type for extension requests would be nice, but probably not necessary.

chrism:
Maybe this isn't the appropriate place for this, but I've been unable to find any information elsewhere and it relates directly so I'd just like to ask.

Without a solution such as this what is the current best practice for working with a specific non-RESTful API endpoint?

I've been adding more conditional logic into the update method in a controller, but it feels like a really bad approach and is very hard to test.

PoslinskiNet:
Hello @chrism,

In fact, the idea of custom actions addresses your case and gives you some flexibility over responses (it can return as response any type of resource/resources - it is up to you). It also simplifies the logic, because you don't have to have some conditionals - instead you can have multiple custom action endpoints which contain for example service object that can be easily testable.

chrism:
Hi @PoslinskiNet,

Thanks very much for responding. Yes absolutely, that is what I understood by the idea and would be absolutely perfect for my use cases, like you have kindly explained.

I was wondering though, until this PR is merged to the satisfaction of @lgebhardt is there a reasonable way to do this with the current release that is more 'best practice' than my 'hack'?

PoslinskiNet:
Would be nice to have it some day integrated to the main repository. Maybe someone is open for taking specs that are already in-place and do some adjustments to the implementation so It can be merged eventually by the gem owner (we can help, but we don't have enough time to rewrite it from scratch)?

Keep in mind that It doesn't break the current version of **jsonapi-resources**, it just extends it with pretty cool behaviour 😈 .

lib/jsonapi/custom_actions/extensions/processor.rb
```diff
@@ -0,0 +1,25 @@
...
+    # Processing custom actions results for single model
+    # It will handle action result if returns one single model
+    #
+    # @return [ResourceOperationResult]
+    def custom_actions_instance
+      resource = resource_klass.resource_for(params[:result], context)
+      JSONAPI::ResourceOperationResult.new(:ok, resource, result_options)
+    rescue
+      JSONAPI::OperationResult.new(:accepted, result_options)
+    end
```
ArneZsng, line 14: Since the internal API of JSONAPI::Resources has changed quite a bit since this PR was opened, I think this method needs a little update.

One proposal:
```suggestion
        # Processing custom actions results for single model
    # It will handle action result if returns one single model
    #
    # @return [ResourceSetOperationResult]
    def custom_actions_instance
      resource = resource_klass.resource_for(params[:result], context)

      include_directives = params[:include_directives]
      fields = params[:fields]
      serializer = params[:serializer]

      find_options = {
        context: context,
        fields: fields,
        filters: { resource_klass._primary_key => resource.id }
      }

      resource_set = find_resource_set(resource_klass,
                                       include_directives,
                                       serializer,
                                       find_options)

      JSONAPI::ResourceSetOperationResult.new(:ok, resource_set, result_options)
    rescue
      JSONAPI::OperationResult.new(:accepted, result_options)
    end
```

lib/jsonapi/custom_actions/extensions/processor.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    # Processing custom actions results for many models
+    # It will handle action result if returns array or ActiveRecord::Relation
+    #
+    # @return [ResourcesOperationResult]
+    def custom_actions_collection
+      resources = resource_klass.resources_for(params[:results], context)
+      JSONAPI::ResourcesOperationResult.new(:ok, resources, result_options)
+    end
```
ArneZsng, line 23: Same as above, this can use an update.

```suggestion
        # Processing custom actions results for many models
    # It will handle action result if returns array or ActiveRecord::Relation
    #
    # @return [ResourceSetOperationResult]
    def custom_actions_collection
      resources = resource_klass.resources_for(params[:results], context)

      include_directives = params[:include_directives]
      fields = params[:fields]
      serializer = params[:serializer]

      find_options = {
        context: context,
        fields: fields,
        filters: { resource_klass._primary_key => resources.map(&:id) }
      }

      resource_set = find_resource_set(resource_klass,
                                       include_directives,
                                       serializer,
                                       find_options)

      JSONAPI::ResourceSetOperationResult.new(:ok, resource_set, result_options)
    rescue
      JSONAPI::OperationResult.new(:accepted, result_options)
    end
```

ArneZsng (changes requested):
As we are using this PR in one of our forks, I just wanted to share an update on the code we needed to incorporate to make it work for us.

PR #931, opened 2016-12-16
Allow to access results of the response document

Exelord:
Hi @lgebhardt,
I added the results of `process_request` to an instance variable. Thanks to that we can write better controllers tests using assertions with `resource` instances instead of parsing JSON string.

### DIFF
--- a/lib/jsonapi/response_document.rb
+++ b/lib/jsonapi/response_document.rb
@@ -1,9 +1,10 @@
 module JSONAPI
   class ResponseDocument
-    attr_reader :serialized_results
+    attr_reader :serialized_results, :results
 
     def initialize(options = {})
       @serialized_results = []
+      @results = []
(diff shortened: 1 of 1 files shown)


lgebhardt:
@Exelord I agree that exposing the results is an improvement, but it conflicts with the multiple operations work I'm in the middle of. This requires each operation to be serialized after processing so subsequent operations can use JSON pointers to the prior results.

Exelord:
Hi @lgebhardt,
I applied the implementation to the current master, what do you think about it now?

Exelord:
ping?

lgebhardt:
I'm still waiting for some clarity on the operations direction. I don't want to merge if it's going to create issues down the road. But I think it's likely we won't be using json pointers, so I can possibly restructure #951.

PR #939, opened 2016-12-21
Support Objects For Filter Apply and Verify, Tests

HoyaBoya:
This is a simpler and less ambitious PR but very much in the spirit of #784 to enhance the filter functionality to be more consistent with Rails.

The current documentation encourages the use of lambdas for implementation of `apply` and `verify`. This has led to some complaints in our code coverage and style tools. Digging around the code base, we noticed the support of symbol / String callbacks as well.

This PR introduces a third way, an object, to be used for `apply` and `verify`. This makes this feature more like [Rails constraints](http://guides.rubyonrails.org/routing.html#advanced-constraints) or [AR validations](http://guides.rubyonrails.org/active_record_validations.html#custom-validators) where an object can be passed in if it adheres to some notion of an object contract, along with symbols, strings, and procs / lambdas.

This should make things a bit more testable and re-usable.

Let me know your thoughts. Happy to follow up with documentation if we feel this is the right direction. I'm also happy to let this MVP make way for #784 if we can unblock it somehow.


lgebhardt:
@HoyaBoya #784 was almost ready to go but I wanted to get some tests created, and in the process review the assumptions it's making. Then I got busy with other things and didn't get back to it. So I'm in favor of finishing up #784.

HoyaBoya:
There's only about 4 lines of overlap here and #784, with the real meat being in #784. I was going to try and apply that PR into here and expand on the test coverage I have some can get this across the finish line.

https://github.com/cerebris/jsonapi-resources/pull/784/files#diff-e6cf31c573bd5f2277b1f16812943df4R702

And then under `elsif respond_to?` `send` for the `verify` call.

Does that sound like a worth while endeavor?

lgebhardt:
@HoyaBoya Lets move this discussion over to #784. There are a few changes I'd like to see in that PR as well as the tests.

HoyaBoya:
Sounds good. Closing this in favor of #784 and helping there.

PR #941, opened 2016-12-21
Use assert_nil instead of assert_equal

scottgonzalez:
Removes lots of deprecation warnings when running tests.


lgebhardt:
@scottgonzalez Thanks! I was planning do this today.

PR #942, opened 2016-12-22
Allow context to be passed through to filter calls

nagirrab:
Hey,

Small fix: I wanted the ability to run filters on the basis of context (e.g. "Show me all records belonging to the current user").

This should be fairly isolated and unlikely to cause other problems.

Let me know if you have any questions about it.


lgebhardt:
@nagirrab Thanks!

PR #944, opened 2016-12-23
Rename local variable include

olleolleolle:
This PR renames a local variable in one of the methods concerning `include`s, the JSONAPI concept.

This avoids collision with the Ruby keyword, which can be confusing for the reader.

(Also: use `#map` to create a list.)

(_I came to read #937 and was a little confused, at first_.)

### DIFF
--- a/lib/jsonapi/request_parser.rb
+++ b/lib/jsonapi/request_parser.rb
@@ -211,7 +211,7 @@ module JSONAPI
     end
 
     def parse_include_directives(raw_include)
-      return if raw_include.nil?
+      return unless raw_include
 
       unless JSONAPI.configuration.allow_include
         fail JSONAPI::Exceptions::ParametersNotAllowed.new([:include])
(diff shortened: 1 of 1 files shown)


lib/jsonapi/request_parser.rb
```diff
@@ -210,23 +210,22 @@ def check_include(resource_klass, include_parts)
       end
     end
 
-    def parse_include_directives(include)
-      return if include.nil?
+    def parse_include_directives(include_directives)
```
lgebhardt: The name `include_directives` might be confused with the parsed `@include_directives`. Maybe rename to raw_include, include_param or something similar.

olleolleolle: Thanks, will edit right now.

olleolleolle:
@lgebhardt There. The `raw_` prefix indicates "user input".

olleolleolle:
I also note that we'd be better off doing `return unless raw_include` since

`CSV.parse_line(false)` would raise an error.

lgebhardt:
@olleolleolle Thanks!

PR #945, opened 2016-12-23
Fix has_one polymorphism

nagirrab:
Problem: `has_one` relationships that are polymorphic do not work as expected: instead of getting the actual polymorphic resource, you get instead the base class.

Fix: extend the handling used in the polymorphic `belongs_to` relationship to cover the `has_one` case as well. I used the exact same code as the block above.

Tested on Rails 4.2.x.

Let me know if you have any questions or concerns.

### DIFF
--- a/lib/jsonapi/relationship_builder.rb
+++ b/lib/jsonapi/relationship_builder.rb
@@ -120,10 +120,16 @@ module JSONAPI
       define_on_resource relationship_name do |options = {}|
         relationship = self.class._relationships[relationship_name]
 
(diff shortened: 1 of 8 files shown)


lgebhardt:
@nagirrab Thanks for the fix. Could you add a test that demonstrates the issue and confirms the fix?

nagirrab:
@lgebhardt done. The test case is a little wonky because has_one polymorphism through other relations is actually not trivial to get working, but hopefully it illustrates the point.

lgebhardt:
@nagirrab Thanks!

PR #946, opened 2016-12-29
Refactor for operations support

lgebhardt:
This PR is preparing for the support of multiple operations in a request. A separate PR to add that support will be coming shortly.

- All base spec requests now are handled in a single operation:
This is needed to have a one to one mapping of responses to operations.

- OperationResults functions moved to the ResponseDocument:
This change is needed to allow subsequent operations to reference the results of prior operations through the use of JSON pointers. For this to function the results must be serialized after processing and the RequestParser will need access to the prior results to resolve pointers. So the need for an intermediate storage of operation results is no longer needed.

- OperationDispatcher functions moved to ActsAsResourceController.


dgeb:
Great work, @lgebhardt! This should be a solid foundation for processing multiple operations.

PR #950, opened 2017-01-04
Add :readonly option to attributes and relationships

hidde-jan:
This PR adds the :readonly flag to both the attribute method and the relationship method on JSONAPI::Resource. Flagging an attribute or relationship as readonly prevents it from being included in either the `creatable_fields` or `updatable_fields`.

It is a second attempt at PR #239. Compared to #239, it adds the readonly flag for both the `attribute` helper and the `relationship` helper. A further difference is that this implementation does not remove the `#{attr}=` method from the Resource, since IMO a readonly flag should only influence whether an attribute or relationship is or isn't in the `creatable_` and `updatable_fields`.

This is an updated approach to #949 with a much smaller set of changes.

### DIFF
--- a/lib/jsonapi/relationship.rb
+++ b/lib/jsonapi/relationship.rb
@@ -60,6 +60,10 @@ module JSONAPI
       false
     end
 
+    def readonly?
+      @options[:readonly]
+    end
+
     class ToOne < Relationship
       attr_reader :foreign_key_on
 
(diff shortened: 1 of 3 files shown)


lgebhardt:
@hidde-jan Thanks!

hidde-jan:
No problem :) It was something I've ran into a couple of times and it was a quick fix.

mattharris5:
Are there plans to add this to a release soon or will it be in `0.10`?

PR #962, opened 2017-01-22
Prevent error if using paginator :none

fsmanuel:
The PR fixes a bug if you use `paginator :none` and have `config.top_level_meta_include_page_count` set to `true`.


travisofthenorth:
+1 this is a current issue for my team and we have to use a custom paginator to get around it

lgebhardt:
@fsmanuel Thanks!

PR #966, opened 2017-01-24
unify checking for @fields if nil or empty with active support blank,…

ristovskiv:
… in resource_serializer#requested_fields

This is a very minimal change. Please don't take it as an offense I don't want to change anything in other developers code (especially someone more experienced than me). I was just reading through the code and thought that checking  `@fields.nil? || @fields.empty?` is unnecessary when you already have `ActiveSupport` with his `blank?`.


lgebhardt:
@ristovskiv Thanks!

PR #969, opened 2017-01-31
Fix wording for missing model

scottgonzalez:


lgebhardt:
@scottgonzalez Thanks!

PR #971, opened 2017-02-01
Verify filters that are passed to show_related_resources

### LINKED ISSUE #964: Filters are not verified for show_related_resource(s)
This is a problem since

>  Verify should return the verified value, which may be modified.
> http://jsonapi-resources.com/v0.8/guide/resources.html#Filters

Doing a request to `/api/v1/some-resource?filter[beer]=valid_but_needs_to_be_modified` might succeed while `/api/v1/parents/1/some-resource?filter[beer]=valid_but_needs_to_be_modified` fails.

hidde-jan:
Fixes #964 

I've also removed the `source_klass.verify_filters` since the source resource shouldn't really know about the filters from another resource.


lgebhardt:
@hidde-jan Thanks!

PR #974, opened 2017-02-07
Filter nil values from include param

hidde-jan:
Sending `?include=,,` results CSV parsing `[nil, nil, nil]`, which in turn trows an internal server error.

This PR fixes the problem by silently ignoring empty include directives.


lgebhardt:
@hidde-jan Thanks!

PR #975, opened 2017-02-07
apply_filter and apply_sort respects :delegate option of attribute

senid231:
Allows to sort and filter by delegated attribute without additional patching.
For example request like
```ruby
get '/posts', params: {
    filter: {name: 'some_title'}, 
    sort: '-name', 
    fields: {posts: 'name'}
}
```
 for such resource
```ruby
  model_name 'Post'

  attribute :name, delegate: :title
  attribute :body

  filter :name
end
```
will responds successfully with correct data.

### DIFF
--- a/lib/jsonapi/active_record_accessor.rb
+++ b/lib/jsonapi/active_record_accessor.rb
@@ -225,6 +225,7 @@ module JSONAPI
               order_by_query = "#{associations.last.name}_sorting.#{column_name} #{direction}"
               records = records.joins(joins_query).order(order_by_query)
             else
+              field = _resource_klass._attribute_delegated_name(field)
               records = records.order(field => direction)
             end
           end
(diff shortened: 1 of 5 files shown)


senid231:
maybe we should also change keys in `model_error_messages`

senid231:
@lgebhardt, @dgeb, @barelyknown, what do you think about this?

hidde-jan:
The fact that the two code changes are exactly the same shows that there needs to be a better solution to mapping resource attributes to model attributes and vice versa. Like @senid231 mentions, a proper solution should probably also take care of validation messages (and the corresponding pointers). Something on a class level like

```ruby
class ProductResource < JSONAPI::Resource
  attribute :name, delegate: :title
  attribute :price

  def self.model_attribute_name(resource_attribute)
    # :name => :title | :price => :price
  end

  def self.resource_attribute_name(model_attribute)
    # :title => :name | :price => :price
  end
end
```

senid231:
rebased and dried
@lgebhardt please take a look

lgebhardt:
@senid231 Thanks!

PR #980, opened 2017-02-15
create bug report templates

senid231:
Add templates for bug reporting like [Rails does](https://github.com/rails/rails/tree/master/guides/bug_report_templates).
This will help to reproduce and debug issues with ease


senid231:
@lgebhardt, @dgeb, @barelyknown, any thoughts?

Fivell:
@lgebhardt , I think this can be good improvement for faster feedbacks

lgebhardt:
@senid231 Thanks, this should help a lot!

PR #983, opened 2017-02-20
fix relationship linkage when it has overridden primary_key

### LINKED ISSUE #982: bug with polymorphic relationship
when primary_key is changed on resource level id mismatch between relatioship linkage and included data.
Reproduced in [bug report template](https://gist.github.com/senid231/dd6a13fd968a5cc926578deb65199ef7) according to PR #980 proposal

Models
```ruby
class Storage < ActiveRecord::Base
  has_one :keeper, class_name: 'Keeper', as: :keepable
end

class Keeper < ActiveRecord::Base
  belongs_to :keepable, polymorphic: true
end
```

resources
```ruby
class StorageResource < JSONAPI::Resource
  model_name 'Storage'
  key_type :string
  primary_key :token

  attribute :name
end

class KeeperResource < JSONAPI::Resource
  model_name 'Keeper'
  has_one :keepable, polymorphic: true

  attribute :name
end

class KeepableResource < JSONAPI::Resource
end
```

Data
```ruby
storage = Storage.create! name: 'Warehouse 1', token: 'some-token'
Keeper.create! name: 'John Doe', keepable: storage
```

Response
```
{
  "data": {
    "id": "1",
    "type": "keepers",
    "links": {
      "self": "http://example.org/keepers/1"
    },
    "attributes": {
      "name": "John Doe"
    },
    "relationships": {
      "keepable": {
        "links": {
          "self": "http://example.org/keepers/1/relationships/keepable",
          "related": "http://example.org/keepers/1/keepable"
        },
        "data": {
          "type": "storages",
          "id": "1"
        }
      }
    }
  },
  "included": [
    {
      "id": "some-token",
      "type": "storages",
      "links": {
        "self": "http://example.org/storages/some-token"
      },
      "attributes": {
        "name": "Warehouse 1"
      }
    }
  ]
}
```

as you see included data for storages has `"id: "some-token"`, but `keepable` relationship linkage has ` "id": "1"`

senid231:
fixes #982

### DIFF
--- a/lib/jsonapi/relationship.rb
+++ b/lib/jsonapi/relationship.rb
@@ -64,6 +64,10 @@ module JSONAPI
       @options[:readonly]
     end
 
+    def redefined_pkey?
+      belongs_to? && primary_key != resource_klass._default_primary_key
+    end
+
     class ToOne < Relationship
       attr_reader :foreign_key_on
 
--- a/lib/jsonapi/resource.rb
+++ b/lib/jsonapi/resource.rb
@@ -783,7 +783,11 @@ module JSONAPI
       end
 
       def _primary_key
-        @_primary_key ||= _model_class.respond_to?(:primary_key) ? _model_class.primary_key : :id
+        @_primary_key ||= _default_primary_key
+      end
+
+      def _default_primary_key
+        @_default_primary_key ||=_model_class.respond_to?(:primary_key) ? _model_class.primary_key : :id
       end
 
       def _cache_field
--- a/lib/jsonapi/resource_serializer.rb
+++ b/lib/jsonapi/resource_serializer.rb
@@ -454,7 +454,7 @@ module JSONAPI
     def foreign_key_value(source, relationship)
       related_resource_id = if source.preloaded_fragments.has_key?(format_key(relationship.name))
         source.preloaded_fragments[format_key(relationship.name)].values.first.try(:id)
-      elsif source.respond_to?(relationship.foreign_key)
+      elsif !relationship.redefined_pkey? && !relationship.polymorphic? && source.respond_to?(relationship.foreign_key)
         # If you have direct access to the underlying id, you don't have to load the relationship
         # which can save quite a lot of time when loading a lot of data.
         # This does not apply to e.g. has_one :through relationships.
--- a/test/fixtures/access_cards.yml
+++ b/test/fixtures/access_cards.yml
@@ -1,0 +1,4 @@
+john_doe_worker_card:
+  id: 1
+  token: "some-token"
+  security_level: "admin"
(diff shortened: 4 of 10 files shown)


senid231:
Should i provide test that reproduces this case? 
I think the right way is to provide PR with test that will fail without this PR, but with current JR tests it's complicated.

lgebhardt:
@senid231 I agree having a test that fails without the fix is ideal. If you can add on that would be great, or I can try to get one setup based on the issue you reported.

senid231:
PR #760 creates a regression for relationships with overridden primary_key
test added that fails without this fix

for polymorphic relationships reproduced [here](https://gist.github.com/senid231/dd6a13fd968a5cc926578deb65199ef7)
for to_one belongs_to relationships reproduced [here](https://gist.github.com/senid231/4face22d00f342b22fcddbaf6e59466e)

@lgebhardt please take a look

chrism:
I've just come across this issue myself and this PR would appear to fix it.

👍

lgebhardt:
@senid231 Thanks!

clemthenem:
Hi!
I think there is still an issue with the `foreign_key_value(source, relationship)` function in `JSONAPI::ResourceSerializer`.
I get the following error
```bash
Internal Server Error: wrong number of arguments (given 1, expected 0) /home/***/.rvm/gems/ruby-2.4.1/gems/activerecord-5.0.3/lib/active_record/attribute_methods/read.rb:34:in `__temp__d6564786f646'
/home/***/.rvm/gems/ruby-2.4.1/bundler/gems/jsonapi-resources-03807f4f09db/lib/jsonapi/active_record_accessor.rb:93:in `foreign_key'
/home/***/.rvm/gems/ruby-2.4.1/bundler/gems/jsonapi-resources-03807f4f09db/lib/jsonapi/resource.rb:963:in `block in build_belongs_to'
/home/***/.rvm/gems/ruby-2.4.1/bundler/gems/jsonapi-resources-03807f4f09db/lib/jsonapi/resource_serializer.rb:462:in `public_send'
/home/***/.rvm/gems/ruby-2.4.1/bundler/gems/jsonapi-resources-03807f4f09db/lib/jsonapi/resource_serializer.rb:462:in `foreign_key_value'
/home/***/.rvm/gems/ruby-2.4.1/bundler/gems/jsonapi-resources-03807f4f09db/lib/jsonapi/resource_serializer.rb:383:in `to_one_linkage'
```

```ruby
class User  < ApplicationRecord
   has_many :verifications, class_name: 'Account::Verification', foreign_key: 'user_id'
end
```

```ruby
class Account::Verification  < ApplicationRecord
   has_one :user
end
```

If I use 
```ruby
source.public_send(relationship.name).try(:id)
```
instead of
```ruby
source.public_send(relationship.foreign_key)
```

I am not sure if the configuration of uuids matters for this problem but I am using them.
`config.resource_key_type = :uuid`

I hope all this makes sense, let me know if you cannot reproduce the error.
Thanks for your hard work!

PR #988, opened 2017-02-22
Use plural of resource type for controller

hidde-jan:
Currently, doing `jsonapi_resource :current_user` will use the `CurrentUserController`, however, any relationships on the `current_user` resource will instead look for the `CurrentUsersController`. By default, rails uses the plural name controller for any singular resource. Keeping this the same for JA:R seems like a good idea.

This is currently not picked up by the tests currently, since `Preferences` is already a plural.

### DIFF
--- a/lib/jsonapi/request_parser.rb
+++ b/lib/jsonapi/request_parser.rb
@@ -152,7 +152,7 @@ module JSONAPI
           resource_klass,
           context: @context,
           relationship_type: relationship_type,
-          parent_key: resource_klass.verify_key(parent_key)
+          parent_key: resource_klass.verify_key(parent_key, @context)
       )
     end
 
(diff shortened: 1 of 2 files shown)


hidde-jan:
Actually, relationships for singular resources also seem broken (at least in 0.8.3). If I `GET` a related resource of a `current_user` (say, `events`), then I get the following error:

```json
{
    "errors": [
        {
            "code": "PARAM_MISSING",
            "detail": "The required parameter, current_user_id, is missing.",
            "status": "400",
            "title": "Missing Parameter"
        }
    ]
}
```

Edit: I did some more checking. The only way this can currently be worked around is to trick the RequestParser into thinking a valid key was passed, i.e. by overriding `CurrentUserResource._as_parent_key` and returning `:source`, then overriding `verify_key`:

```ruby
class CurrentUserResource < JSONAPI::Resource
  def self._as_parent_key
    :source
   end

  def self.verify_key(key, context = {})
    if key == "current_users" && context[:current_user]
      context[:current_user].id
    else
      super
    end
  end
end
```

Edit 2: `:source` also doesn't work, because it is not used for `show_relationship`. The only way I can make it work is by setting `_as_parent_key` to `:action` and then just always returning the id of the `current_user` in the `context`.

amiel:
I'm running against this issue as well. Thanks for the details :)

PR #994, opened 2017-03-01
Store exceptions in request env

travisofthenorth:
I know that there is an exception whitelist, and that all exceptions can be whitelisted in the latest version with the `whitelist_all_exceptions` configuration. IMO, this is tangential issue, because as a developer, I might want the gem to handle an exception for me as far as responding to users with error messages, but I _also_ want this exception to be logged separately.

For example, I rely on a Rollbar middleware to report exceptions for alerting/monitoring of my services. Because the gem was swallowing exceptions (in this case a database timeout, which should absolutely sound the alarms), I wasn't aware of a critical app error. Again, I still want the gem to handle responding to users with error messages, but I want those server errors to be processed by the rest of my middleware chain.

### DIFF
--- a/lib/jsonapi/acts_as_resource_controller.rb
+++ b/lib/jsonapi/acts_as_resource_controller.rb
@@ -284,6 +284,9 @@ module JSONAPI
               }
             end
 
+            # Store exception for other middlewares
+            request.env['action_dispatch.exception'] ||= e
+
             internal_server_error = JSONAPI::Exceptions::InternalServerError.new(e)
             Rails.logger.error { "Internal Server Error: #{e.message} #{e.backtrace.join("\n")}" }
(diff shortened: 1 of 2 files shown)


travisofthenorth:
For reference, this is the Rollbar middleware I have in mind: https://github.com/rollbar/rollbar-gem/blob/master/lib/rollbar/middleware/rails/rollbar.rb#L26-L28

This is also what the `ActionDispatch` middleware sets, hence why Rollbar looks for it: https://github.com/rails/rails/blob/master/actionpack/lib/action_dispatch/middleware/show_exceptions.rb#L46

HoyaBoya:
I think this is a great PR. 👍

lgebhardt:
@travisofthenorth Thanks!

PR #1001, opened 2017-03-14
Use foreign_key option when getting foreign key value.

dougo:
I noticed that I was getting N+1 queries when using `always_include_to_one_linkage_data`. I tracked it down to the `foreign_key_value` method, which does try to use the foreign key without loading the associated object, but it was hardcoded to use `foo_id` rather than using the relationship's `:foreign_key` option if provided. Pretty straightforward fix.

I also added a unit test for this logic, because I couldn't find any obvious way to modify an existing test to cover this. Let me know if you have a better idea (or if you think it doesn't need to be tested).


lgebhardt:
@dougo Thanks!

PR #1002, opened 2017-03-15
Give resource more control over sortable fields

rfb:
I'm in a situation where I want clients to have repeatable, random sort
order on a collection. My solution is to pass a ?sort=rand-N, with N
being the seed used for sorting. This way, if the client was to page
through the list or link to the list, the sort order would be preserved.

For example, `/providers?sort=rand-42` would apply a random sort order
with 42 as the key.

To accommodate this, I need JSONAPI::RequestParser to allow sort keys
that are not in a predetermined list. I think a good solution to this
would be to have the RequestParser defer to the Resource when checking
if a sort key is valid.

The existing behaviour is maintained by moving the check on
sortable_fields to sortable_field?, which can be overloaded by
application specific resource classes.

I will write some tests and docs if this approach is acceptable to the
maintainers.

### DIFF
--- a/lib/jsonapi/request_parser.rb
+++ b/lib/jsonapi/request_parser.rb
@@ -426,9 +426,8 @@ module JSONAPI
 
     def check_sort_criteria(resource_klass, sort_criteria)
       sort_field = sort_criteria[:field]
-      sortable_fields = resource_klass.sortable_fields(context)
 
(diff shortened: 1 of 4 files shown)


rfb:
related PR for the guides: https://github.com/cerebris/jsonapi-resources-site/pull/6/files

lgebhardt:
@rfb Thanks!

PR #1003, opened 2017-03-15
Bug fix: Allow multiple relationships to be defined in one declaration.

dougo:
I ran into some weird errors when I used `has_many :foo, :bar` to define two relationships in one declaration. I tracked it down to the fact that it was mutating the options hash in order to add `class_name: Foo` for the first relationship, which polluted the options for the second relationship.


lgebhardt:
@dougo Thanks!

PR #1014, opened 2017-03-27
require bundler in bug_report_templates_test

### LINKED ISSUE #1013: BugReportTemplatesTest failure
The only test that fails on the master branch is the test from BugReportTemplatesTest.

It fails because Bundler is not loaded in. If I require bundler, it works but then I get this long ouput:

screen shot 2017-03-27 at 12 01 43 am

Was this test intended to be run? Or can I add a defined?(Bundler) check on it so it can be run optionally?

It sucks to constantly run specs and have it fail on this one test.

senid231:
fix #1013
fails when `rake` runs without `bundle exec`


Aryk:
Does this prevent all the output that happens when you require 'bundler'? When I put "require 'bundler'" i got all this output, but could just be me:


screen shot 2017-03-27 at 12 01 43 am

senid231:
add optional log silencer

senid231:
@lgebhardt take a look, please

lgebhardt:
@senid231 This looks fine, though I'd be just as happy excluding the test from being run. That could be done by removing ` assert_bug_report 'rails_5_master.rb'` from `test/bug_report_templates_test.rb`.

Do you see any reason to run the bug_report_templates as part of standard testing?

senid231:
I think we need to be sure that bug report template works for master. But adding it to test suite looks little ugly. It will be better to create separate rake task for it and run after test suite.
@lgebhardt  do you agree with this?

lgebhardt:
@senid231 That sounds right.

senid231:
reworked and rebased
rake task for bug report template added and used on default rake

@lgebhardt, please take a look

lgebhardt:
@senid231 Thanks for cleaning that up. I've merged it as is, but I do wonder if we should exclude the new task from the default rake. But we can do that later if it causes problems.

PR #1020, opened 2017-04-03
Handle empty include= parameter.

dougo:
Out of curiosity, I tried sending my JR server a request with `?include=` with nothing after the `=`, and got a 500 error. It turns out that, annoyingly, `CSV.parse_line('')` returns `nil` rather than an empty array.

I thought about also changing the `nil?` check on line 353 to be `empty?`, because I don't see how `included_resources` can ever be `nil` at that point, but I figured it might be important to return an empty `IncludeDirectives` object in this case. (In which case that line should probably be deleted.)

### DIFF
--- a/lib/jsonapi/request_parser.rb
+++ b/lib/jsonapi/request_parser.rb
@@ -345,7 +345,7 @@ module JSONAPI
 
       included_resources = []
       begin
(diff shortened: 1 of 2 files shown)


lgebhardt:
@dougo Thanks!

PR #1021, opened 2017-04-04
Improve polymorphic links

Exelord:
This PR will improve loading of polymorphic links.
Now it will point directly to model relation without hitting resource which will improve request time in case of custom logic (in downloading relation) [records_for]

In our case, this improved request time by 300%

### DIFF
--- a/lib/jsonapi/relationship.rb
+++ b/lib/jsonapi/relationship.rb
@@ -49,8 +49,8 @@ module JSONAPI
 
     def type_for_source(source)
       if polymorphic?
-        resource = source.public_send(name)
-        resource.class._type if resource
+        model = source._model.public_send(@relation_name)
+        source.class.resource_klass_for_model(model)._type if model
       else
         type
       end
--- a/lib/jsonapi/resource_serializer.rb
+++ b/lib/jsonapi/resource_serializer.rb
@@ -454,7 +454,7 @@ module JSONAPI
     def foreign_key_value(source, relationship)
       related_resource_id = if source.preloaded_fragments.has_key?(format_key(relationship.name))
         source.preloaded_fragments[format_key(relationship.name)].values.first.try(:id)
-      elsif !relationship.redefined_pkey? && !relationship.polymorphic? && source.respond_to?(relationship.foreign_key)
+      elsif !relationship.redefined_pkey? && source.respond_to?(relationship.foreign_key)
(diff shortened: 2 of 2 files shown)


Exelord:
@lgebhardt What do you think about it?

lgebhardt:
@Exelord It's bypassing the RecordAccessor in the default case and any overrides that are on the resource. This will bypass the caching in the default ActiveRecordAccessor and any custom logic built into custom RecordAccessors and Resources.

300% sounds like a major improvement, and I agree we should look for a solution. Any idea what part is taking most of the time?

Exelord:
In our case:
-> `source.public_send(name)` will trigger
-> `records_for(relation_name)` for every of X resources in the `index` operation
-> the `records_for` will run every time authorization method which is time-consuming
-> then the model is serialized to a resource
-> and then we just ask about resource type -.-

I understand that this is important for resources caching but I think we are not using it when we are asking for `type_for_source` during `always_include_linkage_data` process.

I don't see any need to check authorization (using records_for) to record while we are getting just a reference.
In our case when the authorization is a little bit complicated (graphs, access inheritance, roles etc.) and we are returning like 50 records per page, this operation is a killer for our endpoint. Instead of 600ms, we got even 6-8s. So it's even more than 300%.
But on regular endpoints with polymorphic relations, we saved average the mentioned 300%. (Of course, it depends on how complicated your `records_for` looks like)

This method is only used for getting the type of references model, so maybe we should consider this over caching a record that we do not use it later... in the case when the reference will be needed again, the ActiveRecord will cache the relation as expected.

lgebhardt:
N (50 in this case) queries for a request is obviously undesirable. We definitely need a fix. Won't this PR still result in the N queries being run?

Exelord:
Yes, the n-Query on `records_for` is the main problem here.
When we are using this fix, we bypass it, and we are not getting any N-Query thanks to includes.

Example:
I return a list of 20 Members with a polymorphic belongs to relation `resource` which has a flag `always_include_linkage_relation` to true.

Normally I will get an n-query requesting a type of a `resource` for each `member`, but then we have another quick fix. We just use `includes` in `records` method, like:
```ruby
  super(options).includes(:resource)
end
```

and now the endpoint is much faster without any n-query

Exelord:
Hi @lgebhardt, what's the status here? :)

Exelord:
@lgebhardt ping?

PR #1023, opened 2017-04-10
Fixed cache pollution bug

DavidMikeSimon:
Also refactored ActiveRecordAccessor a little.

### DIFF
--- a/lib/jsonapi/active_record_accessor.rb
+++ b/lib/jsonapi/active_record_accessor.rb
@@ -477,7 +477,10 @@ module JSONAPI
             rel_id = row[index+1]
             assoc_rels = res.preloaded_fragments[rel_name]
             if index == path.length - 1
-              assoc_rels[rel_id] = target_resources[klass.name].fetch(rel_id)
+              fragment = target_resources[klass.name].fetch(rel_id)
+              if fragment
+                assoc_rels[rel_id] = fragment
+              end
             else
               res = assoc_rels[rel_id]
             end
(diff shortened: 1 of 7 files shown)


DavidMikeSimon:
@lgebhardt This look alright to you?

lgebhardt:
@DavidMikeSimon I've quickly looked it over and I don't see anything controversial, but I haven't tried to fully understand the changes. Can you provide a link to the issue it's fixing, or a brief write up of the issue?

DavidMikeSimon:
The issue is summed up best by the regression test:

```ruby
 class Api::V2::AuthorsControllerTest < ActionController::TestCase
￼   def test_cache_pollution_for_non_admin_indirect_access_to_banned_books
￼     cache = ActiveSupport::Cache::MemoryStore.new
￼     with_resource_caching(cache) do
￼       $test_user = Person.find(5)
￼       get :show, params: {id: '2', include: 'books'}
￼       assert_response :success
￼       assert_equal 2, json_response['included'].length
￼ 
￼       $test_user = Person.find(1)
￼       get :show, params: {id: '2', include: 'books'}
￼       assert_response :success
￼       assert_equal 1, json_response['included'].length
￼     end
￼   end
￼ end
```

The issue was that during fragment preloading, custom `records` methods in Resource classes were only being called on cache misses. As a result, if one user retrieved some included records through an association, and JR cached the results, then other users might end up seeing included records that they were not meant to see.

Also, in the process I cleaned up `preload_included_fragments` a little, my original implementation had way too many pieces of state being tracked at the same time. It's still pretty hairy, but a little less so now.

I'm going to split out the test into its own commit (as I should have done in the first place) so that the problem and solution can be independently demonstrated.

lgebhardt:
@DavidMikeSimon Thanks!

DavidMikeSimon:
@lgebhardt No problem. :-)

I also have a semi-related docs update PR, cerebris/jsonapi-resources-site#8

PR #1026, opened 2017-04-13
Handle exceptions from process_operations callbacks, remove :rollback callback

DavidMikeSimon:
This PR allows exceptions thrown from callbacks to be handled correctly, in the same way as exceptions thrown during operation processing.

This meant two changes to `process_request` :

* Moves the call to `handle_exceptions` outside of the blocks passed to `run_callbacks` and `run_in_transaction`.
* Not manually raising an `ActiveRecord::Rollback` if an exception was caught at this point (done previously with the `force_rollback` flag), because it was eating the exception. Anyways, AR transaction blocks already correctly do a rollback when an exception raises through the transaction boundary, so there's no need  for this logic to be here.

The logic for when we *do* want to directly cause a rollback (e.g. when there are errors in the response document, even if no exceptions were raised) has also been simplified. Instead of detecting if we are in a transaction and only throwing `ActiveRecord::Rollback` if so, we always throw it. Our "fake" transaction block then just catches and silently discards that to mimic the behavior of real transaction blocks.

Additionally, this PR removes the :rollback callback from ActsAsResourceController. The `around_rollback` and `after_rollback` would not have worked as expected because causing a rollback means throwing an exception, interrupting the `run_callbacks` call. Making this work as expected would probably have to mean re-implementing `ActiveRecord::transaction` so that the callback happens around the actual database `ROLLBACK` statement. I'm not sure about the intention behind this feature, so @lgebhardt if you'd like I can find another way to get it working; for now I'm operating on the assumption that it's a vestige because it's easier this way. :-)

### DIFF
--- a/lib/jsonapi/acts_as_resource_controller.rb
+++ b/lib/jsonapi/acts_as_resource_controller.rb
@@ -10,8 +10,7 @@ module JSONAPI
       base.include Callbacks
       base.cattr_reader :server_error_callbacks
       base.define_jsonapi_resources_callbacks :process_operations,
-                                              :transaction,
-                                              :rollback
+                                              :transaction
     end
 
     attr_reader :response_document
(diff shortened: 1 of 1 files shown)


lgebhardt:
@DavidMikeSimon Thanks!

PR #1027, opened 2017-04-14
Set version constant to 0.10.0.pre

DavidMikeSimon:
This is to make it easier to test gems against the master branch of JR when their dep specs say ">= 0.9.0" or similar.


lgebhardt:
@DavidMikeSimon Thanks!

Issue #927, opened 2016-12-13
Cross namespace relationships


lgebhardt:
#### The problem:

Right now namespaces are assumed to be used for versioning, though they could also be used for separating entire sets of resources in isolated namespaces. JR does not currently support resource relationships which cross namespaces. #893 attempts to solve this, but as I brought up in the comments there is the potential for the system to encounter a situation where the correct behavior is ambiguous or will be inconsistent depending on the order of serialization. 

#### Proposal:

A simple set of resource resolution rules could be used to solve this issue without the need to specifically mark up relationships definitions on a resource.

Rules:
* All resource resolution will take place relative to the namespace of the request ("Request Namespace").
* Related resources will be assumed to be in the Request Namespace.
* Resource definitions will support a new method called `global_resource` which will register a resource type in a "Global Resource Map" to this resource class.
* Declaring two resources classes with the same type as `global_resource` will raise an exception or emit a warning, which should catch conflicts during development.
* If a resource class does not exist in the Request Namespace the Global Resource Map will be consulted for the resource class.
* Resources created from the Global Resource Map will still follow same resolution rules starting with the Request Namespace. In other words relationships in a global resource will first resolve based on the Request Namespace.

I believe these rules should allow unambiguous resolution of resources and prevent runtime issues where a resource is serialized with differing sets of rules (resource definitions), and meet the intent of #893. 

Let's discuss this and other proposals here. I haven't tried to code this, but I think it's possible to modify the code to support these rules without a lot of refactoring. If you see potential road blocks in either this project or your projects please raise them here.

senid231:
nice solution
it will solve problem described in #893, so i will close it

dwillett:
My current project could really use this flexibility. We have an additional `admin` namespace in our resource definitions that does not match the module prefix for the associated models, and I'm struggling to fully workaround the current strategies of resolving resource classes. Happy to offer more details as there may be a way I'm not seeing, but I think @lgebhardt's solution would work for us

toobulkeh:
We could really use namespacing to sort hundreds of models in a single complex application.

There's only so many times we can name something "Environment" 😄

gabrielwillemann:
Which version this feature will be release?

rafaelpivato:
> Resources created from the Global Resource Map will still follow same resolution rules starting with the Request Namespace. In other words relationships in a global resource will first resolve based on the Request Namespace.

[...]

PR #1028, opened 2017-04-14
Fix nested namespaces parsing and resource resolving

elvanja:
With this setup: 
```ruby
class Managing::Profile < ActiveRecord::Base
  belongs_to :company
end

class Company < ActiveRecord::Base
  belongs_to :default_managing_profile, class_name: '::Managing::Profile'
end

class Api::V1::BaseResource < JSONAPI::Resource
  abstract

  attributes :created_at, :updated_at

  def self.updatable_fields(context)
    super - [:created_at, :updated_at]
  end

  def self.creatable_fields(context)
    super - [:created_at, :updated_at]
  end
end

class Api::V1::Managing::ProfileResource < Api::V1::BaseResource
  model_name 'Managing::Profile'
  attributes :name
end

class Api::V1::CompanyResource < Api::V1::BaseResource
  model_name 'Company'
  attributes :name
  has_one :default_managing_profile, class_name: 'Managing::Profile'
end
```
, there were several things that didn't work:
- on startup, without `class_name: 'Managing::Profile'` in `CompanyResource`, it would report that it could not find related model
- requesting included profile via `http://localhost:3000/api/v1/companies/3?include=default_managing_profile`, it would also break, unable to find the related resource
- on posting a new company, with relation to existing managing profile, it would again break, for the same reason

example of post payload: 
```json
{
  "data": {
    "attributes": {
      "name": "Dummy"
    },
    "relationships": {
      "default_managing_profile": {
        "data": {
          "type": "profiles",
          "id": "3"
        }
      }
    },
    "type": "companies"
  }
}
```
I also tried setting the data type for `default_managing_profile` to one of:
- `managing/profiles`
- `/managing/profiles`
- `Managing::Profile`
- `managing_profiles`

and probably some other too, but without luck.

The fix in `resource.rb` solved first two issues. The fix in `request_parser.rb` fixed the last one. I tried adding related tests, but could not find a simple way to test the last issue.

Hopefully this makes sense 😄 

P.S. I've noticed that `RequestParser#parse_to_many_relationship` has a bit different logic than `parse_to_one_relationship` so maybe I'm missing something here. Maybe there is a way to accomplish the desired outcome without this fix?

P.P.S. Possibly related to #893 and #927.

### DIFF
--- a/lib/jsonapi/request_parser.rb
+++ b/lib/jsonapi/request_parser.rb
@@ -529,7 +529,7 @@ module JSONAPI
 
       unless links_object[:id].nil?
         resource = resource_klass || Resource
-        relationship_resource = resource.resource_klass_for(unformat_key(links_object[:type]).to_s)
+        relationship_resource = resource.resource_klass_for(unformat_key(relationship.options[:class_name] || links_object[:type]).to_s)
         relationship_id = relationship_resource.verify_key(links_object[:id], @context)
         if relationship.polymorphic?
           { id: relationship_id, type: unformat_key(links_object[:type].to_s) }
--- a/lib/jsonapi/resource.rb
+++ b/lib/jsonapi/resource.rb
@@ -446,7 +446,7 @@ module JSONAPI
 
       def resource_klass_for(type)
         type = type.underscore
-        type_with_module = type.include?('/') ? type : module_path + type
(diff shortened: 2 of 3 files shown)


lgebhardt:
@elvanja Thanks!

kented:
This issue was causing me some pain.  Thanks for the fix, @elvanja

PR #1041, opened 2017-04-21
Fix apply_sort

Exelord:
Hi @lgebhardt,
We encountered an issue associated with `apply_sort` method.

Doc is saying (http://jsonapi-resources.com/v0.9/guide/resources.html#Sortable-Attributes) that we can override apply_sort like:
```ruby
def self.apply_sort(records, order_options, context = {})
  if order_options.has?(:trending)
    records = records.order_by_trending_scope
    order_options - [:trending]
  end

  super(records, order_options, context)
end
```

But indeed, we can not use `super(records, order_options, context)` because the super method does not exist. And because of that, any other sorts have not been applied.

I found a problem and modified the `apply_sort` method and now it's performing default sorting -> passing the results to custom apply_sort (if exist) -> returning custom apply_sort results or if they are `nil` resulting original results.

So now you can just:
```ruby
def self.apply_sort(records, order_options, context = {})
  if order_options.has?(:trending)
    order_options - [:trending]
    records.order_by_trending_scope
  end
end
```

And in fact, this will work as a `super` but also without a need to call it.

### DIFF
--- a/lib/jsonapi/active_record_accessor.rb
+++ b/lib/jsonapi/active_record_accessor.rb
@@ -210,6 +210,11 @@ module JSONAPI
     end
 
     def apply_sort(records, order_options, context = {})
+      if defined?(_resource_klass.apply_sort)
+        custom_sort = _resource_klass.apply_sort(records, order_options, context)
+        records = custom_sort unless custom_sort.nil?
+      end
+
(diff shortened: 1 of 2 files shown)


lgebhardt:
@Exelord Thanks!

Issue #917, opened 2016-12-01
Caching not working with RedisStore Cache


x4d3:
Using Rails.cache as a `resource_cache` appears to make the caching fail when config.cache_store is redis_store.

I think the reason is due to the way Redis Cache serialise key as string when they are arrays.

When the key(an array) is serialised in the `self.write` method of `CachedResourceFragment`
https://github.com/cerebris/jsonapi-resources/blob/master/lib/jsonapi/cached_resource_fragment.rb#L122 it is serialised as a string.
When the value is retrieved in the `lookup` method the hash `hits` returned by the `read_multi` methods has keys that are not array, but strings. The method `has_key?` is then failing returning false instead of true.
https://github.com/cerebris/jsonapi-resources/blob/master/lib/jsonapi/cached_resource_fragment.rb#L81

Simple example, with  `config.cache_store = :redis_store`
```
Rails.cache.write([1,2], 'value')
hash = Rails.cache.read_multi([1,2])
# returns false
hash.has_key?([1,2])
```
I've tried to serialise manually the keys (https://github.com/cerebris/jsonapi-resources/compare/master...x4d3:master) without success, the test are failing.

Thank you,

DavidMikeSimon:
Thank you for the bug report, I'll look into this

DavidMikeSimon:
@x4d3 I'm not able to reproduce your example results using RedisStore.

```
irb(main):001:0> require 'redis-rails'
=> true

irb(main):003:0> c = ActiveSupport::Cache::RedisStore.new
=> #<ActiveSupport::Cache::RedisStore:0x00000001288010 @options={}, @data=#<Redis client v3.3.2 for redis://127.0.0.1:6379/0>>

irb(main):005:0> c.write([1,2], 'value')
=> "OK"

irb(main):008:0> h = c.read_multi([1,2])
=> {[1, 2]=>"value"}

irb(main):009:0> h.has_key?([1,2])
=> true
```

Could you try again with the current version of redis-rails? Also, in production on my own app, I'm getting good results with the `readthis` gem.

DavidMikeSimon:
@lgebhardt I'd recommend closing this issue, I'm unable to reproduce this bug.

PR #1043, opened 2017-04-21
Fix an issue with redis keys not hitting

joegaudet:
When writing to a redis store, the keys used by JR are converted to
slash delimited strings. So:

key = [record_type, id, updated at]

Gets converted to record_type/id/updated_at

This causes the lookup function to never get hits from the redis store.

This commit adds a configurable function to allow formating of the
lookup key.

### DIFF
--- a/lib/jsonapi/cached_resource_fragment.rb
+++ b/lib/jsonapi/cached_resource_fragment.rb
@@ -81,6 +81,11 @@ module JSONAPI
       hits = JSONAPI.configuration.resource_cache.read_multi(*keys).reject{|_,v| v.nil? }
       return keys.each_with_object({}) do |key, hash|
         (_, id, _, _) = key
+
+        if JSONAPI.configuration.resource_cache_key_format_function
+          key = JSONAPI.configuration.resource_cache_key_format_function(key)
+        end
+
         if hits.has_key?(key)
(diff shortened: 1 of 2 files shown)


DavidMikeSimon:
@joegaudet This particular issue with redis has been reported before in #917, and I wasn't able to reproduce it. Can you tell me more about your environment, e.g. Ruby version, name and version of redis gem used?

joegaudet:
@DavidMikeSimon Gems:

gem 'redis-rails', '~> 4.0.0' # Rails Redis adapters, caching, sessions, etc.
gem 'redis-namespace', '~> 1.5.0' # namespacing for Redis
gem 'rails', '~> 4.2.0'

ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-darwin15]

So the issue seems to be in the way things get written to redis, our keys looks like this:

orders/321/2017-04-26 23:52:58 UTC/SRLZ-sZHh1BNSdDAaht97RdKB23JJ8+D5bQpy28pQdI5xFhg=/ATTR-CTX-k_+raGkUaGivBBbrqmIyPlG3wldRXXdUeNbg1_d37K0=

record-type/record-id/updated-at/hash

Where the JR system appears to use an array as the key.

PR #1048, opened 2017-05-04
WIP: Resource level pagination options

samf89:
I've recently had the need to implement resource level pagination options. I've put in a `WIP` MR for this - not fully tested, and I suspect that the changes to `request_parser` will break *something* somewhere. 

```
class MyResource < JSONAPI::Resource
  paginator :paged, default_page_size: 50, default_page_number: 3
end
```

Is this an idea that is worth pursuing?


blindMoe:
Just ran into this same situation.  Is this something we can move forward with?  If so, let me know if I can help.

PR #1049, opened 2017-05-04
Tightens the error handling on request parsing.

lgebhardt:
Now fails on the first error encountered.


dgeb:
Looks simpler and safer 👍

PR #1061, opened 2017-05-10
Allow sorting on related attributes of has_one relationships.

Ross-Hunter:
Current master has a bug when sorting on related attributes, if the relationship is a `has_one`/`foreign_key_on: :related`. The following code will produce this error

```
Mysql2::Error: Unknown column 'containers.container_id' in 'on clause': SELECT  `containers`.* FROM `containers` LEFT JOIN sensors AS sensor_sorting ON sensor_sorting.id = containers.container_id
```

``` ruby
class Container < ApplicationRecord
  has_one :inventory_sensor, required: false
end

class Sensor < ApplicationRecord
  belongs_to :container, required: false
end

class ContainerResource < BaseResource
  has_one :sensor, foreign_key_on: :related

  def self.sortable_fields(context)
    super(context) + [:"sensor.reading"]
  end
end
```
`_build_joins()` was looking at the wrong table for the related foreign_key. The join needs to be built differently depending on where the foreign key will live. The current tests don't catch this because they seem to all use a self-referential `has_one` relationship, so the foreign_key is there.

So I just added a conditional that alters the query based on whether or not it is a `has_one` relationship, and updated the test. I have integration testing around this in my project, but I was having difficulty dealing with the `jsonapi-resources` test suite, so I just fixed the one test I broke, but it would be great to have some integration testing around this.

Feel free to make any necessary adjustments or provide me with any feedback.

Thanks for the awesome gem.

### DIFF
--- a/jsonapi-resources.gemspec
+++ b/jsonapi-resources.gemspec
@@ -21,7 +21,7 @@ Gem::Specification.new do |spec|
 
   spec.add_development_dependency 'bundler', '~> 1.5'
   spec.add_development_dependency 'rake'
-  spec.add_development_dependency 'minitest'
+  spec.add_development_dependency 'minitest', '~> 5.10', '!= 5.10.2'
   spec.add_development_dependency 'minitest-spec-rails'
   spec.add_development_dependency 'simplecov'
   spec.add_development_dependency 'pry'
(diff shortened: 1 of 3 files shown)


lgebhardt:
@Ross-Hunter Thanks! Merged via #1106

PR #1062, opened 2017-05-11
Add custom sorts feature

Exelord:
Hi,
I've just added a feature which adds a possibility to define custom sorts as filters do.
```rb
  sort :my_sort, apply: (lambda do |records, order_direction, context|
    records.my_custom_order_by(order_direction)
  end)
```

Thanks to that users can just forgot about overriding `apply_sort` method to just add a custom sort property like:
```rb
    def apply_sort(records, order_options, _context = {})
      if order_options.key?('my_sort')
        records = records.order_by_my_sort
      end
    end

    def sortable_fields(context)
      super(context) + [:my_sort]
    end
```

So now, this is unneeded.


Exelord:
@lgebhardt what do you think about it?

PR #1066, opened 2017-05-19
Included the ability to define a custom type to a resource

bsofiato:
This enables the developer to define a customize the resource type. In order to do so, the developer should invoke the method `type` passing the required type as a parameter. See the example below:

```
class PersonWithCustomResourceTypeResource < PersonResource
  type 'this-is-a-custom-type'
end
```

### DIFF
--- a/lib/jsonapi/resource.rb
+++ b/lib/jsonapi/resource.rb
@@ -834,6 +834,10 @@ module JSONAPI
         @abstract = val
       end
 
+      def type(type)
+        self._type = type
+      end
+
       def _abstract
         @abstract
       end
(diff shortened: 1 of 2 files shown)


lgebhardt:
While this will work to allow a resource to serialize with the alternate type, it doesn't handle the case where we need to look up the resource class by type using the `Resource.resource_klass_for` method. Why not just create `ThisIsACustomResource` so it's serialized/deserialized correctly?

bsofiato:
@lgebhardt The rationale behind this PR is allow the developer to explicitly define the type name of a resource. It should bey used when the resource naming scheme does not suffice -- once, I had to name a resource something like `folder/resource` to prevent conflicts. Other users have needed this ability as well (see Issue #1070).

I can develop the PR further and cover the case of using the `Resource.resource_klass_for` method to find a resource by its name. If you guys feel that this feature is worth having on the library's code base.

PR #1069, opened 2017-05-25
Use original key format when displaying error for filter not allowed

kevintraver:
`/clients?filter[first-name]='bob'` 

which results in:

```ruby
{
   "errors":[
      {
         "code":"102",
         "detail":"first_name is not allowed.",
         "status":"400",
         "title":"Filter not allowed"
      }
   ]
}
```

Which could be potentially confusing.


lgebhardt:
@kevintraver Thanks!

PR #1080, opened 2017-07-01
Fix caching for namespaced polymorphic associations

alachaum:
Hi guys,

I noticed there was an issue with caching when looking up namespaced resource classes for polymorphic associations. This issue prevents include queries on resources having a polymorphic association.

Something like this:
```ruby
class Api::V1::PictureResource < JSONAPI::Resource
  caching
  has_one :imageable, polymorphic: true
end

class Api::V1::ProductResource < JSONAPI::Resource
  has_one :picture
end
```

..will fail when doing:
```
# Fails because JR attempts to lookup PictureResource when compiling cached fragments
GET /api/v1/pictures?include=imageable
```

The test attached to this PR does not create a 500 if the patch is not applied (because both PictureResource and Api::V1::PictureResource are defined) but clearly show that the links are incorrect when caching is enabled.

### DIFF
--- a/lib/jsonapi/cached_resource_fragment.rb
+++ b/lib/jsonapi/cached_resource_fragment.rb
@@ -65,7 +65,8 @@ module JSONAPI
     end
 
     def to_real_resource
-      rs = Resource.resource_klass_for(self.type).find_by_keys([self.id], {context: self.context})
+      rs_klass = @resource_klass || Resource.resource_klass_for(self.type)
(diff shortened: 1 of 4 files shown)


lgebhardt:
@alachaum I think this has been fixed with the rework that has happened in #1045. Please check it out and reopen this if it's still a problem. We'll need to fix it another way since the `to_real_resource` method no longer exists.

PR #1084, opened 2017-07-07
Enable custom related links on resource

travisofthenorth:
I ran into a use case where a related resource lives on another server, and by providing a custom `related` link for the relationship my Ember code can seamlessly fetch the relationship.

### DIFF
--- a/lib/jsonapi/resource.rb
+++ b/lib/jsonapi/resource.rb
@@ -167,6 +167,11 @@ module JSONAPI
       {}
     end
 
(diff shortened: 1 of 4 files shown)


travisofthenorth:
@lgebhardt any idea when the tests in `test/unit/serializer/serializer_test.rb` will get updated?

chrism:
This would be really useful!

fengshuo:
this would be really helpful

PR #1085, opened 2017-07-07
Break into an individual function in case users want to override & operate on the content

krainboltgreene:
In my case I wanted to cache the entire outgoing document based on a very specific cache path:

``` ruby
private def response_document_contents
  Rails.cache.fetch(cache_key) { super }
end
```


lgebhardt:
@krainboltgreene Could you please rebase this against master and make sure the tests all pass.

krainboltgreene:
Rebased. @lgebhardt You should turn on autocanceling for PRs. It's in the settings on travis.

krainboltgreene:
I have to admit, I have no idea how these errors are happening or what they mean.

lib/jsonapi/acts_as_resource_controller.rb
```diff
@@ -215,19 +215,17 @@ def base_response_links
...
       render_options = {}
       if response_document.has_errors?
-        render_options[:json] = content
+        render_options[:json] = response_document_contents
       else
         # Bypassing ActiveSupport allows us to use CompiledJson objects for cached response fragments
-        render_options[:body] = JSON.generate(content)
+        render_options[:body] = JSON.generate(response_document_contents)
+  
+        return unless response_document.status == 201
```
lgebhardt, line 225: This is the cause of the errors. You can't return early here.

lgebhardt:
Looking briefly at the code I'm not sure it's going to do what you want it to. By the time the code has made it to `render_response_document` the work has been done in processing the request and generating the response. At this point I don't see an advantage to caching the whole document. The hooks seem like they would be better in `process_request` so you could avoid the work, if desired.

The errors are coming from where you return early in `render_response_document`.

krainboltgreene:
Alright.

PR #1096, opened 2017-08-08
Allow more control over allowing includes

lgebhardt:
Deprecates and replaces the `allow_include` config option

Replaces `allow_includes` with `default_allow_include_to_one` and `default_allow_include_to_many` configuration options.
Adds support for `allow_include` option on Relationships to override the config settings.

Purpose:
- Includes without pagination allows potentially resource intensive requests.
- Pagination of includes is a non-trivial problem to solve efficiently with SQL databases
- By allowing includes defaults to be set for to_one and to_many relationship types separately includes for the riskier to_many relationships can be disabled by default, while still allowing them on specific relationships if desired.

### DIFF
--- a/lib/jsonapi/relationship.rb
+++ b/lib/jsonapi/relationship.rb
@@ -5,6 +5,8 @@ module JSONAPI
                 :parent_resource, :eager_load_on_include, :custom_methods,
                 :inverse_relationship, :allow_include
 
+    attr_writer :allow_include
+
     def initialize(name, options = {})
       @name = name.to_s
       @options = options
@@ -134,7 +136,7 @@ module JSONAPI
 
       def allow_include?(context = nil)
         strategy = if @allow_include.nil?
-                     JSONAPI.configuration.default_allow_include_to_one
+                     JSONAPI.configuration.default_allow_include_to_many
                    else
                      @allow_include
                    end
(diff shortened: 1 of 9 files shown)


lib/jsonapi/relationship.rb
```diff
@@ -100,6 +101,14 @@ def belongs_to?
       def polymorphic_type
         "#{name}_type" if polymorphic?
       end
+
+      def allow_include?
```
joegaudet: This is awesome, would be cool to have the context passed in, similarly to fetchable_fields useful to remove some depending on the current user.

lgebhardt: @joegaudet Good idea

lgebhardt:
Updated for @joegaudet so you can now provide a boolean, a callable name, or a lambda for `allow_include`. Context is passed to the callable and lambda.

lib/jsonapi/relationship.rb
```diff
@@ -114,6 +131,23 @@ def initialize(name, options = {})
           @inverse_relationship = options.fetch(:inverse_relationship, parent_resource._type.to_s.singularize.to_sym)
         end
       end
+
+      def allow_include?(context = nil)
+        strategy = if @allow_include.nil?
+                     JSONAPI.configuration.default_allow_include_to_one
```
dgeb: should be `JSONAPI.configuration.default_allow_include_to_many`

dgeb (changes requested):
Looks pretty good. Needs one simple fix. I'd also feel more comfortable if the recursion logic in `check_include` were tested.

dgeb (approved):

PR #1099, opened 2017-08-18
fixes #909 add possibility to set generic validation error

### LINKED ISSUE #909: ValidationErrors treats :base as an attribute and doesn't use details
Hi, 

 I would like to use the `error.detail` field to display customised errors messages for the front end. After some study, I'd like to explore some options of enhancement on ValidationErrors: 

**Attribute `:base` should be treated as the primary resource entity instead of an attribute**

Currently, `:base` is being treated as an attribute in source.point, e.g. `
"source": {
        "pointer": "/data/attributes/base"
      }
`. However, according to rails's doc at http://api.rubyonrails.org/classes/ActiveModel/Errors.html, "attribute should be set to :base if the error is not directly associated with a single attribute."  The `pointer` can be improved by pointing to `'/data'` as JSONAPI Spec's 'MAY' spec at http://jsonapi.org/format/#errors. And it will provide consistency with Rails as well.

**Will be be preferable if ValidationErrors use messages as detail with prefixing**

My reasoning is on JSONAPI spec's "MAY" recommendations on `title` and `detail` say: 

> - title: a short, human-readable summary of the problem that SHOULD NOT change from occurrence to occurrence of the problem, except for purposes of localization.
> - detail: a human-readable explanation specific to this occurrence of the problem. Like title, this field’s value can be localized.

It seems that rails's error messages behave more closely to JSONAPI Spec's `detail` as they allow dynamically use `%{value}`, `%{attribute}`, and `%{model}` to provide specifics to this occurrence. So JSONAPI spec's `detail` may just be rails's error message.  

However, in `ValidationErrors#json_api_error`, it prefixes the message with the attribute and a hyphen. This may be not always desirable as some, as in my case, would prefer to use their customized validation messages  as is for client facing UI. 

`detail: "#{format_key(attr_key)} - #{message}"`

Of course, we can just use the `title` as it is the error message as is at the moment. But it will slightly deviate from the `title`'s MAY recomendation. 

This can be even  taken  a step further to allow users to provide custmised ValidationErrors in `resource.rb` to define their user of the `title`, `detail`, `code` and other fields in JSONAPI Spec's error object. But that may take some futher elaboration. 

It will be nice to take them into consideration. 

Many thanks, 

Mytram

senid231:
Fixes #909 
very simple example usage:
```ruby
class Person < ActiveRecord::Base
  validate do
    errors.add(:base, 'invalid attributes') if name.blank?
  end
end

class PersonResource < JSONAPI::Resource
  model 'Person'
  attributes :name

  def model_error_messages
    errors = super
    errors[nil] = errors[:base] if errors.has_key?(:base)
    errors
  end
end
```
request
```ruby
post '/people', {data: {type: 'people', attributes: {name: ''}}}, json_api_headers
```
will be responded with
```http
HTTP/1.1 422 Unprocessable Entity

{
  "errors": [
    {
      "title": "invalid attributes",
      "detail": "invalid attributes",
      "code": "100",
      "source": {
        "pointer": "/data"
      },
      "status": "422"
    }
  ]
}
```

I didn't add tests for it because I don't know where I should put it.
I decide to use `nil` key - not `:base` because `base` is a valid attribute name for a `jsonapi`, and to not break existing functionality.

### DIFF
--- a/lib/jsonapi/exceptions.rb
+++ b/lib/jsonapi/exceptions.rb
@@ -465,11 +465,12 @@ module JSONAPI
     end
 
     class ValidationErrors < Error
-      attr_reader :error_messages, :error_metadata, :resource_relationships
+      attr_reader :error_messages, :error_metadata, :resource_relationships, :resource_class
 
       def initialize(resource, error_object_overrides = {})
         @error_messages = resource.model_error_messages
         @error_metadata = resource.validation_error_metadata
+        @resource_class = resource.class
         @resource_relationships = resource.class._relationships.keys
         @key_formatter = JSONAPI.configuration.key_formatter
         super(error_object_overrides)
@@ -491,7 +492,7 @@ module JSONAPI
         create_error_object(code: JSONAPI::VALIDATION_ERROR,
                             status: :unprocessable_entity,
                             title: message,
-                            detail: "#{format_key(attr_key)} - #{message}",
+                            detail: detail(attr_key, message),
                             source: { pointer: pointer(attr_key) },
                             meta: metadata_for(attr_key, message))
       end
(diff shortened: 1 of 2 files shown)


senid231:
@dgeb @lgebhardt any thoughts about this?

lgebhardt:
I'd like to see a solution to handle the errors on `:base` without requiring the override of `model_error_messages`.  

In `ValidationErrors`  we do know the Resource so we could check for an attribute named `base`. If it exists we treat  `:base` as an attribute, and if it doesn't we treat it as a general validation error.

Does that sound reasonable?

senid231:
@lgebhardt I agree with you

PR changed and rebased

lgebhardt:
@senid231 Thanks! I added a few tests and merged this in #1104.

PR #1107, opened 2017-09-07
Fix a bug that occurs when fetching relationships

clemoberti:
I think there was a little mistake in here, your probably meant
`context = options[:context]` instead of `context = context`
This solve an issue I had while fetching relationships which was returning a nil context.
Thanks a lot for all your efforts!


lgebhardt:
@clemthenem Thanks!

PR #1111, opened 2017-09-09
Do not verify_filters in the RequestParser

### LINKED ISSUE #1110: Filter values for related resources are wrapped in nested arrays?
I'm a little confused as to what's going on, but upgrading from 0.8.3 to 0.9 broke some filters I have in place. e.g.:

```ruby
VERIFY_RADIUS_FILTER = ->(value, context) {
  return NoLimit.new unless context[:current_location].present?

  value
}

FILTER_RADIUS_FROM_LOCATION = ->(records, value, options) {
  location = options.dig(:context, :current_location)
  limit = Array(value).first

  return records unless location.present? && limit.present?

  records.near(location, Float(limit))
}

filter :radius, default: 100.0,
                verify: VERIFY_RADIUS_FILTER,
                apply: FILTER_RADIUS_FROM_LOCATION
```

I think this commit my be the culprit? https://github.com/cerebris/jsonapi-resources/commit/3a691b29adb3e7aff0743ef59e44556bedb260c9

From what I can tell, this results in `Resource#verify_resource` being called twice. For a custom filter, at least (which is what I've got), that means the `raw` value gets doubly-nested in an array: https://github.com/cerebris/jsonapi-resources/blob/19f4d7b59f92b05fa28c4f925391cf5998123308/lib/jsonapi/resource.rb#L753

Instead I wonder if that line should be something like:

```ruby
filter_values += raw.is_a?(String) ? CSV.parse_line(raw) : Array(raw)
```

To make sure we don't wrap an array?

/cc @hidde-jan

stevenharman:
A prior change to allow filters to be passed to related resources was incompletely backported from `master` to the `release-0-9` branch (see: 3a691b29adb3e7aff0743ef59e44556bedb260c9). This completes the backport, which moves the `verify_filters` from the `RequestParser` down to the
`Processor`. When `verify_filters` happens in both places, we end up running `verify_filter` twice for each filter, which can result in the filter values being wrapped in nested Arrays.

fixes #1110

### DIFF
--- a/lib/jsonapi/request_parser.rb
+++ b/lib/jsonapi/request_parser.rb
@@ -353,7 +353,7 @@ module JSONAPI
         relationship_type: relationship_type,
         source_klass: @source_klass,
         source_id: @source_id,
-        filters: @source_klass.verify_filters(@filters, @context),
+        filters: @filters,
         sort_criteria: @sort_criteria,
         paginator: @paginator,
         fields: @fields,


stevenharman:
These failures seem to be a problem in minitest rather than with any changes I made. Thoughts?

stevenharman:
Any thoughts on this PR? It would be nice to get the `0.9` series working - at least from the official branch, so we don't have to maintain our own forks. As mentioned in #1110, this change is finishing an incomplete cherry-pick.

stevenharman:
@lgebhardt Sorry to @ you on this one, but any chance of merging this into the `release-0-9` branch to finish the cherry pick? The original author agreed the changes here should have been picked over originally. As for the test failures... thanks minitest?

lgebhardt:
@stevenharman Thanks, and sorry for the delay. Thanks for pointing out the errors in the tests, though I must admit I missed that when i first reviewed this PR.

PR #1130, opened 2017-11-13
Rsheldon/ch5194/fork jr and get rid of serializer shortcut

rsheldiii:
we're doing some funky business in JR with nonprimary keys being the identifier for resources. JR didn't really plan for this in the past, and assumed that underlying DB keys would be the keys of the resources. We've gotten by with monkey patches for the most part now, and this could also be a monkey patch if we wanted, but so far we've monkey patched extension classes and not underlying classes. I figured it'd be better to have this in a fork, since the function no longer exists on master of JR, and if we were to upgrade without a fork we may lose this functionality.


rsheldiii:
whoops, my apologies, wrong source!

Issue #1131, opened 2017-11-14
Remove `get_` prefix from controller actions


scottgonzalez:
The `get_related_resource` and `get_related_resources` actions violate Rubocop's `Style/AccessorMethodName` rule since they start with `get_`.

After a quick discussion with @lgebhardt in Gitter, we came up with the following naming suggestion:

* `get_related_resource` -> `show_related_resource`
* `get_related_resources` -> `index_related_resources`

PR #1132, opened 2017-11-14
Rename controller actions that begin with `get_`

scottgonzalez:
The following methods were renamed to comply with Rubocop's
`Style/AccessorMethodName` rule:

* `get_related_resource` -> `show_related_resource`
* `get_related_resources` -> `index_related_resources`

Fixes gh-1131


lgebhardt:
@scottgonzalez Thanks!

PR #1134, opened 2017-11-17
Processor#show_related_resources pass options to result if they are calculated

senid231:
same behaviour as for Processor#find


lgebhardt:
@senid231 Thanks!

Issue #1058, opened 2017-05-08
Removing /v1 from the links e.g. self in JSON results


ghost:
I am using HTTP headers to add versioning capabilities in my API. I am accessing my API endpoint as such...
`http://localhost:3000/api/posts`
However, for the links in relationship like comments. It shows something like this...
`http://localhost:3000/api/v1/posts/1/relationships/comments`
What I want to do is to have this...
`http://localhost:3000/api/posts/1/relationships/comments`

I read the suggestion from #257 but I could not figure out where to put the monkey patch file (apologies for that.)

Thank you! 

PS: This is connected to these issues: #591 #257 #159

bolaum:
To monkey patch, you can create an initializer like this:
```ruby
# config/initializers/jsonapi_resources_patch.rb

module JSONAPI
  class LinkBuilder
    def formatted_module_path_from_class(klass)
      '/api/'
    end
  end
end
```

Issue #591, opened 2016-01-18
Links broken when used with `scope module:` routes


mitnal:
Hi,

we use a version parameter inside of the `Accept` header to be able to switch Api versions. The different versions are grouped in modules. And to remove these modules as namespaces from the generated URLs we use `scope module:`. This works great but the problem ist that the generated links inside a response still include the module.

The code just checks the module hierarchy of the resource class and completely ignores how rails generates the routes. (See: https://github.com/cerebris/jsonapi-resources/blob/master/lib/jsonapi/link_builder.rb#L102). 
Also see #361 for a similar problem.

As a workaround we monkey patch the `formatted_module_path_from_class` method to return the correct path `\api\`. But that is not an ideal solution.

An example route file:

```
namespace :api do
  scope module: :v1, constraint: ApiConstraint.new(version: 1) do
    jsonapi_resources :my_resource
  end
end
```

And an example resource:

```
module Api
  module V1
    class MyResource < JSONAPI::Resource
      [...]
    end
  end
end
```

So Rails generates routes like: `/api/my-resource`
But the generated links look like this: `/api/v1/my-resource`

Sascha

ghost:
:+1:

I have this issue as well, except I use my version in the url instead of the 'api' namespace.

routes.rb

```
constraints(subdomain: "api") do
  scope module: "api", defaults: { format: :json } do
    namespace :v1 do
      jsonapi_resources :news
    end
  end
end
```

In the JSON, I'm getting links like:

```
self: "http://api.iu.com:5000/api/v1/news/1"
```

When they should look like:

```
self: "http://api.iu.com:5000/v1/news/1"
```

coryondemand:
Also related to:

https://github.com/cerebris/jsonapi-resources/issues/257

It would be nice to have a solution to handle this as we also version our API via headers and do not include the version in the path.

beechnut:
:+1: 

@mitnal: Could you post the code for the monkey patch? Like you say, it's not ideal, but it's better than nothing at the moment.

Thanks in advance!

beechnut:
Borrowing from the gem's `LinkBuilder` tests, I added the following hacks to my app, for the moment. 
With this, my links are pointing to the right place.

I'm matching the slash at the end because doing the one at the beginning would catch the '/api' in http:/**/api**.example.com.

``` ruby
# test/lib/jsonapi_test.rb

require 'test_helper'

[...]

Issue #257, opened 2015-06-16
Namespaced (with custom path option) and scoped routes support


thibaudgg:
My API endpoints are defined like that:

``` ruby
namespace :api, path: '' do
  scope ':customer' do
    api_version(module: 'V1', header: { ... }) do # versionist gem
      jsonapi_resources :users
    end
  end
end
```

This ends up with routes looking like:
 `/:customer/users/:id`
 but JR returns self links like:
`/api/v1/users/:id`

 I managed to fixing that by monkey-patching `JSONAPI::ResourceSerializer` with:

``` ruby
module JSONAPI
  class ResourceSerializer
    def formatted_module_path(source)
      "#{currrent_customer}/"
    end
  end
end
```

but it's less than ideal. It would be nice to have a proper support for custom routes.

It could be fixed by #159 maybe.

Thanks!

lgebhardt:
Ok, I'll revisit #159 soon.

thibaudgg:
@lgebhardt awesome, thanks!

Yoshyn:
Hi.
I use your monkey patch in my codebase and it work well.

But i have an issue with the pagination. it's still generate me url with the `api/v1/` segment.
I was unable to find where theses routes are generated.

Is someone have an idea (monkey-patch or not) to make this work?

pmukerji:
Updated monkey patch that works with JR `0.5.0`:

``` ruby
module JSONAPI
  class LinkBuilder
    def formatted_module_path_from_class(klass)
      "/#{replacement_for_api_v1_here}/"
    end
  end
end
```

webPapaya:
Hey is there a way to access params in LinkBuilder? I have a dynamic route like

webPapaya:
Hey is there a way to access params in LinkBuilder? I have a dynamic route like this

```ruby
namespace :api do
  scope 'membership/:membership_id' do
    jsonapi_resources :users
  end
end
```

webpapaya:
Hey is there a way to access the dynamic route segments in the monkey patch? 

I have a route like this:

``` ruby
namespace :api do
  scope 'memberships/:membership_id' do
    jsonapi_resources :users
  end
end
```

And my monkey patch looks like this

``` ruby
module JSONAPI
  class LinkBuilder
    def formatted_module_path_from_class(klass)
      "/memberships/#{params[:membership_id]}/"
    end
  end
end
```

[...]

PR #1137, opened 2017-12-01
added optional route formatting feature

gatopanx:
Solves #1058 #591 #257 #159.

For when you have an implicit way of accessing a namespace (headers, subdomain, or other)

My use case is:
```
  # NOTE: api.yourapplication.test
  namespace :api, path: '', constraints: { subdomain: 'api' ) } do
    namespace :v1 do
      # ...
    end
  end
```

I created a custom route formatter class:

```
module YourApplication
  class CustomRouteFormatter < JSONAPI::RouteFormatter
    class << self
      def format(route)
        return if route == 'api'
        super
      end

      def unformat(formatted_route)
        super
      end
    end
  end
end
```

and referenced it in my base controller:

```
      def route_formatter
        YourApplication::CustomRouteFormatter
      end
```

### DIFF
--- a/lib/jsonapi/link_builder.rb
+++ b/lib/jsonapi/link_builder.rb
@@ -115,7 +115,7 @@ module JSONAPI
       scopes = module_scopes_from_class(klass)
 
       unless scopes.empty?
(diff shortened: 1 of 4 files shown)


lgebhardt:
@gatopan Thanks!

PR #1143, opened 2017-12-15
Draft - Added HTTP `OPTIONS` endpoint and ability to expose enumerations via it

gatopanx:
Basically allows for exposure of data via options endpoint, facilitates works of API clients.
screen shot 2017-12-14 at 4 53 07 pm


gatopanx:
What do you think @lgebhardt ?

lgebhardt:
I'd say it looks like an interesting idea, but without support in the JSON API spec I'd rather not add this as a feature. It could get added in a different way and then we're going to need to reconcile the differences.

PR #1165, opened 2018-06-20
Allow pagination for nested resources but not for included resources

### LINKED ISSUE #1162: Related resources are not included if there is only one element on page

- [x] Problem/bug report.
- [ ] Feature request.

- [x] I've searched for an existing issue.
- [x] I've asked my question on [Gitter](https://gitter.im/cerebris/jsonapi-resources) and have not received a satisfactory answer.
- [ ] The feature I'm asking for is compliant with the [JSON:API](http://jsonapi.org/) spec.

## Description

When including a resource on a paginated list with only one element on the page, the response does not contain any includes.

Example:
6 Elements in total, 5 per page, page number 2 of some resource.

The reason are the lines:
https://github.com/cerebris/jsonapi-resources/blame/5dfeb0f741e3fc98ee9786f7eb182b15f391ca63/lib/jsonapi/active_relation_resource_finder.rb#L184-L190

```
        paginator = options[:paginator]

        # ToDO: Remove count check. Currently pagination isn't working with multiple source_rids (i.e. it only works
        # for show relationships, not related includes).
        if paginator && source_rids.count == 1
          records = related_klass.apply_pagination(records, paginator, order_options)
        end
```

What is the reason for these lines? Currently, I am not sure what effect it has to purely remove these lines because it does not seem to make a noticeable difference to me (expect for fixing the bug described above).

ArneZsng:
### All Submissions:

### Reviewer Checklist:

Fixes #1162. The issue is that nested resources as well as included resources use `find_related_monomorphic_fragments` in a similar fashion. If we query nested resources, we want pagination to have an effect on the nested resources. If we include resources (at least currently) we don't want the pagination for the main resource to have an effect on the included resources. Thus, we have to differentiate between the two cases.

The issue in #1162 is that if some page of the main resource has only 1 resource, the condition to apply pagination (`source_rids.count == 1`) would be true and thus no includes are loaded.

### DIFF
--- a/lib/jsonapi/active_relation_resource_finder.rb
+++ b/lib/jsonapi/active_relation_resource_finder.rb
@@ -110,13 +110,13 @@ module JSONAPI
       # @return [Hash{ResourceIdentity => {identity: => ResourceIdentity, cache: cache_field, attributes: => {name => value}, related: {relationship_name: [] }}}]
       #    the ResourceInstances matching the filters, sorting, and pagination rules along with any request
       #    additional_field values
-      def find_related_fragments(source_rids, relationship_name, options = {})
+      def find_related_fragments(source_rids, relationship_name, options = {}, included_key = nil)
         relationship = _relationship(relationship_name)
 
         if relationship.polymorphic? && relationship.foreign_key_on == :self
           find_related_polymorphic_fragments(source_rids, relationship, options)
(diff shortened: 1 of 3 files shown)


lgebhardt:
@ArneZsng Thanks!

PR #896, opened 2016-11-03
filtering of included resources is now available. Caching doesn't wor…

beniutek:
…k when we do this and needs to be turned off though

In future will look into caching and check what options are there to make it available

PR #1183, opened 2018-09-19
Implement Included Resource Filtering

### LINKED ISSUE #314: Filtering included resources
We have a use-case where it would be useful to apply filters to the included resources.

Dummy example:

``` ruby
  has_many :socks
  has_many :shoes
end
class SockResource < JSONAPI::Resource
  has_one :user
  filter :in_use
end
class ShoeResource < JSONAPI::Resource
  has_one :user
  filter :in_use
end
```

Individually, one can filter the `shoes` and `socks`
`/shoes?filter[in_use]=true`
`/socks?filter[in_use]=true`

I was wondering if it would be possible to filter the shoes and socks when included with the `User`?

e.g. To get the user, and the shoes and socks he currently has in use, something like:

 `/users/1?include=shoes,socks&filter[shoes][in_use]=true&filter[socks][in_use]=true`

Is this in the scope of jsonapi-resources? or would it be best to find a custom approach?

### LINKED ISSUE #844: Add support for filtering by related resource fields
Would be great if filtering by a related resource's field is possible. Example:

`/comments?filter[todo.name]=todoname`

### LINKED ISSUE #890: Filter included resources or way around it
Hi,

Is there currently a way of filtering resources that are included to our response?

ie: I want to get and include books with that response I'll just do 

`address/shelves?include=books`

but what if I would like to filter by only available books? 

ie: `address/shelves/?include=books&filter[books.available]=true`

books.available is just a scope in this case 

or is the preffered method to do this by declaring additional relationships in the shelf resource?
like below

``` ruby
class ShelfResource < JSONAPI::Resource
  has_many :books
  has_many :available_books
  ...
end
```

If the latter, how do I exactly specify the method for finding resources for this relationship?

I've tried that already by adding 

``` ruby
def records_for_available_books
  _model.books.available
end
```

to the shelf resource but I keep getting errors when doing that.

carlthuringer:
Inspirational Credit and original code: @beniutek

Closes #890
Closes #844
Closes #896
Closes #314 

After much thrashing and some time out in the wilderness with our forked version, I came back here to finish the job, also noticing some bizarre behavior stemming from the original implementation.

Bizarreness:
The loop in the serializer which goes through included resources and adds them to the included resources cache is taking advantage of active record association cacheing such that if any modification of the scope happened during the loading of the primary resource `data`, then the included resources would also follow suit. However, this is broken if certain things are done to the association scope in the `records_for(relation)` public method, which may be overridden in the resource. When such happens, the touched-scope in the cache is lost and the records are retrieved anew. Under those conditions, the blasé attitude of the serializer towards the RequestParser's findings re: filtering results in the queries being re-run without filter, thus you may get more included records than you bargained for if you've tripped over the ActiveRecord Association cache.

### DIFF
--- a/lib/jsonapi/resource.rb
+++ b/lib/jsonapi/resource.rb
@@ -741,6 +741,9 @@ module JSONAPI
           next memo unless relationship && relationship.is_a?(JSONAPI::Relationship::ToMany)
           filtering_resource = relationship.resource_klass
 
+          # Don't try to merge where clauses when relation isn't already being joined to query.
+          next memo unless config[:include_in_join]
+
           filters = config[:include_filters]
           next memo unless filters
 
--- a/lib/jsonapi/resource.rb
+++ b/lib/jsonapi/resource.rb
@@ -732,9 +732,27 @@ module JSONAPI
         records
       end
 
+      def apply_included_resources_filters(records, options = {})
+        include_directives = options[:include_directives]
+        return records unless include_directives
+        related_directives = include_directives.include_directives.fetch(:include_related)
+        related_directives.reduce(records) do |memo, (relationship_name, config)|
+          relationship = _relationship(relationship_name)
+          next memo unless relationship && relationship.is_a?(JSONAPI::Relationship::ToMany)
+          filtering_resource = relationship.resource_klass
+
+          filters = config[:include_filters]
+          next memo unless filters
+
+          rel_records = filtering_resource.apply_filters(filtering_resource.records(options), filters, options).references(relationship_name)
+          memo.merge(rel_records)
+        end
+      end
+
       def filter_records(filters, options, records = records(options))
         records = apply_filters(records, filters, options)
-        apply_includes(records, options)
+        records = apply_includes(records, options)
+        apply_included_resources_filters(records, options)
       end
 
       def sort_records(records, order_options, context = {})
--- a/lib/jsonapi/include_directives.rb
+++ b/lib/jsonapi/include_directives.rb
@@ -40,6 +40,16 @@ module JSONAPI
       delve_paths(get_includes(@include_directives_hash, false))
     end
 
+    def merge_filter(relation, filter)
+      config = include_config(relation.to_sym)
+      config[:include_filters] ||= {}
+      config[:include_filters].merge!(filter)
+    end
+
+    def include_config(relation)
+      @include_directives_hash[:include_related][relation]
+    end
+
     private
 
     def get_related(current_path)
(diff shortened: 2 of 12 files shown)


willisplummer:
Just running into this - would be awesome to get it in!

lgebhardt:
@carlthuringer Thanks!

kdelvare:
How does one add that kind of filter to the allowed filters?

lgebhardt:
@kdelvare You can add the filter as `filter 'tags.name'` to filter to posts that have tags matching a particular name.

lib/jsonapi/resource.rb
```diff
@@ -732,9 +732,30 @@ def apply_filters(records, filters, options = {})
         records
       end
 
+      def apply_included_resources_filters(records, options = {})
+        include_directives = options[:include_directives]
+        return records unless include_directives
+        related_directives = include_directives.include_directives.fetch(:include_related)
+        related_directives.reduce(records) do |memo, (relationship_name, config)|
+          relationship = _relationship(relationship_name)
+          next memo unless relationship && relationship.is_a?(JSONAPI::Relationship::ToMany)
```
josephbridgwaterrowe, line 741: @carlthuringer does this mean that the filtering won't work on a relationship declared as a `has_one`?

I've been trying to filter with `/option_date_defaults?filter[exchange_commodity.code]=XC01`, but the filter is not added to the where statement of the query.

These are my resources:

```ruby
# /app/resource/api/v1/option_date_default_resource.rb
# frozen_string_literal: true

module Api
  module V1
    class OptionDateDefaultResource < JSONAPI::Resource # :nodoc:
      attributes :due_date_month, :exchange_commodity_code, :option_date_month

      filter 'due_date_month'
      filter 'exchange_commodity.code'

      has_one :exchange_commodity

      def exchange_commodity_code
        @model.exchange_commodity.code
      end
    end
  end
end

# /app/resource/api/v1/exchange_commodity_resource.rb
# frozen_string_literal: true

module Api
  module V1
    class ExchangeCommodityResource < JSONAPI::Resource # :nodoc:
      attributes :code, :exchange_code, :exchange_name, :name, :uuid

      filter :code

      has_one :exchange
      has_many :option_date_defaults
    end
  end
end
```

Any clarification or help you could provide would be appreciated.

Thanks.

In the meantime I'm going to use the approach [here](https://github.com/cerebris/jsonapi-resources/issues/890#issuecomment-299910365) from #890

codeguru42:
I couldn't find any references in the documentation that shows how to use this. I finally figured it out, but it would be great to include examples in the docs for users.

Issue #1177, opened 2018-09-12
Compatibility with JSONAPI::Authorization (Resource Caching)


harisadam:
## This issue is a (choose one):

- [x] Feature request.

## Checklist before submitting:

- [x] I've searched for an existing issue.
- [x] The feature I'm asking for is compliant with the [JSON:API](http://jsonapi.org/) spec.

## Description

JSONAPI::Authorization is a pundit based authorization tool that works out-of-the-box with jsonapi-resources' ActiveRecord backed resources. 
The problem is with Cached Resource as it does not have a model key.

JSONAPI::Authorization's related issue https://github.com/venuu/jsonapi-authorization/issues/67

### Features:

Exposing the original model into cache in CachedResourceFramgment(https://github.com/cerebris/jsonapi-resources/blob/master/lib/jsonapi/cached_response_fragment.rb#L87) would solve the problem easily – not sure about the cons.

I have a working solution in my fork so I can submit a PR If you are OK with the idea.

adambedford:
I'd also love to get resource caching working with authorization. Any updates here?

PR #1185, opened 2018-09-25
write model into cache, fix Rails 5.2 deprecations (#1177)

harisadam:
more: https://github.com/cerebris/jsonapi-resources/issues/1177

### DIFF
--- a/lib/jsonapi/cached_resource_fragment.rb
+++ b/lib/jsonapi/cached_resource_fragment.rb
@@ -29,16 +29,17 @@ module JSONAPI
       return results
     end
 
-    attr_reader :resource_klass, :id, :type, :context, :fetchable_fields, :relationships,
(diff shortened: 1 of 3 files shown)


lgebhardt:
@harisadam I'm hesitant to stuff the model into the cache for both efficiency and consistency reasons. I haven't investigated why jsonapi-authorization is calling the _model method on the CachedResourceFragment but it seems like it would be better to detect the type and assume it's authorized for the user (if that's what's going on). If the `attribute_caching_context` is set correctly I think this would be a safe strategy. See http://jsonapi-resources.com/v0.9/guide/resource_caching.html

adambedford:
@lgebhardt It looks like this approach may not be the most suitable but do you by chance have guidance on how to get jsonapi-authorization to play nicely with resource caching?

PR #1186, opened 2018-09-25
Fix inconsistent type of filters

carlthuringer:
Another issue we run into is where the type of the `value` passed to a custom filter proc/method is sometimes just the bare value, not an array-wrapped value. This is because the framework does not pass verified_filters to filter records even though it was already calculated previously in `show_related_resources`.

I would have just manifested this bug by using the existing `banned` filter on the books related to the author, but cannot because of the second (bug?) that `rel_opts` is not passed to `filter_records`, so the `current_user.book_admin` is not available. The manifestation there is that due to a poorly-written filter, `records` is nullified and `.sort` is called on nil. Therefore I replicated the `banned` filter, but less the admin check.

The manifest bug is that the `meta: { record_count: 2 }` where it should be 1. It's not consistent with the length of the `data` collection. With the fix, the bug does not happen anymore. In our platform we have a much more serious crash where a bad type is sanitized and used in a query param.

### DIFF
--- a/lib/jsonapi/processor.rb
+++ b/lib/jsonapi/processor.rb
@@ -197,7 +197,7 @@ module JSONAPI
           (paginator && paginator.class.requires_record_count) ||
           (JSONAPI.configuration.top_level_meta_include_page_count))
         related_resource_records = source_resource.public_send("records_for_" + relationship_type)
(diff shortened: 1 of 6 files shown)


lgebhardt:
@carlthuringer Thanks!

PR #1218, opened 2019-02-07
Always include to_one linkage

lgebhardt:
This PR reenables the `always_include_to_one_linkage_data` configuration option. This joins in the to_one relationship in order to retrieve the linkage data without additional database requests. 

The path parsing logic for includes, filters and sorts was moved to a Path class that parses into Relationship and Field path parts.

Filtering is extended across polymorphic relationships.

Enables the a new `join_left` ponyfill to extend left_joins to rails 4 and works around an issue with nested joins not aliasing correctly on rails 5 and 5.1.

Streamlines the JoinTree and Joins handling of polymorphic relationships in paths. Filters can now specify the specific resource type of a polymorphic relation using the format `relationshipname#resource_type`. For example `pictures.imageable#document.name`

### All Submissions:

### New Feature Submissions:

### Reviewer Checklist:


test/test_helper.rb
```diff
@@ -596,7 +594,7 @@ def assert_cacheable_get(action, *args)
       if mode == :all
         # TODO Should also be caching :show_related_resource (non-plural) action
```
joegaudet: Just a question about this comment, is caching show_related_resource still disabled in 10?

lib/jsonapi/path_part.rb
```diff
@@ -0,0 +1,66 @@
+module JSONAPI
+  class PathPart
+    def self.parse(source_resource_klass:, part_string:, parse_fields: true)
+      first_part, last_part = part_string.split('#', 2)
+      relationship = source_resource_klass._relationship(first_part)
+
+      if relationship
+        if last_part
+          resource_klass = source_resource_klass.resource_klass_for(last_part)
+          # ToDo: compare to relationship and raise error if not a match?
```
joegaudet: Missing todo?

lib/jsonapi/path_part.rb
```diff
@@ -0,0 +1,66 @@
...
+      def path_specified_resource_klass?
+        !@resource_klass.nil?
+      end
+    end
+
+    class Field
+      attr_reader :resource_klass, :field_name
+
+      def initialize(resource_klass:, field_name:)
+        # ToDo: Should we enforce the resource has the field?
```
joegaudet: Todo.

lib/jsonapi/path.rb
```diff
@@ -0,0 +1,41 @@
...
+
+      if ensure_default_field && parse_fields && @parts.last.is_a?(PathPart::Relationship)
+        last = @parts.last
+        @parts << PathPart::Field.new(resource_klass: last.resource_klass,
+                                      field_name: last.resource_klass._primary_key)
+      end
+    end
+
+    def relationship_parts
+      relationships = []
```
joegaudet:
```suggestion
      relationships = @parse.select {|p| p.is_a?(PathPart::Relationship)}
```

lib/jsonapi/path.rb
```diff
@@ -0,0 +1,41 @@
...
+      @parts.each do |part|
+        relationships << part if part.is_a?(PathPart::Relationship)
+      end
+      relationships
+    end
+
+    def relationship_path_string
+      relationship_parts.collect do |part|
```
joegaudet:
```suggestion
      relationship_parts.collect(&:to_s).join('.')
```

lib/jsonapi/active_relation_resource_finder/join_tree.rb
```diff
@@ -4,22 +4,116 @@ class JoinTree
...
+        if part.is_a?(PathPart::Relationship)
+          node[:resource_klasses][resource_klass][:relationships][part.relationship] ||= {}
+
+          # join polymorphic as left joins
+          node[:resource_klasses][resource_klass][:relationships][part.relationship][:join_type] ||=
+              part.relationship.polymorphic? ? default_polymorphic_join_type : default_join_type
+
+          part.relationship.resource_types.each do |related_resource_type|
+            related_resource_klass = resource_klass.resource_klass_for(related_resource_type)
+            if !part.path_specified_resource_klass? || related_resource_klass == part.resource_klass
```
joegaudet: Matter of style, but perhaps pull these individual booleans out of the if condition giving them names more descriptive of their intent?

lib/jsonapi/relationship.rb
```diff
@@ -142,6 +161,18 @@ def initialize(name, options = {})
...
 
+      def to_s
+        # :nocov: useful for debugging
+        "#{parent_resource._type}.#{name} => (ToMany) #{resource_klass._type}"
+        # :nocov:
+      end
+
+      def include_optional_linkage_data?
+        # :nocov:
+        @always_include_optional_linkage_data || JSONAPI::configuration.always_include_to_many_linkage_data
```
joegaudet, line 172: So there's no way for me to say that the broad configuration is always included, but for one relationship do not? Not that I think this is a likely case, but a case nonetheless.

lib/jsonapi/resource.rb
```diff
@@ -902,6 +904,34 @@ def paginator(paginator)
...
+
+      def polymorphic(polymorphic = true)
+        @_polymorphic = polymorphic
+      end
+
+      def _polymorphic_types
+        @poly_hash ||= {}.tap do |hash|
+          ObjectSpace.each_object do |klass|
+            next unless Module === klass
+            if ActiveRecord::Base > klass
```
joegaudet: point of style, I would think ```klass < ActiveRecord::Base``` would be more intuitive. That's at least how we've done this in our monkey patch  of this code ;)

joegaudet (commented):
A few comments, largely looks good. 

The polymorphic resource specification stuff is interesting, I take it not part of the JSONAPI spec though?

dgeb:
Let's deprecate `Resource#find_records` properly to discourage its use.

lib/jsonapi/resource.rb
```diff
@@ -710,15 +710,15 @@ def fields
...
+     def resources_for(model_instances, context)
+       model_instances.collect do |model_instance|
+          resource_for(model_instance, context)
         end
       end
 
-      def resource_for(model_record, context)
-        resource_klass = self.resource_klass_for_model(model_record)
-        resource_klass.new(model_record, context)
+      def resource_for(model_instance, context)
```
dgeb: maybe `record` instead of `model_instance`

lib/jsonapi/resource.rb
```diff
@@ -710,15 +710,15 @@ def fields
...
+        resource_klass = self.resource_klass_for_model(model_instance)
```
dgeb: maybe `resource_klass_for` instead of `resource_klass_for_model`?

lib/jsonapi/path_part.rb
```diff
@@ -0,0 +1,62 @@
+module JSONAPI
+  class PathPart
```
dgeb: perhaps `PathSegment` instead of `PathPart`?

lib/jsonapi/active_relation_resource_finder/join_tree.rb
```diff
@@ -92,35 +154,69 @@ def relation_join_hash(path, path_hash = {})
...
-            current_relation_path << source_relationship.relation_name(options)
+      def construct_joins(node, current_relation_path = [], current_relationship_path = [])
+        node.each do |relationship, relationship_details|
+          join_type = relationship_details[:join_type]
+          if relationship == :root
+            @joins[:root] = {alias: resource_klass._table_name, join_type: :root}
+
+            # alias to the default table unless a source_relationship is specified
+            unless source_relationship
+              @joins[''] = {alias: resource_klass._table_name, join_type: :root}
```
dgeb, line 173: Would it make sense to use `@joins[:default]` instead of `@joins['']`? Even if we stick with an empty string, I'd be more comfortable sharing a constant with some semantic meaning so that I understand the intent.

lib/jsonapi/relationship.rb
lgebhardt: It's not supported for ActiveRecord whether set globally or per relationship. The code just doesn't exist yet, and it's going to potentially be expensive so I'm not planning to tackle it soon.
What do you think the best way to handle this in the relationship would be? Conceivably we could write a Mongo adapter that easily supports this. I want the `Relationship` to work with either Adapter. We could interrogate the Adapter for capabilities.

lib/jsonapi/active_relation_resource_finder/join_tree.rb
lgebhardt: We need the string for polymorphic support to specify the related resource type, like '#documents'.

lib/jsonapi/path_part.rb
lgebhardt: Good idea. Fixed

lib/jsonapi/resource.rb
lgebhardt: I reverted this commit.

lib/jsonapi/resource.rb
lgebhardt: I reverted this commit.

lib/jsonapi/resource.rb
lgebhardt: Fixed! Thanks.

lib/jsonapi/path.rb
lgebhardt: Nice improvement - thanks.

lib/jsonapi/path_part.rb
lgebhardt: Fixed - decided to implement the check mentioned in the todo.

test/test_helper.rb
lgebhardt: Good catch - I had forgotten about that one. I added the method.

lib/jsonapi/path.rb
lgebhardt: Modified

lib/jsonapi/active_relation_resource_finder/join_tree.rb
lgebhardt: Good call. I added a comment and hope clarified the logic.

lgebhardt:
@joegaudet, @dgeb Thanks for the reviews. I think I've got all the suggestions implemented or addressed, with the exception of deprecating the `find_records` method. I'm still looking for a good way to handle that so people who have it overridden can be notified they may need to change it. As it is now it will at least fail so I haven't renamed the method. Any suggestions?

dgeb (approved):

PR #1221, opened 2019-02-21
Add relationship apply_join callable for custom joins

lgebhardt:
## Overview

This PR adds back a similar functionality to overriding the `records_for` method found in earlier releases. This allows you to customize the way relationships are joined for includes, to_one linkage, filters and sorts.

For example you can restrict the comments to only approved comments:
```ruby
      has_many :comments, apply_join: -> (records, relationship, resource_type, join_type, options) {
        case join_type
        when :inner
          records = records.joins(relationship.relation_name(options))
        when :left
          records = records.joins_left(relationship.relation_name(options))
        end
        # Note that the`apply_join` method can perform the above join logic if all you need is to add additional filters.
        # records = apply_join(records: records, relationship: relationship, resource_type: resource_type, join_type: join_type, options: options)

        records.where(comments: {approved: true})
      }
    end
```

`records` passed in is from the resource's parent resource's `records` method. The `relation` returned from the callable is then merged with the parent relation.

Of course you can continue to use the `relation_name` option on relationships to specify scopes from the model.

## JoinTree changes

`JoinTree` was renamed to `JoinManager` and the alias tracking and join applying logic was moved to this class.

**Note:** The order the joins is made in has been adjusted so those closest to the trunk are made first, rather than walking down each branch to the leaves and then back to the next branch. This helps prevent rails from reorder the joins, and thus its assigned aliases, as more joins are merged. If the same alias is detected (indicating a reordering of the joins) an exception is raised. 

I would appreciate feedback on whether we should be handling collisions in a different way. This should be a rare occurrence, and I think can only happen when the same table is joined in twice at the same level.

### All Submissions:

### New Feature Submissions:

### Reviewer Checklist:


senid231:
@lgebhardt what about apply_includes ?

I think we also need a possibility to override default `records.includes(:relation_name)`

lgebhardt:
@senid231 The release handles includes (as in included resources, not the `includes` ActiveRelation method) quite differently than earlier versions. In earlier releases the models were loaded and the serializer walked the included relationships loading the related models as it went. This required using `includes` on `records` to allow rails to efficiently load the related models. 

This release takes a different approach and finds just the ids and cache keys (if caching) that match the request. This is done using pluck so no models are loaded at this stage. The included resource ids are also pulled using pluck for each included relationship. This set of ids is first populated from the cache (if caching), and any cache misses are loaded as model instances with one query per type. This mostly negates the need to call `includes`. If your model does need included records, beyond what the Resource uses for relationships, you can add the `includes` call by overriding the the `self.records` call. This might be needed if your model has some computed values that pull from other models.

lib/jsonapi/active_relation_resource_finder/join_manager.rb
```diff
@@ -0,0 +1,288 @@
...
+        join_array
+      end
+
+      def add_join_details(join_key, details, check_for_duplicate_alias = true)
+        fail "details already set" if @join_details.has_key?(join_key)
+        @join_details[join_key] = details
+
+        if check_for_duplicate_alias && @collected_aliases.include?(details[:alias])
+          fail "alias '#{details[:alias]}' has already been added. Possible relation reordering"
+        end
```
dgeb, line 135: Documentation needed

dgeb (approved):
Looks good once some documentation is added related to when/why the join alias duplication check might fail.

lib/jsonapi/active_relation_resource_finder/join_manager.rb
lgebhardt: Good idea. Added.

PR #1230, opened 2019-03-14
Allow setting of the resource finder on a per resource basis

lgebhardt:
This gets us closer to supporting alternate resource finders. This should work with multiple resource finders as long as relationships do not span the different data stores.

I wanted to keep the resource finders as mix-ins to allow existing code to override the `records` and other methods. In order to support a default resource finder and allow setting this with a method in the resource class definition the mix-in needed to delay loading. This was accomplished with `method_missing` to load the resource finder if it isn't loaded.

### All Submissions:

### New Feature Submissions:

### Reviewer Checklist:

### DIFF
--- a/lib/jsonapi/resource.rb
+++ b/lib/jsonapi/resource.rb
@@ -421,74 +421,20 @@ module JSONAPI
           subclass.attribute :id, format: :id, readonly: true
         end
 
+        # Ensure that the parent's resource finder is included before inheriting from the parent is completed
+        if !_resource_finder_included && self != JSONAPI::Resource
+          include_resource_finder
+        end
+        subclass._resource_finder_included = _resource_finder_included
+
         check_reserved_resource_name(subclass._type, subclass.name)
-
-        subclass.include JSONAPI.configuration.default_resource_finder if JSONAPI.configuration.default_resource_finder
       end
 
-      # A ResourceFinder is a mixin that adds functionality to find Resources and Resource Fragments
-      # to the core Resource class.
-      #
-      # Resource fragments are a hash with the following format:
-      # {
-      #   identity: <required: a ResourceIdentity>,
-      #   cache: <optional: the resource's cache value>
-      #   attributes: <optional: attributes hash for attributes requested - currently unused>
-      #   related: {
-      #     <relationship_name>: <ResourceIdentity of a source resource in find_included_fragments>
-      #   }
-      # }
-      #
-      # begin ResourceFinder Abstract methods
-      def find(_filters, _options = {})
-        # :nocov:
-        raise 'Abstract ResourceFinder method called. Ensure that a ResourceFinder has been set.'
-        # :nocov:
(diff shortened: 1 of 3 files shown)


lib/jsonapi/resource.rb
```diff
@@ -1071,6 +1018,33 @@ def register_relationship(name, relationship_object)
...
+        end
+        include _resource_finder
+        self._resource_finder_included = true
+      end
+
+      def method_missing(m, *args, &block)
+        if _resource_finder_included
+          super
+        else
+          # handle the case where a resource finder has not been include yet. This should only happen once per class.
```
scottgonzalez: s/include/included/

s/handle/Handle/ since this is a full sentence.

lib/jsonapi/resource.rb
```diff
@@ -1071,6 +1018,33 @@ def register_relationship(name, relationship_object)
 
       private
 
+      def _resource_finder=(resource_finder)
```
scottgonzalez: Is this only so you can clear the value on inheritance?

lib/jsonapi/resource.rb
```diff
@@ -421,75 +421,22 @@ def inherited(subclass)
...
-      def find_included_fragments(_source_rids, _relationship_name, _options = {})
-        # :nocov:
-        raise 'Abstract ResourceFinder method called. Ensure that a ResourceFinder has been set.'
-        # :nocov:
-      end
+        # Ensure that the parent's resource finder is included before inheriting from the parent is completed
+        if !_resource_finder_included && self != JSONAPI::Resource
+          include_resource_finder
+        end
+        subclass._resource_finder = nil
```
scottgonzalez: Why is this needed? Shouldn't this already be `nil` at this point because the interpreter hasn't gotten past `class SubClass < ParentClass`? I didn't see anything that would indicate the resource finder is being copied from the parent to the child.

scottgonzalez (changes requested):

lib/jsonapi/resource.rb
lgebhardt: Good catch. This, and the call you caught above, were left over from an earlier approach. I've removed them.

lib/jsonapi/resource.rb
lgebhardt: Removed - see below.

lib/jsonapi/resource.rb
```diff
@@ -1070,6 +1016,28 @@ def register_relationship(name, relationship_object)
...
+        end
+        include _resource_finder
+        self._resource_finder_included = true
+      end
+
+      def method_missing(m, *args, &block)
+        if _resource_finder_included
+          super
+        else
+          # Handle the case where a resource finder has not been include yet. This should only happen once per class.
```
scottgonzalez: You missed changing "include" to "included"

lgebhardt: D'oh. Fixed

scottgonzalez (approved):

PR #1217, opened 2019-01-28
Support creating polymorphic has-many relationships

saravanak:
Consider a resource which has a polymorphic has-many relation . 

Existing behavior: While creating the payload for the has-many relation should have the base class name as the `type` field in the `relationships` post data. 

New behavior: Allows creating a resource which has a polymorphic has-many relation even with the subclass type. 

Why: Because ember sends in the subclass `modelName` (the instance's `modelName`) while creating a resource of the said specifications.

I have raised a PR against the `0-9` branch as this is the stable branch and this version was working fine against our codebase (master had some errors which I did not spend time on investigating any further..)

### All Submissions:

- [x] I've checked to ensure there aren't other open [Pull Requests](https://github.com/cerebris/jsonapi-resources/pulls) for the same update/change.
- [ ] I've submitted a [ticket](https://github.com/cerebris/jsonapi-resources/issues) for my issue if one did not already exist.
Per https://github.com/cerebris/jsonapi-resources/blob/v0.9.3/lib/jsonapi/request_parser.rb#L489-L502, this issue is evident from the code, and I stumbled on it on my usecase.
- [x] My submission passes all tests. (Please run the full test suite locally to cut down on noise from travis failures.)
- [ ] I've used Github [auto-closing keywords](https://help.github.com/articles/closing-issues-via-commit-messages/) in the commit message or the description.
- [x] I've added/updated tests for this change.

### New Feature Submissions:

- [ ] I've submitted an issue that describes this feature, and received the go ahead from the maintainers.
- [ ] My submission includes new tests.
- [ ] My submission maintains compliance with [JSON:API](http://jsonapi.org/).

### Bug fixes and Changes to Core Features:

- [x] I've included an explanation of what the changes do and why I'd like you to include them.
- [x] I've provided test(s) that fails without the change.

### Reviewer Checklist:
- [ ] Maintains compliance with JSON:API
- [ ] Adequate test coverage exists to prevent regressions

PR #1231, opened 2019-03-15
Continuation of #1217: Support creating polymorphic has-many relationships

st0012:
This is a continuation of #1217, which fixes its failing tests (thanks for @saravanak's hard work!). 

The cause of #1217's failure is data mutation between test cases: the integration test's post request creates a new record, and it fails other tests (under certain execution order). So my additional commit is to keep the database clean (with [database_cleaner](https://github.com/DatabaseCleaner/database_cleaner) before/after each integration test suite.

### All Submissions:

### New Feature Submissions:

### Reviewer Checklist:

### DIFF
--- a/lib/jsonapi/request_parser.rb
+++ b/lib/jsonapi/request_parser.rb
@@ -524,13 +524,27 @@ module JSONAPI
       if links_object.length == 0
         add_result.call([])
       else
-        defined_relationship_resource = Resource.resource_for(@resource_klass.module_path + unformat_key(relationship.type).to_s)
         links_object.each_pair do |type, keys|
-          relationship_resource = Resource.resource_for(@resource_klass.module_path + unformat_key(type).to_s)
-          if !relationship_resource.ancestors.include?(defined_relationship_resource)
-            fail JSONAPI::Exceptions::TypeMismatch.new(links_object[:type])
+          resource = self.resource_klass || Resource
+          type_name = unformat_key(type).to_s
+          relationship_resource_klass = resource.resource_for(relationship.class_name)
+          relationship_klass = relationship_resource_klass._model_class
+          linkage_object_resource_klass = resource.resource_for(type_name)
+          linkage_object_klass = linkage_object_resource_klass._model_class
+
+          unless linkage_object_klass == relationship_klass || linkage_object_klass.in?(relationship_klass.subclasses)
+            fail JSONAPI::Exceptions::TypeMismatch.new(type_name)
           end
-          add_result.call relationship_resource.verify_keys(keys, @context)
+
+          relationship_ids = relationship_resource_klass.verify_keys(keys, @context)
+
+          result = if relationship.polymorphic?
+            { type: type, ids: relationship_ids }
+          else
+            relationship_ids
+          end
+
+          add_result.call result
         end
       end
     end
--- a/lib/jsonapi/request_parser.rb
+++ b/lib/jsonapi/request_parser.rb
@@ -524,27 +524,35 @@ module JSONAPI
       if links_object.length == 0
         add_result.call([])
       else
-        links_object.each_pair do |type, keys|
-          resource = self.resource_klass || Resource
-          type_name = unformat_key(type).to_s
-          relationship_resource_klass = resource.resource_for(relationship.class_name)
-          relationship_klass = relationship_resource_klass._model_class
-          linkage_object_resource_klass = resource.resource_for(type_name)
-          linkage_object_klass = linkage_object_resource_klass._model_class
+        if relationship.polymorphic?
+          polymorphic_results = []
 
-          unless linkage_object_klass == relationship_klass || linkage_object_klass.in?(relationship_klass.subclasses)
-            fail JSONAPI::Exceptions::TypeMismatch.new(type_name)
+          links_object.each_pair do |type, keys|
+            resource = self.resource_klass || Resource
+            type_name = unformat_key(type).to_s
+            relationship_resource_klass = resource.resource_for(relationship.class_name)
+            relationship_klass = relationship_resource_klass._model_class
+            linkage_object_resource_klass = resource.resource_for(type_name)
+            linkage_object_klass = linkage_object_resource_klass._model_class
+
+            unless linkage_object_klass == relationship_klass || linkage_object_klass.in?(relationship_klass.subclasses)
+              fail JSONAPI::Exceptions::TypeMismatch.new(type_name)
+            end
+
+            relationship_ids = relationship_resource_klass.verify_keys(keys, @context)
+             polymorphic_results << { type: type, ids: relationship_ids }
           end
 
-          relationship_ids = relationship_resource_klass.verify_keys(keys, @context)
-
-          result = if relationship.polymorphic?
-            { type: type, ids: relationship_ids }
-          else
-            relationship_ids
+          add_result.call polymorphic_results
+        else
+          if links_object.length > 1 || !links_object.has_key?(unformat_key(relationship.type).to_s)
+            fail JSONAPI::Exceptions::TypeMismatch.new(links_object[:type])
           end
 
-          add_result.call result
+          links_object.each_pair do |type, keys|
+            relationship_resource = Resource.resource_for(@resource_klass.module_path + unformat_key(type).to_s)
+            add_result.call relationship_resource.verify_keys(keys, @context)
+          end
         end
       end
     end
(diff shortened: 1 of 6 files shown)


lib/jsonapi/request_parser.rb
```diff
@@ -521,19 +521,15 @@ def parse_to_many_relationship(link_value, relationship, &add_result)
...
-          fail JSONAPI::Exceptions::TypeMismatch.new(links_object[:type])
-        end
-
+        defined_relationship_resource = Resource.resource_for(@resource_klass.module_path + unformat_key(relationship.type).to_s)
         links_object.each_pair do |type, keys|
           relationship_resource = Resource.resource_for(@resource_klass.module_path + unformat_key(type).to_s)
+          if !relationship_resource.ancestors.include?(defined_relationship_resource)
+            fail JSONAPI::Exceptions::TypeMismatch.new(links_object[:type])
+          end
           add_result.call relationship_resource.verify_keys(keys, @context)
```
lgebhardt: This is going to call `add_result` for each type, which will overwrite the previous type. In addition the type information will be lost. Look at line 503 for how `parse_to_one_relationship` handles this. Then `Resource._replace_fields` will need some changes.

test/integration/requests/request_test.rb
```diff
@@ -211,6 +211,52 @@ def test_post_single
...
+        }
+      }.to_json,
+      headers: {
+        'CONTENT_TYPE' => JSONAPI::MEDIA_TYPE,
+        'Accept' => JSONAPI::MEDIA_TYPE
+      }
+
+    assert_jsonapi_response 400, msg: "Submitting a thing as a vehicle should raise a type mismatch error"
+  end
+
```
lgebhardt, line 277: Let's test cases where multiple vehicles are provided. Tests should include mismatches of types, such as 
```
'relationships' => {
             'vehicles' => {'data' => [{'type' => 'car', 'id' => '1'}, {'type' => 'car', 'id' => '2'}]}, #vehicle 2 is actually a boat
           }
```

Test should probably be repeated for updates, though as of now they should go through the same path.

test/integration/requests/request_test.rb
```diff
@@ -211,6 +211,52 @@ def test_post_single
...
+            'vehicles' => {'data' => [{'type' => 'car', 'id' => '1'}]},
+          }
+        }
+      }.to_json,
+      headers: {
+        'CONTENT_TYPE' => JSONAPI::MEDIA_TYPE,
+        'Accept' => JSONAPI::MEDIA_TYPE
+      }
+
+    assert_jsonapi_response 201
```
lgebhardt, line 244: Ensure that the changes went through as requested. The code will currently fail if mixed vehicle types are provided.

lgebhardt (changes requested):
Thanks @saravanak and @st0012 This is a good start, but it's not thoroughly tested and further work is needed downstream in the Resource for creating the relationship links.

lib/jsonapi/resource.rb
```diff
@@ -302,6 +302,20 @@ def _replace_to_many_links(relationship_type, relationship_key_values, options)
         to_add = relationship_key_values - (relationship_key_values & existing)
         _create_to_many_links(relationship_type, to_add, {})
 
+        @reload_needed = true
+      elsif relationship.polymorphic?
+        relationship_resource_klass = self.class.resource_for(relationship_key_values[:type])
+        relationship_klass = relationship_resource_klass.model_hint.classify.safe_constantize
```
st0012: @lgebhardt have one question here: when `relationship_resource_klass` is a subclass resource (like `CarResource`), its `_model_class` still returns `Vehicle` instead of `Car`. Is this intentional? Is there an easier way for me to get `Car` from `CarResource`?

lib/jsonapi/resource.rb
```diff
@@ -302,6 +302,20 @@ def _replace_to_many_links(relationship_type, relationship_key_values, options)
...
+        relation_name = relationship.relation_name(context: @context)
+
+        begin
+          related_records = relationship_klass.find(relationship_key_values[:ids])
+        rescue ActiveRecord::RecordNotFound
+          fail JSONAPI::Exceptions::RecordNotFound.new(relationship_key_values[:ids])
```
st0012: The reason to raise RecordNotFound instead of TypeMismatch is that type mismatch requires 2 conditions:
      - the id can't be used to find any record in current subclass (Car)
      - but in the meantime, it does exists for the superclass (Vehicle)
And this means we need to perform additional query to prove the second condition, which is kind of a waste.

lgebhardt: I tend to agree it's kind of a waste, but from a JSON API perspective it's not known that the backing table is sharing a common set of IDs. So there really isn't a Car with id 2 so the system should raise an error.

lgebhardt: Let's use `find_by_keys`, like https://github.com/cerebris/jsonapi-resources/blob/master/lib/jsonapi/resource.rb#L251. That way any overrides will get applied. This might be the case if people are doing permission checking for these related resources.

lib/jsonapi/resource.rb
lgebhardt: This wasn't intentional. It's happening because the CarResource inherits from VehicleResource and the model_name is inherited. Changing this will break a lot of the tests and probably production code.

So as a workaround I propose we add model_name calls to CarResource and BoatResource:
```ruby
class CarResource < VehicleResource
  model_name 'Car'
  attributes :drive_layout
end

class BoatResource < VehicleResource
  model_name 'Boat'
  attributes :length_at_water_line
end
```

and clear `@model_class` when the `model_name` is set in `Resource. model_name`:
```ruby

     def model_name(model, options = {})
        @model_class = nil #this is new
        @_model_name = model.to_sym

        model_hint(model: @_model_name, resource: self) unless options[:add_model_hint] == false

        rebuild_relationships(_relationships)
      end
```

That should allow you to use `_model_class` instead.

I'll think about a better solution as I don't like needing to specify the `model_name`.

lib/jsonapi/resource.rb
```diff
@@ -302,6 +302,24 @@ def _replace_to_many_links(relationship_type, relationship_key_values, options)
         to_add = relationship_key_values - (relationship_key_values & existing)
         _create_to_many_links(relationship_type, to_add, {})
 
+        @reload_needed = true
+      elsif relationship.polymorphic?
+        relationship_resource_klass = self.class.resource_for(relationship_key_values[:type])
+        ids = relationship_key_values[:ids]
+
+        related_records = relationship_resource_klass
+          .records
```
st0012: @lgebhardt for supporting customizations, I think using `records` is enough. Use `find_by_keys` will also run includes for this query, which is unncessary.

st0012:
@lgebhardt added tests for update, is there anything missing for merging?

lib/jsonapi/resource.rb
lgebhardt: I missed this earlier: `records` needs to be called as `records(options)`

lib/jsonapi/resource.rb
```diff
@@ -302,6 +302,24 @@ def _replace_to_many_links(relationship_type, relationship_key_values, options)
...
+          .where({relationship_resource_klass._primary_key => ids})
+
+        missed_ids = ids - related_records.map(&:id)
```
lgebhardt: We need to account for alternate primary keys, and I think we can skip creating the models so pluck will save a few cycles.
```ruby
missed_ids = ids - related_records.pluck(relationship_resource_klass._primary_key)
```

test/integration/requests/request_test.rb
```diff
@@ -211,6 +216,93 @@ def test_post_single
...
+
+    body = JSON.parse(response.body)
+    person = Person.find(body.dig("data", "id"))
+
+    assert "Reo", person.name
+    assert 2, person.vehicles.count
```
lgebhardt: These asserts should be `assert_equal`. Once fixed this test is failing with only the last vehicle type being created.

lgebhardt (changes requested):
@st0012 There are a few test failures masked by using `assert` instead of `assert_equal`.

test/integration/requests/request_test.rb
st0012: ah sorry 😅

lib/jsonapi/resource.rb
st0012: 👍

st0012:
@lgebhardt all updated!

lgebhardt:
@st0012 Thanks for working on this. I squashed your changes and pushed them to release-0-9

PR #1234, opened 2019-03-21
Pass `rel_opts` when filtering records for record count

hatch-carl:
Naturally I uncovered this because it caused a more complex bug in my app.
The problem here is that if you have a filter which depends on something
that is set in, for example, the context, you will not have that context
when the filter is re-invoked for record count, which happens when the
record count configuration is active. This only happens when fetching
related resources.

In this case I add a test where we try to fetch an author's banned books,
of which there should be only one. However, because the prior code
discards the rel_opts and instead provides an empty hash (!) the filter
does not execute because it will only operate when an admin user is in
the context. The failing test then reports the record count as *3*, where it should be *1*. The change to the processor class solves this issue.


lgebhardt:
@hatch-carl Thanks!

PR #1235, opened 2019-03-21
Include directives may be nil if no includes are directed

hatch-carl:
If you attempt to filter an included resource, but didn't include anything, it will crash.

If you had included a different resource this would return an error response. With this change, we safe-navigate the `@include_directives`, avoiding a NoMethodError on nil.

### DIFF
--- a/test/controllers/controller_test.rb
+++ b/test/controllers/controller_test.rb
@@ -2638,7 +2638,7 @@ class Api::V5::PaintersControllerTest < ActionController::TestCase
 
   def test_show_with_filters_and_no_included_resources
     get :show, params: { id: 1, filter: { 'paintings.category' => 'oil' } }
-    assert_response 400
+    assert_response :bad_request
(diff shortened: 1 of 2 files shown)


test/controllers/controller_test.rb
```diff
@@ -2635,6 +2635,13 @@ def test_show_with_filters_and_included_resources_with_filters
     assert_equal 2, json_response['included'].size
     assert_equal '4', json_response['included'][0]['id']
   end
+
+  def test_show_with_filters_and_no_included_resources
+    get :show, params: { id: 1, filter: { 'paintings.category' => 'oil' } }
+    assert_response 400
```
scottgonzalez: Use `:bad_request` instead of `400`.

scottgonzalez (changes requested):
Looks good to me. Just one minor change.

lgebhardt:
@hatch-carl Thanks!

PR #1236, opened 2019-03-22
Allow marking relationships as async

krasnoukhov:
### Description

This change introduces an option to mark relationships as non-async (globally and for each relationship separately) so the links to these relationships won't be included in the response.

It's meant to be used in conjunction with route definition that takes a custom block:

```ruby
jsonapi_resources(:posts, only: [:index, :show]) {
  # We don't call `jsonapi_related_resources` here
}
```

This is useful if one wants to avoid async requests to fetch the relationships and rather utilize sideloading.

Summary:

* Global config to allow marking relationships as non-async by default
* Do not return links for non-async relationships
* In a case when non-async relationship is not sideloaded, it does not have any data/links.
  In this case we don't want to mention it at all since ember-data currently resets previous known state.

### All Submissions:

### New Feature Submissions:

- [x] My submission includes new tests.
- [x] My submission maintains compliance with [JSON:API](http://jsonapi.org/).

### Reviewer Checklist:

### DIFF
--- a/lib/jsonapi/configuration.rb
+++ b/lib/jsonapi/configuration.rb
@@ -27,6 +27,7 @@ module JSONAPI
                 :whitelist_all_exceptions,
                 :always_include_to_one_linkage_data,
                 :always_include_to_many_linkage_data,
+                :relationships_async_default,
                 :cache_formatters,
                 :use_relationship_reflection,
                 :resource_cache,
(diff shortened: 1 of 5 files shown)


lgebhardt:
@krasnoukhov Could you please write up a description of this feature?

lgebhardt:
We're going to handle this by allowing control over the generation of links as the relationship is defined. We can also use this to control the generation of the relationship routes.

PR #1246, opened 2019-04-23
Route based links - master

lgebhardt:
Updates the links generation in LinkBuilder to use the routes url helper functions.

Adds `build_default_links` flags for resources and relationships to turn off building the default links.

When a route helper function can not be found the link is not built and a warning is emitted.

Updates the tests to work with this. This is the bulk of the changes.

Adds a `singleton` flag for singleton resources and a configurable way to resolve the singleton ids.

NaiveCache has been removed from the LinksBuilder. I wasn't seeing an appreciable performance difference with the way it was being used, and it may have been harming performance on typical requests where there is not duplication of the generated links. I do think we should be able to add a more long lived caching system that could improve performance.

There could also be a follow-on PR that adds an `except` flag to `jsonapi_relationships` along with a new `jsonapi_relationship` method to make it easier to create non-routable relationships.

### All Submissions:

### New Feature Submissions:

### Test Plan:

Testing this on some real world apps should be done. It should be a drop in replacement. Assuming all resources are routable there should be no changes needed and no warnings emitted.

If there are some relationships or resources that do not have routes these will emit warnings when they are serialized (say from includes). The `build_default_links` option should be used:

```ruby
class EmployeeResource < JSONAPI::Resource
  build_default_links false

  has_many :expense_entries, build_default_links: false
end
```

Singleton resources now need to be declared as singletons with the `singleton` macro. This macro allows the singleton status (true or false) and options to be set. The only current option is a callable for the `singleton_key`.

```ruby
class PreferencesResource < JSONAPI::Resource
 ...

  singleton singleton_key: -> (context) {
    key = context[:current_user].try(:preferences).try(:id)
    raise JSONAPI::Exceptions::RecordNotFound.new(nil) if key.nil?
    key
  }

end
```
It's also possible to override the `singleton_key` method on the resource instead of providing a callable.

Since this PR is on master and most production apps are still running v0.9 you can test with #1247 as well. The core feature should be the same.

### Reviewer Checklist:

### DIFF
--- a/lib/jsonapi/basic_resource.rb
+++ b/lib/jsonapi/basic_resource.rb
@@ -961,8 +961,15 @@ module JSONAPI
       end
 
       def exclude_links(exclude)
-        @_exclude_links = exclude.collect do |link|
-          link.to_sym
+        case exclude
+          when :default, "default"
+            @_exclude_links = [:self]
+          when :none, "none"
+            @_exclude_links = []
+          when Array
+            @_exclude_links = exclude.collect {|link| link.to_sym}
+          else
+            fail "Invalid exclude_links"
         end
       end
 
--- a/lib/jsonapi/basic_resource.rb
+++ b/lib/jsonapi/basic_resource.rb
@@ -423,7 +423,7 @@ module JSONAPI
         subclass.immutable(false)
         subclass.caching(_caching)
         subclass.singleton(singleton?, (_singleton_options.dup || {}))
-        subclass.build_default_links(_build_default_links)
+        subclass.exclude_links(_exclude_links)
         subclass.paginator(_paginator)
         subclass._attributes = (_attributes || {}).dup
         subclass.polymorphic(false)
@@ -960,16 +960,18 @@ module JSONAPI
         !@immutable
       end
 
-      def build_default_links(build)
-        @build_default_links = build
+      def exclude_links(exclude)
+        @_exclude_links = exclude.collect do |link|
+          link.to_sym
+        end
       end
 
-      def _build_default_links
-        @_build_default_links
+      def _exclude_links
+        @_exclude_links ||= []
       end
 
-      def build_default_links?
-        @_build_default_links.nil? || @_build_default_links == true
+      def exclude_link?(link)
+        _exclude_links.include?(link.to_sym)
       end
 
       def caching(val = true)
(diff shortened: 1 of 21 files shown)


lib/jsonapi/basic_resource.rb
```diff
@@ -629,6 +630,19 @@ def model_hint(model: _model_name, resource: _type)
         _model_hints[model.to_s.gsub('::', '/').underscore] = resource_type.to_s
       end
 
+      def singleton(*attrs)
+        @_singleton = (!!attrs[0] == attrs[0]) ? attrs[0] : true
```
dgeb, line 634: Is this effectively the same as `@_singleton = !!attrs[0]` ?

dgeb: Ah, perhaps you want it to be `true` if `attrs[0]` is `nil` - so just a bare `singleton` declaration

lgebhardt: I should ensure all combinations of this are tested.

dgeb (approved):
@lgebhardt looks good! I like the `:default` and `:none` options for `exclude_links`.

lib/jsonapi/relationship.rb
```diff
@@ -114,7 +137,7 @@ def initialize(name, options = {})
 
       def to_s
         # :nocov: useful for debugging
-        "#{parent_resource._type}.#{name} => (#{belongs_to? ? 'ToOne' : 'BelongsToOne'}) #{resource_klass._type}"
+        "#{parent_resource}.#{name}(#{belongs_to? ? 'ToOne' : 'BelongsToOne'})"
```
dgeb:
```suggestion
        "#{parent_resource}.#{name}(#{belongs_to? ? 'BelongsToOne' : 'ToOne'})"
```

lib/jsonapi/resource_serializer.rb
```diff
@@ -100,17 +100,12 @@ def serialize_to_links_hash(source, requested_relationship, resource_ids)
...
 
-      {
-          'links' => {
-              'self' => self_link(source, requested_relationship),
-              'related' => related_link(source, requested_relationship)
-          },
-          'data' => data
-      }
-    end
+      links_hash = { 'data': data }
```
dgeb: should `links_hash` be `relationship_hash`?

lgebhardt: Good catch. I thought I had fixed this naming issue a while ago.

PR #1262, opened 2019-06-20
Build Links with strings, track route setup on resource and relationships

### LINKED ISSUE #1259: Link generation performance

- [X] Problem/bug report.
- [ ] Feature request.

- [ ] I've searched for an existing issue.
- [ ] I've asked my question on [Gitter](https://gitter.im/cerebris/jsonapi-resources) and have not received a satisfactory answer.
- [ ] The feature I'm asking for is compliant with the [JSON:API](http://jsonapi.org/) spec.

## Description

#1247 introduced route based links. This uses the Rails `url_helpers` to generate the links. It turns out there are two issues with this. The first is the `Rails.application.routes.url_helpers` no longer memoizes the generated module, which results in a lot of extra modules being created, and possibly not garbage collected. This is corrected in #1257 by using the controller as it has the `url_helper` module included.

However there's still a performance issue. In testing using the `url_helpers` is significantly slower than the old way of generating the links (string concatenation). But the `url_helpers` generated links are generated from the actual routes, which allows us to ensure the route for the resource or relationship actually exists and issue warnings if that's not the case.

What we need is the speed of the earlier string based LinkBuilder with the ability to detect missing routes of the `url_helpers` based LinkBuilder.

I'd like to discuss options in this issue before I start coding up the fix.

lgebhardt:
Revert building links with 'url_helpers' for performance reasons

Now tracking whether routes have been setup for resources and relationships in the respective objects. This allows for the efficiency of string based link generation as well as warning when a route has not been setup. This should restore the performance we lost with the `url_helpers`.

Also reworked way the LinkBuilder handles Engines.

Closes #1259 

### All Submissions:

### New Feature Submissions:

### Reviewer Checklist:

### DIFF
--- a/lib/jsonapi/acts_as_resource_controller.rb
+++ b/lib/jsonapi/acts_as_resource_controller.rb
@@ -125,7 +125,8 @@ module JSONAPI
         base_url: base_url,
         key_formatter: key_formatter,
         route_formatter: route_formatter,
-        serialization_options: serialization_options
+        serialization_options: serialization_options,
+        controller: self
       )
       @resource_serializer
     end
--- a/lib/jsonapi/link_builder.rb
+++ b/lib/jsonapi/link_builder.rb
@@ -2,23 +2,24 @@ module JSONAPI
   class LinkBuilder
     attr_reader :base_url,
                 :primary_resource_klass,
+                :route_formatter,
                 :engine,
-                :routes
+                :engine_mount_point,
+                :url_helpers
+
+    @@url_helper_methods = {}
 
     def initialize(config = {})
-      @base_url               = config[:base_url]
+      @base_url = config[:base_url]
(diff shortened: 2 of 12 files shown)


lib/jsonapi/configuration.rb
```diff
@@ -52,8 +51,6 @@ def initialize
 
       self.raise_if_parameters_not_allowed = true
 
-      self.warn_on_missing_routes = true
```
krasnoukhov: Can we keep this option? It was useful to turn off if one knows what they are doing, so there are no annoying warnings

lgebhardt: That makes sense

lgebhardt: I've restored the setting.

krasnoukhov (approved):

PR #1296, opened 2019-11-14
Remove unnecessary setting of `missed_resource_ids`

scottgonzalez:
Since we're overwriting on the next line, there's no need to initialize to an empty hash.


lgebhardt (approved):
@scottgonzalez Thanks! Looks good to me.

PR #1311, opened 2020-02-24
Simplify Include Directives Hash

tommy-russoniello:
This simplifies include directives hashes by removing the `include_related` subhash under each entity. This indirection exists because of an `include` key that used to be in each hash, but [was removed](https://github.com/cerebris/jsonapi-resources/commit/5d1884985d967d5935d2d1082c4d71889a075f30). Since there are never any other keys in the hashes, the `include_related` subhashes are no longer necessary.

In doing this, I also found (and fixed) [a small bug](https://github.com/cerebris/jsonapi-resources/blob/978f590f85fe9e65d5806f206b661fb35332c1bd/lib/jsonapi/processor.rb#L446) where the processor tries to use an include directives hash as a key for its own subhash instead of the symbol `:include_related`.

---

### All Submissions:

### New Feature Submissions:

### Reviewer Checklist:

### DIFF
--- a/test/controllers/controller_test.rb
+++ b/test/controllers/controller_test.rb
@@ -4329,7 +4329,8 @@ class Api::BoxesControllerTest < ActionController::TestCase
                             "links" => {
                                 "self" => "http://test.host/api/things/40/relationships/things",
                                 "related" => "http://test.host/api/things/40/things"
-                            }
+                            },
+                            "data" => []
                         }
                     }
                 },
--- a/lib/jsonapi/active_relation_resource.rb
+++ b/lib/jsonapi/active_relation_resource.rb
@@ -92,7 +92,7 @@ module JSONAPI
       def find_fragments(filters, options = {})
         include_directives = options[:include_directives] ? options[:include_directives].include_directives : {}
         resource_klass = self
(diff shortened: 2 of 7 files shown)


test/controllers/controller_test.rb
```diff
@@ -4329,7 +4329,8 @@ def test_complex_includes_things_nested_things
                             "links" => {
                                 "self" => "http://test.host/api/things/40/relationships/things",
                                 "related" => "http://test.host/api/things/40/things"
-                            }
+                            },
+                            "data" => []
```
scottgonzalez, line 4333: @lgebhardt I would like you to confirm something about this change.

@tommy-russoniello asked me why this would have changed but L4389 has another resource that did not change to include `data`. My explanation for this is that this resource is not a leaf in the `things.things.things` include, but the other one is, so we did the work of looking up the related things for this one, but not the other.

scottgonzalez (approved):

test/controllers/controller_test.rb
lgebhardt: @scottgonzalez I believe you are correct.

lgebhardt (approved):
@tommy-russoniello Thanks for cleaning this up.

PR #1387, opened 2022-04-20
Co-erce non-present values like false to string

bf4:
We have a resource which calls another resource, and specifically calls 'verify_filters'. We discovered that in 0.9 and 0.10 `verify_filter` calls `raw.present?` which is false when raw is `false`.  This change makes the processing of the true and false consistent.

### All Submissions:

### Reviewer Checklist:

### DIFF
--- a/lib/jsonapi/resource.rb
+++ b/lib/jsonapi/resource.rb
@@ -895,6 +895,13 @@ module JSONAPI
 
       def verify_filter(filter, raw, context = nil)
         filter_values = []
+        if raw == true
+          raw = "true"
+        elsif raw == false
+          raw = "false"
+        else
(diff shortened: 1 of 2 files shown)


lib/jsonapi/resource.rb
```diff
@@ -895,6 +895,13 @@ def is_filter_relationship?(filter)
 
       def verify_filter(filter, raw, context = nil)
         filter_values = []
+        if raw == true
+          raw = "true"
+        elsif raw == false
+          raw = "false"
+        else
+          nil # no-op
+        end
```
bf4, line 904: @lgebhardt any objection to this? otherwise false is treated as not present

PR #1370, opened 2021-05-29
Add MySQL to CI workflow

karlwilbur:
This PR adds MySQL to the CI workflow, as mentioned in #1369.

There is currently a MySQL syntax error in the `0.10.x` code that would be detected in CI with MySQL in the testing matrix.

### All Submissions:

- [x] I've checked to ensure there aren't other open [Pull Requests](https://github.com/cerebris/jsonapi-resources/pulls) for the same update/change.
- [x] I've submitted a [ticket](https://github.com/cerebris/jsonapi-resources/issues) for my issue if one did not already exist.
- [ ] My submission passes all tests. (Please run the full test suite locally to cut down on noise from travis failures.)
- [ ] I've used Github [auto-closing keywords](https://help.github.com/articles/closing-issues-via-commit-messages/) in the commit message or the description.
- [ ] I've added/updated tests for this change.

### New Feature Submissions:

- [ ] I've submitted an issue that describes this feature, and received the go ahead from the maintainers.
- [ ] My submission includes new tests.
- [ ] My submission maintains compliance with [JSON:API](http://jsonapi.org/).

### Bug fixes and Changes to Core Features:

- [x] I've included an explanation of what the changes do and why I'd like you to include them.
- [ ] I've provided test(s) that fails without the change.

### Reviewer Checklist:
- [x] Maintains compliance with JSON:API
- [ ] Adequate test coverage exists to prevent regressions

PR #1398, opened 2023-01-25
deps: Add MySQL to CI workflow

bf4:
- also add dependabot workflow
- also standardize rails/ruby version quoting/ordering

Based on #1370

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -2,7 +2,7 @@ name: CI
 
 on:
   push:
-    branches: [ 'master', 'release-0-8', 'release-0-9', 'release-0-10' ]
+    branches: [ 'master', 'release-0-8', 'release-0-9', 'release-0-10', 'v0-11-dev' ]
   pull_request:
     branches: ['**']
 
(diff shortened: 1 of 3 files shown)


.github/workflows/ruby.yml
```diff
@@ -26,43 +38,44 @@ jobs:
       fail-fast: false
       matrix:
         ruby:
-          - 2.6
-          - 2.7
+          - '3.2'
```
bf4, line 41: YAML can be weird about strings vs. number, so just being explicit in all cases

also, always descending version number in all cases

.github/workflows/ruby.yml
```diff
@@ -26,43 +38,44 @@ jobs:
...
-          - 7.0.4
-          - 6.1.7
-          - 6.0.6
-          - 5.2.8.1
-          - 5.1.7
+          - '7.0'
+          - '6.1'
+          - '6.0'
+          - '5.2'
+          - '5.1'
```
bf4, line 51: removed the minor/patch version numbers since I'm not sure they matter for our purposes

Issue #1369, opened 2021-05-29
MySQL Syntax Error in ActiveRelationResource (0.10.x)


karlwilbur:
## This issue is a (choose one):

- [x] Problem/bug report.
- [ ] Feature request.
- [ ] Request for support. **Note: Please try to avoid submitting issues for support requests. Use [Gitter](https://gitter.im/cerebris/jsonapi-resources) instead.** 

## Checklist before submitting:

- [x] I've searched for an existing issue.
- [ ] I've asked my question on [Gitter](https://gitter.im/cerebris/jsonapi-resources) and have not received a satisfactory answer.
- [ ] I've included a complete [bug report template](https://github.com/cerebris/jsonapi-resources/blob/master/lib/bug_report_templates/rails_5_master.rb). This step helps us and allows us to see the bug without trying to reproduce the problem from your description. It helps you because you will frequently detect if it's a problem specific to your project.
- [ ] The feature I'm asking for is compliant with the [JSON:API](http://jsonapi.org/) spec.

## Description

Choose one section below and delete the other:

### Bug reports:

Version `0.10.5`

```
#<ActiveRecord::StatementInvalid: Mysql2::Error: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '."id" AS "individuals_id", individuals.id FROM `individuals` WHERE `individuals`' at line 1: SELECT  "individuals"."id" AS "individuals_id", individuals.id FROM `individuals` WHERE `individuals`.`deleted_at` IS NULL ORDER BY individuals.id asc LIMIT 10 OFFSET 0>
>> fragments = resource_klass.find_fragments(find_options[:filters], options)
```

This is related to #1310, but is not a multi-tenancy issue. This a MySQL syntax issue that I believe it created by the [`ActiveRelationResource#quote`](https://github.com/cerebris/jsonapi-resources/blob/v0.10.5/lib/jsonapi/active_relation_resource.rb#L833-L835) method which is uses explicit double quotes for quoting table names in [`ActiveRelationResource#concat_table_field`](https://github.com/cerebris/jsonapi-resources/blob/v0.10.5/lib/jsonapi/active_relation_resource.rb#L789-L807) and [`ActiveRelationResource#alias_table_field`](https://github.com/cerebris/jsonapi-resources/blob/v0.10.5/lib/jsonapi/active_relation_resource.rb#L813-L831).

[...]

PR #1400, opened 2023-01-31
fix: Reliably quote columns/tables

bf4:
see https://github.com/rails/rails/blob/7-0-stable/activerecord/lib/active_record/connection_adapters/abstract/quoting.rb#L8-L146

Addresses https://github.com/cerebris/jsonapi-resources/issues/1369

### DIFF
--- a/lib/jsonapi/active_relation_resource.rb
+++ b/lib/jsonapi/active_relation_resource.rb
@@ -860,6 +860,7 @@ module JSONAPI
       end
 
       def quote_column_name(column_name)
+        return column_name if column_name == "*"
         if _model_class.try(:connection)
           _model_class.connection.quote_column_name(column_name)
         else
@@ -868,7 +869,6 @@ module JSONAPI
       end
 
       def quote(field)
-        raise "zomg, we are using the fallback for #{field}"
         %{"#{field.to_s}"}
       end
 
--- a/lib/jsonapi/active_relation_resource.rb
+++ b/lib/jsonapi/active_relation_resource.rb
@@ -804,14 +804,14 @@ module JSONAPI
         if table.blank? || field.to_s.include?('.')
           # :nocov:
           if quoted
-            quote(field)
+            quote(field) # split on '.'?
           else
             field.to_s
           end
           # :nocov:
         else
           if quoted
-            "#{quote(table)}.#{quote(field)}"
+            "#{quote_table_name(table)}.#{quote_column_name(field)}"
           else
             # :nocov:
             "#{table.to_s}.#{field.to_s}"
@@ -828,7 +828,7 @@ module JSONAPI
         if table.blank? || field.to_s.include?('.')
           # :nocov:
           if quoted
-            quote(field)
+            quote_column_name(field)
           else
             field.to_s
           end
@@ -836,12 +836,20 @@ module JSONAPI
         else
           if quoted
             # :nocov:
-            quote("#{table.to_s}_#{field.to_s}")
+            quote_column_name("#{table.to_s}_#{field.to_s}")
             # :nocov:
           else
             "#{table.to_s}_#{field.to_s}"
           end
         end
+      end
+
+      def quote_table_name(table_name)
+        quote(table_name)
+      end
+
+      def quote_column_name(column_name)
+        quote(column_name)
       end
 
       def quote(field)
--- a/lib/jsonapi/active_relation_resource.rb
+++ b/lib/jsonapi/active_relation_resource.rb
@@ -852,7 +852,7 @@ module JSONAPI
       end
 
       def quote_table_name(table_name)
-        if _model_class.try(:connection)
+        if _model_class&.connection
           _model_class.connection.quote_table_name(table_name)
         else
           quote(table_name)
@@ -861,13 +861,14 @@ module JSONAPI
 
       def quote_column_name(column_name)
         return column_name if column_name == "*"
-        if _model_class.try(:connection)
+        if _model_class&.connection
           _model_class.connection.quote_column_name(column_name)
         else
           quote(column_name)
         end
       end
 
+      # fallback quote identifier when database adapter not available
       def quote(field)
         %{"#{field.to_s}"}
       end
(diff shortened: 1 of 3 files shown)


lib/jsonapi/active_relation_resource.rb
```diff
@@ -820,30 +831,42 @@ def concat_table_field(table, field, quoted = false)
...
+      def quote_column_name(column_name)
+        if _model_class.try(:connection)
+          _model_class.connection.quote(column_name)
+        else
+          quote(column_name)
+        end
+      end
+
       def quote(field)
         "\"#{field.to_s}\""
```
bf4: errors on 
```
DATABASE_URL='postgres://postgres@localhost/jr_test' ./scripts/cibuild TESTOPTS="--name=test_show_author_recursive -v"
```

<table>
<tr>
<td>

<pre>
>> non_caching_response
=>
{"errors"=>
  [{"title"=>"Internal Server Error",
    "detail"=>"Internal Server Error",
    "code"=>"500",
    "status"=>"500",
    "meta"=>
     {"exception"=>
       "PG::SyntaxError: ERROR:  syntax error at or near \".\"\nLINE 1: SELECT 'people'.'id' AS 'people_id', \"people\".\"id\" FROM \"peo...\n                       ^\n"}}]}
</pre>
</td>
<td>
<pre>
>> json_response_sans_all_backtraces
=>
{"errors"=>
  [{"title"=>"Internal Server Error",
    "detail"=>"Internal Server Error",
    "code"=>"500",
    "status"=>"500",
    "meta"=>
     {"exception"=>
       "PG::SyntaxError: ERROR:  syntax error at or near \".\"\nLINE 1: SELECT 'people'.'id' AS 'people_id', 'people'.'updated_at' A...\n                       ^\n"}}]}
</pre>
</td>
</tr>
</table>

bf4: oh, typo in there

```diff
-          _model_class.connection.quote(column_name)
+          _model_class.connection.quote_table_name(column_name)
```

`quote` treats the input as a literal and so it gets single quotes `'`
`quote_table_name`/`quote_column_name` treats the input as an identifier and so it gets double quotes `"`

lib/jsonapi/active_relation_resource.rb
```diff
@@ -820,32 +831,44 @@ def concat_table_field(table, field, quoted = false)
...
+      def quote_column_name(column_name)
+        if _model_class.try(:connection)
+          _model_class.connection.quote_column_name(column_name)
+        else
+          quote(column_name)
+        end
+      end
+
       def quote(field)
-        "\"#{field.to_s}\""
```
bf4, line 848: in pg, is basically `quote_ident`.  for values, should be single-quotes like `quote_literal`

bf4:
Not sure what the current failure means.  Work so far is an improvement, I think.. not sure what to do next

lib/jsonapi/active_relation_resource.rb
```diff
@@ -820,32 +831,45 @@ def concat_table_field(table, field, quoted = false)
...
+            quote_column_name("#{table.to_s}_#{field.to_s}")
             # :nocov:
           else
             "#{table.to_s}_#{field.to_s}"
           end
         end
       end
 
+      def quote_table_name(table_name)
+        if _model_class.try(:connection)
```
bf4: could be `_model_class&.connection`

lgebhardt: Yes, this would be better now that we don't support ruby versions older than 2.3

bf4: I was thinking 'what if _model_class exists but doesn't respond to connection' but then I was thinking 'this is the ar resource, it's fine', but then I was thinking, ... eh

test/controllers/controller_test.rb
```diff
@@ -494,15 +494,15 @@ def test_sorting_by_relationship_field
 
     assert_response :success
     assert json_response['data'].length > 10, 'there are enough records to show sort'
+    expected = Post
```
bf4: @lgebhardt rather than assert different rules here based on the value of `adapter_sorts_nulls_last` and try to reproduce the db behavior, I thought it would be sufficient to assert that the query results on the AR model give the same result as the resource.  

Thoughts? I briefly tried using defining `adapter_sorts_nulls_last` then writing a transform in Ruby but I didn't duplicate the pg/mysql/sqlite logic and then thought, "why am I even?"

```ruby
  def adapter_specific_sort_comparison(l,r, sort_type:)
    if l && r
      comparison = l <=> r
      case sort_type
      when :sort then comparison
      when :minus_sort then -comparison
      else fail ArgumentError, "Unhandled sort_type #{sort_type} in #{__callee__}"
      end
    elsif l.nil? && r.nil?
      0
    elsif l.nil?
      adapter_sorts_nulls_last ? -1 : 1
    else
      adapter_sorts_nulls_last ? 1 : -1
    end
  end
```

bf4: technically this is out of scope of quoting and is part of "make mysql" tests pass.

test/controllers/controller_test.rb
```diff
@@ -494,15 +494,15 @@ def test_sorting_by_relationship_field
...
+      .all
+      .left_joins(:author)
+      .merge(Person.order(name: :asc))
+      .map(&:id)
+      .map(&:to_s)
+    ids = json_response['data'].map {|data| data['id'] }
 
-    # Postgres sorts nulls last, whereas sqlite and mysql sort nulls first
-    if ENV['DATABASE_URL'].starts_with?('postgres')
```
bf4: I replaced all "what adapter is this" cases I found with "what behavior do I expect"

bf4: I extracted this into https://github.com/cerebris/jsonapi-resources/pull/1402

lib/jsonapi/active_relation_resource.rb
bf4: updated to `&.`

test/controllers/controller_test.rb
bf4: I extracted this into https://github.com/cerebris/jsonapi-resources/pull/1402

bf4:
@lgebhardt I decided to merge this since the target is a dev branch and this meets the goals I set re: quoting and now it can be the basis for further work

lgebhardt:
@bf4 sounds good to me. I'm not sure how to best handle the adapter based sorting differences, but we can definitely handle that in a separate pr.

PR #1411, opened 2023-09-19
V0_11 dev

lgebhardt:
**NOTE: This is a draft PR and it is expected that things will change. Currently this PR has been released as v0.11.0.beta2 to ease testing. I do not recommend using the beta release in a production environment.**

In order to avoid maintaining the older branches and versions v0.11 will support retrieving resources in similar ways to v0.09 and v0.10 in addition to the changes made to the new default.

This brings together support for retrieving resources using the methods of the past versions. This can be controlled by the new configuration option `default_resource_retrieval_strategy`.  The following options are supported:
- 'JSONAPI::ActiveRelationRetrieval' (default)
- 'JSONAPI::ActiveRelationRetrievalV09'
- 'JSONAPI::ActiveRelationRetrievalV10'
- :none
- :self

`JSONAPI::ActiveRelationRetrievalV09`:

`JSONAPI::ActiveRelationRetrievalV10`: Similar to how v0.10.x retrieves related resources by joining the related resource's model to the primary resource's model and merging in the related resource's `records`. Differences from the earlier version is the additional pluck queries have been eliminated in the case where caching is not used.

`JSONAPI::ActiveRelationRetrieval`: This retrieves related resources directly and joins to the primary resource through the inverse relationship

`:none` and `:self`: The resource class, or a base class, is expected to implement or include the methods for retrieving resources.

### Additional Features to implement

- [ ] Deprecate direct override of `records_for_populate`. Instead define `records_for_populate` to call new methods for:
  - [ ] New resource setting `serialization_includes` for includes needed for serialization. 
  - [ ] New overridable method `apply_serialization` (name needs rework)
  - This will allow serialization options such as includes and added sql clauses for computed attributes to only be added for building relations that will be used for serialization. 
 - [ ] Create demo/test applications that demonstrate authorization for each of the retrieval strategies
 - [ ] Investigate whether retrieval strategies can be mixed
 - [ ] Documentation
 - [ ] Documentation for PORO resources


lib/jsonapi/configuration.rb
```diff
@@ -160,6 +162,26 @@ def initialize
...
+      # To use multiple retrieval strategies in an app set this to :none and set a custom retrieval strategy
+      # per resource (or base resource) using the class method `load_resource_retrieval_strategy`.
+      #
+      # Available strategies:
+      # 'JSONAPI::ActiveRelationRetrieval'
+      # 'JSONAPI::ActiveRelationRetrievalV09'
+      # 'JSONAPI::ActiveRelationRetrievalV10'
+      # :none
+      # :self
+      self.default_resource_retrieval_strategy = 'JSONAPI::ActiveRelationRetrieval'
```
bf4, line 189: 👀

lib/jsonapi/configuration.rb
```diff
@@ -160,6 +162,26 @@ def initialize
...
+
+      # For 'JSONAPI::ActiveRelationRetrievalV10': use a related resource's `records` when performing joins.
+      # This setting allows included resources to account for permission scopes. It can be overridden explicitly per
+      # relationship. Furthermore, specifying a `relation_name` on a relationship will cause this setting to be ignored.
+      self.use_related_resource_records_for_joins = true
```
bf4, line 194: right now only changes behavior in ActiveRelationRetrievalV10

lib/jsonapi/active_relation_retrieval.rb
```diff
@@ -0,0 +1,885 @@
...
+          end
+          fragments
+        else
+          relationship.resource_klass.find_related_fragments_from_inverse(source_fragments, relationship, options, true)
+        end
+      end
+
+      def find_related_fragments_from_inverse(source, source_relationship, options, connect_source_identity)
+        relationship = source_relationship.resource_klass._relationship(source_relationship.inverse_relationship)
+        raise "missing inverse relationship" unless relationship.present?
```
krasnoukhov: It would be nice to be more specific in this exception, would be easier for people to identify problematic relationship

lib/jsonapi/active_relation_retrieval_v09.rb
```diff
@@ -0,0 +1,715 @@
...
+                                          primary: primary)
+
+            fragments[related_resource.identity] = fragment
+            complete_linkages(fragment, linkage_relationships)
+          end
+
+          if source_resource
+            source_resource.add_related_identity(source_relationship.name, related_resource.identity)
+            fragment.add_related_from(source_resource.identity)
+            fragment.add_related_identity(source_relationship.inverse_relationship, source_resource.identity)
```
krasnoukhov: It's possible that there's no inverse relationship, in this case this fails. Would it be ok to wrap this in a conditional to not do this if there's no inverse?

krasnoukhov: I tried adding this conditional while integrating this in one of our apps and it worked just fine, everything passes

lib/jsonapi/active_relation_retrieval.rb
lgebhardt: Good idea

lib/jsonapi/active_relation_retrieval_v09.rb
lgebhardt: I think that's a good idea. In order to provide clarity for the behavior we probably should log why it's being skipped (in dev mode) and allow a flag is set on the source_relationship to mute this message. I'll take a look at it.

jsonapi-resources.gemspec
```diff
@@ -27,7 +27,9 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'pry'
   spec.add_development_dependency 'concurrent-ruby-ext'
   spec.add_development_dependency 'database_cleaner'
+  spec.add_development_dependency 'hashie'
   spec.add_dependency 'activerecord', '>= 5.1'
   spec.add_dependency 'railties', '>= 5.1'
   spec.add_dependency 'concurrent-ruby'
+  spec.add_dependency 'sorted_set'
```
bf4: I'm getting a compile error in CRuby 3.1.2 against the rbtree gem. Looking into it.

bf4: I made an issue in the related rbtree gem https://github.com/mame/rbtree/issues/5 

Looking at the removal of sorted_set from the standard lib and reasons (performance) https://github.com/ruby/set/pull/2/files and a gem which was changed to work without it ( https://github.com/bokmann/business_time/pull/214/files ) I wonder if it's worth including here?

lgebhardt: I originally added SortedSet to solve a change in the results order, but upon testing this again it appears other changes have resolved the issue. I'll approve your PR in to remove it. However I didn't see a significant change in speed with or without SortedSet, so I'm not sure it's slow enough to matter.

lgebhardt: Well I merged #1423 and see my local testing wasn't correctly using Postgres and the issue is still there with Postgres in the test suite. I may reverse this, but changing the tests is another possibility since none of the failures are real failures since includes do not have a required order.

lib/jsonapi/active_relation_retrieval_v09.rb
krasnoukhov: Hey @lgebhardt, were you able to look at this? This is the only thing that is preventing us from adopting further betas

lgebhardt: @krasnoukhov I haven't tackled this yet, but I will get to it very soon. I have made some progress on the rake task to detect missing inverse relationships, and should be able to get PRs out for both of these soon.

krasnoukhov: Hey @lgebhardt, any progress here?

lgebhardt: @krasnoukhov Sorry to long wait. I had a project that took much more time than expected and it pulled me away from JR for the last 3 weeks. I'm getting back into it now and hope to have a new beta out soon.

PR #1427, opened 2024-01-11
Make SortedSet for identity arrays optional

lgebhardt:
Restores SortedSet as an option.

Removes default sorting for includes. Because of how included resources are collected the database sort doesn't actually produce properly sorted results, but does slow the queries.

Fixes `connect_source_identity` option when  calling `find_related_fragments_from_inverse` for polymorphic relationship routes

### All Submissions:

### New Feature Submissions:

### Reviewer Checklist:

### DIFF
--- a/jsonapi-resources.gemspec
+++ b/jsonapi-resources.gemspec
@@ -28,6 +28,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'concurrent-ruby-ext'
   spec.add_development_dependency 'database_cleaner'
   spec.add_development_dependency 'hashie'
+  spec.add_development_dependency 'sorted_set'
   spec.add_dependency 'activerecord', '>= 5.1'
   spec.add_dependency 'railties', '>= 5.1'
   spec.add_dependency 'concurrent-ruby'
--- a/jsonapi-resources.gemspec
+++ b/jsonapi-resources.gemspec
@@ -31,4 +31,5 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'activerecord', '>= 5.1'
   spec.add_dependency 'railties', '>= 5.1'
   spec.add_dependency 'concurrent-ruby'
+  spec.add_dependency 'sorted_set'
 end
--- a/jsonapi-resources.gemspec
+++ b/jsonapi-resources.gemspec
@@ -32,5 +32,4 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'activerecord', '>= 5.1'
   spec.add_dependency 'railties', '>= 5.1'
   spec.add_dependency 'concurrent-ruby'
-  spec.add_dependency 'sorted_set'
 end
--- a/lib/jsonapi/active_relation_retrieval.rb
+++ b/lib/jsonapi/active_relation_retrieval.rb
@@ -271,7 +271,7 @@ module JSONAPI
           source_resource_klasses.each do |resource_klass|
             inverse_direct_relationship = _relationship(resource_klass._type.to_s.singularize)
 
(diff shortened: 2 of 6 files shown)


jsonapi-resources.gemspec
```diff
@@ -31,4 +31,5 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'activerecord', '>= 5.1'
   spec.add_dependency 'railties', '>= 5.1'
   spec.add_dependency 'concurrent-ruby'
+  spec.add_dependency 'sorted_set'
```
bf4:
```suggestion
  spec.add_development_dependency 'sorted_set'
```

right?

lib/jsonapi/resource_fragment.rb
```diff
@@ -25,7 +25,7 @@ def initialize(identity, resource: nil, cache: nil, primary: false)
       @primary = primary
 
       @related = {}
-      @related_from = Set.new
+      @related_from = JSONAPI.configuration.sort_related_identities_by_primary_key ? SortedSet.new : Set.new
```
bf4: maybe have a constant like

```ruby
RELATION_SET = begin
  require "sorted_set"
rescue LoadError
   SortedSet
else
  Set
end
```

then you can have

```ruby
      @related_from = JSONAPI.configuration.sort_related_identities_by_primary_key ? RELATION_SET.new : Set.new
```

?  Or just have `RELATION_SET` and remove the config?

same in lib/jsonapi/resource_set.rb

lib/jsonapi/active_relation_retrieval.rb
```diff
@@ -271,7 +271,7 @@ def find_related_fragments(source_fragment, relationship, options = {})
           source_resource_klasses.each do |resource_klass|
             inverse_direct_relationship = _relationship(resource_klass._type.to_s.singularize)
 
-            fragments.merge!(resource_klass.find_related_fragments_from_inverse([source_fragment], inverse_direct_relationship, options, true))
+            fragments.merge!(resource_klass.find_related_fragments_from_inverse([source_fragment], inverse_direct_relationship, options, false))
```
bf4, line 274: should this true, false depend on the setting? it seems that this could also  be something which boils down to `JSONAPI.configuration.related_identities_set != Set` i.e. if the setting isn't the default (Set) then `connect_source_identity` is true.

Since below we have `unless connect_source_identity` (which I usually would write as `sort_by_includes = !connect_source_identity; if sort_by_includes` or something like that to avoid the unless)

lib/jsonapi/configuration.rb
```diff
@@ -182,6 +183,12 @@ def initialize
       # This setting allows included resources to account for permission scopes. It can be overridden explicitly per
       # relationship. Furthermore, specifying a `relation_name` on a relationship will cause this setting to be ignored.
       self.use_related_resource_records_for_joins = true
+
+      # Collect the include keys into a Set or a SortedSet. SortedSet carries a small performance cost in the rails app
+      # but produces consistent and more human navigable result sets.
+      # Note: If using SortedSet be sure to add `sorted_set` to your Gemfile and require 'sorted_set'`
+      # require 'sorted_set'
```
bf4, line 190:
```suggestion
      # require 'sorted_set'
      # config.related_identities_set = SortedSet
```

lib/jsonapi/active_relation_retrieval.rb
lgebhardt: `find_related_fragments_from_inverse` is called for both getting the primary resources from a relationship route and from includes. The `connect_source_identity` flag distinguishes the two, and the above change fixed an error I introduced at some point when handling the polymorphic case. In a separate PR I may look at clarifying the code since this distinction does seem to get lost.

I like your coding convention to eliminate the `unless` and have introduced a commit that should clarify things.

PR #1428, opened 2024-01-12
Store the resource_klass and id in an array for efficiency

lgebhardt:
Removes the need to allocate a new array for every comparison.

Found a large number of allocations in the `hash` method when profiling the code. This removes about 35K allocations from a result set that contained about 3500 resources (1000 primary with 2500 includes). Not surprisingly that doesn't actually make a statistically significant performance issue for small requests like the one tested, but every little bit helps.

Also changes the `eql?` to use the `hash` method 

### All Submissions:

### New Feature Submissions:

### Reviewer Checklist:

### DIFF
--- a/lib/jsonapi/resource_identity.rb
+++ b/lib/jsonapi/resource_identity.rb
@@ -13,11 +13,17 @@ module JSONAPI
   # rid = ResourceIdentity.new(PostResource, 12)
   #
   class ResourceIdentity
-    attr_reader :resource_klass, :id
-
+    # Store the identity parts as an array to avoid allocating a new array for the hash method to work on
     def initialize(resource_klass, id)
-      @resource_klass = resource_klass
-      @id = id
+      @identity_parts = [resource_klass, id]
+    end
+
+    def resource_klass
+      @identity_parts[0]
+    end
+
+    def id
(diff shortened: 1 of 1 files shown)


bf4 (approved):
Looks good

lib/jsonapi/resource_identity.rb
```diff
@@ -27,11 +33,11 @@ def ==(other)
...
-      other.is_a?(ResourceIdentity) && other.resource_klass == @resource_klass && other.id == @id
+      hash == other.hash
     end
 
     def hash
-      [@resource_klass, @id].hash
+      @identity_parts.hash
     end
 
     def <=>(other_identity)
```
bf4, line 43: Interesting that eql? and <=> diverge in what they compare  but is out of scope to this change

lgebhardt: For <=> it really should be comparing the resource_klasses first and then the ids second to determine the response. I think that would only affect the ordering of identities in the related identities when we're using a SortedSet and storing identities for a polymorphic relationship. It looks like that comparison can be done without any allocations so it's probably a good idea to make the change, but I'll handle that in a separate PR.

PR #1431, opened 2024-01-15
V0 11 dev performance

lgebhardt:
Implements some caching of frequently computed values. Significantly reduces allocations and run times for larger requests.

### All Submissions:

### New Feature Submissions:

### Reviewer Checklist:

### DIFF
--- a/lib/jsonapi/link_builder.rb
+++ b/lib/jsonapi/link_builder.rb
@@ -49,8 +49,8 @@ module JSONAPI
 
     def relationships_related_link(source, relationship, query_params = {})
       if relationship._routed
-        url = "#{ self_link(source) }/#{ route_for_relationship(relationship) }"
-        url = "#{ url }?#{ query_params.to_query }" if query_params.present?
+        url = +"#{ self_link(source) }/#{ route_for_relationship(relationship) }"
+        url << "?#{ query_params.to_query }" if query_params.present?
         url
       else
         if JSONAPI.configuration.warn_on_missing_routes && !relationship._warned_missing_route
@@ -129,16 +129,12 @@ module JSONAPI
       @_resources_path[source_klass] ||= formatted_module_path_from_class(source_klass) + format_route(source_klass._type.to_s)
     end
 
-    def resource_path(source)
-      if source.class.singleton?
-        resources_path(source.class)
-      else
-        "#{resources_path(source.class)}/#{source.id}"
-      end
-    end
-
     def resource_url(source)
-      "#{ base_url }#{ engine_mount_point }#{ resource_path(source) }"
+      if source.class.singleton?
+        "#{ base_url }#{ engine_mount_point }#{ resources_path(source.class) }"
+      else
+        "#{ base_url }#{ engine_mount_point }#{resources_path(source.class)}/#{source.id}"
+      end
     end
 
     def route_for_relationship(relationship)
--- a/lib/jsonapi/resource_common.rb
+++ b/lib/jsonapi/resource_common.rb
@@ -39,7 +39,7 @@ module JSONAPI
     end
 
     def id
-      _model.public_send(self.class._primary_key)
+      @id ||= _model.public_send(self.class._primary_key)
     end
 
     def identity
--- a/lib/jsonapi/resource_common.rb
+++ b/lib/jsonapi/resource_common.rb
@@ -1181,18 +1181,6 @@ module JSONAPI
         end
       end
 
-      def _setup_relationship(klass, *attrs)
-        _clear_fields_cache
-
-        options = attrs.extract_options!
-        options[:parent_resource] = self
-
-        relationship_name = attrs[0].to_sym
-        check_duplicate_relationship_name(relationship_name)
-
-        define_relationship_methods(relationship_name.to_sym, klass, options)
-      end
-
       # ResourceBuilder methods
       def define_relationship_methods(relationship_name, relationship_klass, options)
         relationship = register_relationship(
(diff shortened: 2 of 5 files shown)


lib/jsonapi/link_builder.rb
```diff
@@ -49,8 +49,8 @@ def query_link(query_params)
 
     def relationships_related_link(source, relationship, query_params = {})
       if relationship._routed
-        url = "#{ self_link(source) }/#{ route_for_relationship(relationship) }"
-        url = "#{ url }?#{ query_params.to_query }" if query_params.present?
+        url = +"#{ self_link(source) }/#{ route_for_relationship(relationship) }"
```
bf4, line 52: +1

lib/jsonapi/resource_common.rb
```diff
@@ -39,7 +39,7 @@ def _model
     end
 
     def id
-      _model.public_send(self.class._primary_key)
+      @id ||= _model.public_send(self.class._primary_key)
```
bf4, line 42: Off topic-- we'll want to handle composite primary keys here

lib/jsonapi/resource_common.rb
```diff
@@ -553,11 +560,16 @@ def _resource_name_from_type(type)
       end
 
       def resource_type_for(model)
-        model_name = model.class.to_s.underscore
-        if _model_hints[model_name]
-          _model_hints[model_name]
-        else
-          model_name.rpartition('/').last
+        @_cached_resource_types_for_model.fetch(model.class) do
+          model_name = model.class.name.underscore
```
bf4, line 564: Missing setting the value to the hash inside the fetch block

bf4: Nm

lib/jsonapi/resource_common.rb
```diff
@@ -1218,6 +1218,14 @@ def _clear_cached_attribute_options
         @_cached_attribute_options = {}
       end
 
+      def _clear_cached_resource_types_for_model
+        @_cached_resource_types_for_model = {}
```
bf4: No sure how allocations differ re garbage collection to set a new ivar vs calling `clear`

lib/jsonapi/resource_serializer.rb
```diff
@@ -262,7 +262,7 @@ def links_hash(source)
       if !links.key?('self') && !source.class.exclude_link?(:self)
         links['self'] = link_builder.self_link(source)
       end
-      links.compact
+      links.delete_if {|k,v| v.nil? }
```
bf4, line 265: Might as well use compact! Or compact_blank! no?

lib/jsonapi/resource_tree.rb
```diff
@@ -68,13 +68,13 @@ def add_resource(resource, include_related)
     private
 
     def init_included_relationships(fragment, include_related)
-      include_related && include_related.each_key do |relationship_name|
+      include_related&.each_key do |relationship_name|
```
bf4, line 71: +1

lib/jsonapi/resource_tree.rb
```diff
@@ -68,13 +68,13 @@ def add_resource(resource, include_related)
...
         fragment.initialize_related(relationship_name)
       end
     end
 
     def load_included(resource_klass, source_resource_tree, include_related, options)
-       include_related.try(:each_key) do |key|
+       include_related&.each_key do |key|
```
bf4, line 77: +1

lib/jsonapi/resource_common.rb
lgebhardt: That's an interesting point. I don't see anything in the spec to provide support for a multi attribute composite key in the JSONAPI section on [identification](https://jsonapi.org/format/#document-resource-object-identification). We could attempt to support this in JR by combining the composite key components, but I suspect it's going to run into the same issues I encountered trying to find a way to support an alternate public key field (one different from the primary key in the database). It was a harder problem than I initially thought it would be.

lib/jsonapi/resource_serializer.rb
lgebhardt: Thanks, for this case `compact_blank!` does look to be the method I initially tried to find. It has the benefit of working as a one liner as opposed to `compact!` which returns nil if no entries are removed. But I think I would like to keep the code as is since the method wasn't introduced until rails 6.1 ☹️

bf4: Oh, my mistake. For some reason I thought it had been added to standard Ruby but I just checked and it appears not

lib/jsonapi/resource_common.rb
bf4: I figure someone could also just override this method. Since keys need to be consistently ordered,  a custom synthetic key would be fine and keep the risk of a spec error off of jsonapi.

lgebhardt: I think that's the way I would handle it too. To work with rails we would need an identity class to convert the composite string keys to an array and back. It would an interesting thing to test.

lib/jsonapi/resource_common.rb
lgebhardt: I'm not sure either. I assume, probably a bad thing to do, that a Set or Has would be cleaned up the same as the values it holds when no longer referenced. But I took a look at the code and reworked how the class variables for the caches are initialized and cleared. Probably doesn't matter much since this is code that's only run at startup. Still I hope it makes it a little clearer.

bf4 (approved):
Looks good

PR #1450, opened 2024-02-16
Fix include_optional_linkage_data with joins

### LINKED ISSUE #1449: get_join_arel_node fails with include_optional_linkage_data if there is already a join

- [x] Problem/bug report.
- [ ] Feature request.

- [x] I've searched for an existing issue.
- [ ] I've asked my question on [Gitter](https://gitter.im/cerebris/jsonapi-resources) and have not received a satisfactory answer.
- [x] The feature I'm asking for is compliant with the [JSON:API](http://jsonapi.org/) spec.

## Description

If the user adds a join (e.g. in `records`) and that association has `include_optional_linkage_data: true`, then `get_join_arel_node` prints "get_join_arel_node: No join added" to stdout, returns nil, and a little later we crash because of the nil.

This happens because `get_join_arel_node` compares the node count before & after adding a join, but the count doesn't change because the join was already there.

We should fix this by looking for a compatible join and only failing if we can't find one. If we are trying to add a left outer join, then any join type will do (since adding an outer join wouldn't change the cardinality). If we are trying to add an inner join, then only another inner join is compatible (since otherwise the filtering we expect wouldn't happen).

I have a test case in my related PR, so I haven't included one here.

I hit this bug on `master`  (a3a2a7a3ce62df3c537d51e7f677a11cecca8ef9) and ruby 3.1.2p20.

pjungwir:
When the user adds a join in `records`, it confuses `get_join_arel_node` because it finds the same number of joins before & after adding a join for `include_optional_linkage_data` (or equivalently for `always_include_*_linkage_data`).

This commit falls back to searching the existing arel nodes for a compatible join and uses that if found.

Fixes cerebris/jsonapi-resources#1449

### All Submissions:

### Reviewer Checklist:

### DIFF
--- a/lib/jsonapi/active_relation/join_manager.rb
+++ b/lib/jsonapi/active_relation/join_manager.rb
@@ -70,7 +70,7 @@ module JSONAPI
         @join_details[segment]
       end
 
-      def self.get_join_arel_node(records, options = {})
+      def self.get_join_arel_node(records, relationship, join_type, options = {})
         init_join_sources = records.arel.join_sources
         init_join_sources_length = init_join_sources.length
 
@@ -80,9 +80,27 @@ module JSONAPI
         if join_sources.length > init_join_sources_length
           last_join = (join_sources - init_join_sources).last
         else
+          # Try to find a pre-existing join for this table.
+          # We can get here if include_optional_linkage_data is true
+          # (or always_include_to_xxx_linkage_data),
+          # and the user's custom `records` method has already added that join.
+          #
+          # If we want a left join and there is already an inner/left join,
+          # then we can use that.
+          # If we want an inner join and there is alrady an inner join,
+          # then we can use that (but not a left join, since that doesn't filter things out).
+          valid_join_types = [Arel::Nodes::InnerJoin]
+          valid_join_types << Arel::Nodes::OuterJoin if join_type == :left
+          table_name = relationship.resource_klass._table_name
+
+          last_join = join_sources.find { |j|
+            valid_join_types.any? { |t| j.is_a?(t) } && j.left.name == table_name
+          }
+        end
+
+        if last_join.nil?
           # :nocov:
           warn "get_join_arel_node: No join added"
-          last_join = nil
           # :nocov:
         end
 
@@ -154,7 +172,7 @@ module JSONAPI
               next
             end
 
-            records, join_node = self.class.get_join_arel_node(records, options) {|records, options|
+            records, join_node = self.class.get_join_arel_node(records, relationship, join_type, options) {|records, options|
               related_resource_klass.join_relationship(
                 records: records,
                 resource_type: related_resource_klass._type,
--- a/test/fixtures/active_record.rb
+++ b/test/fixtures/active_record.rb
@@ -1059,6 +1059,8 @@ module Api
   module V3
     class PostsController < JSONAPI::ResourceController
     end
+    class MoonsController < JSONAPI::ResourceController
+    end
   end
 
   module V4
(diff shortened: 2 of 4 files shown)


averydev:
This has had a hugely positive impact on performance for us.  Thank you!

lgebhardt (approved):
This looks good to me. Thanks @pjungwir

Jasonlgrd:
Hey, thanks for this fix. But I have another problem, more tricky 😬

I have a `user` and `patient_enrolled_journal_event` resources

In `user` resource, I have this relation: 
```ruby
has_one :patient_enrolled_journal_event
```

In `patient_enrolled_journal_event` resource, I have this relation:
```ruby
has_one :patient_enrolled_journal_event,
              exclude_links: :default,
              foreign_key_on: :related,
              relation_name: :patient_enrolled_journal_event
```

In `PatientEnrolledJournalEvent` model, I have this scope: 
```ruby
scope :active, lambda {
    where(archived_at: nil).left_outer_joins(:user).where.not(
      user: {
        id: nil
      }
    )
  }
```

And in `PatientEnrolledJournalEventResource`, I rewrite `records` method to apply this scope like this:
```ruby
    super.active
  end
```

And when I get `http://localhost:3001/api/v1/users?include=patient-enrolled-journal-event` I have this SQL error :
```
SQLite3::SQLException: ambiguous column name: id:\nSELECT patient_enrolled_journal_events.*, \"id\" AS jr_source_id FROM \"patient_enrolled_journal_events\" INNER JOIN \"users\" \"user\" ON \"user\".\"id\" = \"patient_enrolled_journal_events\".\"user_id\" WHERE \"patient_enrolled_journal_events\".\"archived_at\" IS NULL AND \"user\".\"id\" IS NOT NULL AND \"patient_enrolled_journal_events\".\"id\" = ? ORDER BY patient_enrolled_journal_events.id asc\n^",
```

I have investigated, my breakpoint is in `get_join_arel_node` method:
```
last_join = join_sources.find { |j|
            valid_join_types.any? { |t| j.is_a?(t) } && j.left.name == table_name
          }
```
`j.left.name` = user
`table_name` = users
So, the condition is false :(

If I replace my `left_outer_joins` in scope with this:
```
scope :active, lambda {
    where(archived_at: nil).joins("LEFT OUTER JOIN users ON users.id = #{table_name}.user_id").where.not(
      user: {
        id: nil
      }
    )
  }
```
it fix the problem, but I think isn't a good solution.

What do you think about this error ? Thanks 🙏

Jasonlgrd:
I can create an issue but before I would like your opinion on this problem :D

pjungwir:
Thanks for your note! It looks like you found a bug with this PR. We are comparing an attribute name (singular) to a table name (plural). It worked for me because I was joining through a `has_many`, so the attribute name was also plural. But joining in the other direction, as you're doing, has a mismatch. So the fix is to use the *table name* of `j.left`. Diving into Arel for 10 minutes hasn't shown me how to do that yet, but if you find a way then I think it would make a good PR.

pjungwir:
Just guessing, but probably `j.left.relation.name` (assuming `j.left` is going to have a `Arel::Attributes::Attribute`)?

Jasonlgrd:
```
table_name = relationship.resource_klass._table_name

last_join = join_sources.find { |j|
    valid_join_types.any? { |t| j.is_a?(t) } && j.left.name == table_name
}
```
You compare a relation name with a table name above.
What do you think about to replace this:
```
table_name = relationship.resource_klass._table_name
```
with
```
table_name = relationship.name
```

(It's a idea for the moment, I don't launched tests)

pjungwir:
Does it make more sense to match on the table name or the Rails association name? I lean toward the table name (see my comments just above), since associations could have different names but refer to the same table---but maybe the association name is okay instead. If that's what you prefer, then you should rename the local variable from `table_name` to something like `association_name`.

Jasonlgrd:
The `join_sources[0].left` is a `Arel::Nodes::TableAlias` class so we have a `table_name` getter. I think we can use this getter instead of `name` ?

My suggestion
```
 last_join = join_sources.find { |j|
            valid_join_types.any? { |t| j.is_a?(t) } && j.left.table_name == table_name
          }
```

And now we compare the table names. What do you think ?

pjungwir:
That looks right to me!

Jasonlgrd:
Thanks, I will create a PR to propose a fix !

